Issue #10, opened 2018-11-05
Duplicate return values from InternalOakMap.getThreadIndex()


yonigottesman:
In the current implementation 2 threads may return the same value of 
(Thread.currentThread().getId() % Chunk.MAX_THREADS)
Causing everything to break.
Fix:
Each thread will get assigned an index on demand. 
getThreadIndex becomes non static, and each thread that enters get its index from a map that maps threadId to index. first time it enters its index is taken from an atomic counter.

PR #11, opened 2018-11-05
[#10] Properly calculate threadIndex. 

yonigottesman:
fix issue #10 
Reorg of code. add jmh benchmarking


yonigottesman:
Hi I did some reorganization of the code into modules so that we can use the jmh benchmarks framework.
Now after you build oak you can run:
 java -jar -Xmx8g -XX:MaxDirectMemorySize=8g ./benchmarks/target/benchmarks.jar Ingestion -p numRows=100000
And see oak ingestion speed

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -129,15 +128,16 @@
         else
             this.state = new AtomicReference<>(State.INFANT);
         this.next = new AtomicMarkableReference<>(null, false);
-        this.pendingOps = new AtomicReferenceArray<>(MAX_THREADS);
+        this.pendingOps = new AtomicReferenceArray<>(ThreadIndexCalculator.MAX_THREADS);
         this.rebalancer = new AtomicReference<>(null); // to be updated on rebalance
         this.statistics = new Statistics();
         this.comparator = comparator;
-        this.byteBufferPerThread = new ByteBuffer[MAX_THREADS]; // init to null
+        this.byteBufferPerThread = new ByteBuffer[ThreadIndexCalculator.MAX_THREADS]; // init to null
```
Eshcar:
```suggestion
        this.byteBufferPerThread = new ByteBuffer[ThreadIndexCalculator.MAX_THREADS]; 
```

core/src/main/java/com/oath/oak/OakMapBuilder.java
```diff
@@ -47,66 +47,60 @@ public OakMapBuilder() {
     this.memoryAllocator = null;
   }
 
-  public OakMapBuilder setKeySerializer(OakSerializer<K> keySerializer) {
+  public OakMapBuilder<K, V> setKeySerializer(OakSerializer<K> keySerializer) {
```
Eshcar: These <K,V> supplements are not really part of this pull request, correct?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -0,0 +1,35 @@
...
+        if (index != null) {
+            return index;
+        }
+        int newIndex = nextFreeIndex.getAndIncrement();
+        assert (newIndex < MAX_THREADS);
+        map.put(tid, newIndex);
+        return newIndex;
+    }
+
+    public static ThreadIndexCalculator newInstance() {
```
Eshcar: What is the rational behind having a static public method that invokes a private c'tor vs having a public c'tor?

core/src/main/java/com/oath/oak/OakMapBuilder.java
yonigottesman: Correct.
Im piggybacking this pr to fix some coding errors.
I also added jmh benchmarks and that needed some reorg of the code into modules.
I also fixed some test errors that caused tests to fail because autoclosable objects where not in try-with-resource block.

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+    public int getIndex() {
+        long tid = Thread.currentThread().getId();
+        Integer index = map.get(tid);
+        if (index != null) {
+            return index;
+        }
+        int newIndex = nextFreeIndex.getAndIncrement();
+        assert (newIndex < MAX_THREADS);
```
Eshcar: Why do we need this limit of max threads?

benchmarks/pom.xml
```diff
@@ -0,0 +1,148 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Eshcar: A general request - please avoid stacking several orthogonal changes in a single pull request
it makes it harder to review and focus on the relevant changes to check

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
yonigottesman: I dont know why its "32" but many places in the code statically allocate arrays of this size. Changing this to not have an upper limit would affect many places and maybe we should do it as another issue

core/src/main/java/com/oath/oak/Chunk.java
Eshcar:
```suggestion
        this.byteBufferPerThread = new ByteBuffer[ThreadIndexCalculator.MAX_THREADS]; 
```

core/src/main/java/com/oath/oak/OakMapBuilder.java
Eshcar: These <K,V> supplements are not really part of this pull request, correct?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -0,0 +1,35 @@
...
+    static final int MAX_THREADS = 32;
+
+    private ThreadIndexCalculator() {
+        this.nextFreeIndex = new AtomicInteger(0);
+        map = new ConcurrentHashMap<>();
+    }
+
+
+    public int getIndex() {
+        long tid = Thread.currentThread().getId();
```
Eshcar: I don't understand why is this solution correct or necessary.
Are we only trying to change the order of the indices?
I thought we had to maintain a set of "available" indices, which means a adding and removing items from the set.
Right now it seems that once all thread ask for their index it is permanent.
So why not using a simple modulo operation?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
Eshcar: Why do we need this limit of max threads?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
Eshcar: What is the rational behind having a static public method that invokes a private c'tor vs having a public c'tor?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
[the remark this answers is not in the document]
yonigottesman: because the jvm has no guarantee on thread ids.
So there are cases that there are 3 threads and the ids are 0,32,64 this causes all threads to go to same index.

benchmarks/pom.xml
yonigottesman: You are right

Issue #12, opened 2018-11-12
Oak iterator remove AutoCloseable


yonigottesman:
In InternalOakMap.Iter next() function we do:
```
        public T next() {
            try {
                memoryManager.startOperation();
                return internalNext();
            } finally {
                memoryManager.stopOperation();
            }
        }
```
So why do we need start and stop operation in object constructor and close function?
There are 2 problems with this:
1) as long as an iterator exists buffers that are freed after it was created will never be really freed.
2) When the iterator is statically assigned to an Iterator variable then the close is never called causing memory leaks.

PR #14, opened 2018-11-22
[#12] Implement new iterator without autoclose

yonigottesman:
Still need to add more tests but the algorithm is there

### DIFF
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -900,12 +900,19 @@ public class Chunk<K, V> {
             next = get(binaryFind(from), OFFSET_NEXT);
 
             int handle = get(next, OFFSET_HANDLE_INDEX);
+
+            int compare=-1;
+            if (next != Chunk.NONE)
+                compare = compare(from,readKey(next));
+
             while (next != Chunk.NONE &&
-                    (compare(from,readKey(next)) > 0 ||
-                    (compare(from,readKey(next)) >= 0 && !inclusive)||
+                    (compare > 0 ||
+                    (compare >= 0 && !inclusive)||
                     handle < 0)) {
                 next = get(next, OFFSET_NEXT);
                 handle = get(next, OFFSET_HANDLE_INDEX);
+                if (next != Chunk.NONE)
+                    compare = compare(from,readKey(next));
             }
         }
 
@@ -1105,7 +1112,7 @@ public class Chunk<K, V> {
     }
 
 
-    public void signIterator(InternalOakMap.Iter iterator) {
+    public void signInIterator(InternalOakMap.Iter iterator) {
         viewers.add(iterator);
     }
 
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -147,6 +147,7 @@ public class Chunk<K, V> {
         INFANT,
         NORMAL,
         FROZEN,
+        DETACHED,
         RELEASED
     }
 
@@ -168,8 +169,13 @@ public class Chunk<K, V> {
 
     /*-------------- Methods --------------*/
 
+    void detach() {
+        state.compareAndSet(State.FROZEN, State.DETACHED);
+    }
+
+
     void release() {
-        if (state.compareAndSet(State.FROZEN, State.RELEASED)) {
+        if (state.compareAndSet(State.DETACHED, State.RELEASED)) {
             keysManager.release();
         }
     }
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -12,6 +12,7 @@ import java.lang.reflect.Constructor;
 import java.nio.ByteBuffer;
 import java.util.Comparator;
 import java.util.EmptyStackException;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicMarkableReference;
@@ -82,6 +83,7 @@ public class Chunk<K, V> {
     // for writing the keys into the bytebuffers
     private final OakSerializer<K> keySerializer;
     private final OakSerializer<V> valueSerializer;
+    private final ConcurrentSkipListSet<InternalOakMap.Iter> viewers;
 
     /*-------------- Constructors --------------*/
 
@@ -138,6 +140,7 @@ public class Chunk<K, V> {
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
         this.threadIndexCalculator = threadIndexCalculator;
+        viewers = new ConcurrentSkipListSet<>(Comparator.comparingInt(System::identityHashCode));
     }
 
     enum State {
@@ -862,8 +865,8 @@ public class Chunk<K, V> {
         return new AscendingIter();
     }
 
-    AscendingIter ascendingIter(K from) {
-        return new AscendingIter(from);
+    AscendingIter ascendingIter(K from, boolean inclusive) {
+        return new AscendingIter(from, inclusive);
     }
 
     DescendingIter descendingIter() {
@@ -888,19 +891,19 @@ public class Chunk<K, V> {
             next = get(HEAD_NODE, OFFSET_NEXT);
             int handle = get(next, OFFSET_HANDLE_INDEX);
             while (next != Chunk.NONE && handle < 0) {
-//                while (next != Chunk.NONE && (handle < 0 || (handle > 0 && handles[handle].isDeleted()))) {
-                // if there is a next but it was removed then try next item
                 next = get(next, OFFSET_NEXT);
                 handle = get(next, OFFSET_HANDLE_INDEX);
             }
         }
 
-        AscendingIter(K from) {
+        AscendingIter(K from, boolean inclusive) {
             next = get(binaryFind(from), OFFSET_NEXT);
+
             int handle = get(next, OFFSET_HANDLE_INDEX);
-            while (next != Chunk.NONE && handle < 0) {
-//                while (next != Chunk.NONE && (handle < 0 || (handle > 0 && handles[handle].isDeleted()))) {
-                // if there is a next but it was removed then try next item
+            while (next != Chunk.NONE &&
+                    (compare(from,readKey(next)) > 0 ||
+                    (compare(from,readKey(next)) >= 0 && !inclusive)||
+                    handle < 0)) {
                 next = get(next, OFFSET_NEXT);
                 handle = get(next, OFFSET_HANDLE_INDEX);
             }
@@ -910,8 +913,6 @@ public class Chunk<K, V> {
             next = get(next, OFFSET_NEXT);
             int handle = get(next, OFFSET_HANDLE_INDEX);
             while (next != Chunk.NONE && handle < 0) {
-//                while (next != Chunk.NONE && (handle < 0 || (handle > 0 && handles[handle].isDeleted()))) {
-                // if there is a next but it was removed then try next item
                 next = get(next, OFFSET_NEXT);
                 handle = get(next, OFFSET_HANDLE_INDEX);
             }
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -164,11 +167,25 @@ long memorySize() {
 
         updateIndexAndNormalize(engaged, newChunks);
 
-        for (Chunk chunk : engaged
-        ) {
+        for (Chunk chunk : engaged) {
+            ConcurrentSkipListSet<Iter> viewingIterators = chunk.getSignedIterators();
             chunk.release();
```
Eshcar: shouldn't we wait with the release till after we set last key in iterators?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -164,11 +167,25 @@ long memorySize() {
...
-        }
+            viewingIterators.forEach(iterator -> {
+                if (iterator.getNextKeyIndex() != Chunk.NONE) {
```
Eshcar: call getNextKeyIndex() once and compare result twice
is it safe to call it twice? can the index change? if not add a comment explaining exactly why it has not changed?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -164,11 +167,25 @@ long memorySize() {
 
         updateIndexAndNormalize(engaged, newChunks);
 
-        for (Chunk chunk : engaged
-        ) {
+        for (Chunk chunk : engaged) {
+            ConcurrentSkipListSet<Iter> viewingIterators = chunk.getSignedIterators();
```
Eshcar: add a comment explaining what is the goal of this checks - this is a crucial piece of code
also it helps reviewing

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -164,11 +167,25 @@ long memorySize() {
...
-        }
+            viewingIterators.forEach(iterator -> {
+                if (iterator.getNextKeyIndex() != Chunk.NONE) {
 
+                    int index = iterator.getNextKeyIndex();
+
+                    // Check if iterator moved to next chunk.
+                    // index can be 0 if iterator just finished iterating all chunks, and this thread started this
+                    // iteration before the iterator removed itself from viewers
+                    if (iterator.nextChunk == chunk && index != 0) {
```
Eshcar: but isn't 0 a valid index?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -786,15 +807,10 @@ private Chunk findChunk(Object key) {
             // we use another nested attach-detach invocation to allow releasing the memory where
             // iterator already traversed. Finally to mark the thread idle we need the detach to be
             // invoked from the close of this closeable iterator.
-            memoryManager.startOperation();
-            next = Chunk.NONE;
+            nextIndex = Chunk.NONE;
             nextHandle = null;
+            memoryManager.startOperation();
```
Eshcar: move the start-stop code into a findFirstKey() method

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -810,12 +826,17 @@ boolean tooHigh(Object key) {
                     (c == 0 && !hiInclusive)));
         }
 
+
         boolean inBounds(Object key) {
-            return !tooLow(key) && !tooHigh(key);
+            if (!isDescending) {
```
Eshcar: you have assumption on where it is called
need to document this assumption
and make this method private

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
         abstract T internalNext();
 
         /**
          * Advances next to higher entry.
+         * Return previous index
          */
-        void advance() {
-            if (next == Chunk.NONE)
+        Pair<ByteBuffer, Handle> advance() {
+
```
Eshcar: need to add comment explaining why these tests are needed

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
+            Object lastKey = lastKeyBeforeChunkDelete.get();
+            if (!(lastKey instanceof Chunk)) {
```
Eshcar: change to if(lastKey instanceof K)

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
+                else {
+                    loInclusive = true;
+                    lo = (K)lastKeyBeforeChunkDelete.get();
+                }
+                initChunk();
+                if (nextChunkIter == null) {
+                    throw new NoSuchElementException();
+                }
+                if (nextIndex == Chunk.NONE) {
+                    throw new ConcurrentModificationException();
```
Eshcar: add message in exception something like "next key may have been deleted while iterating the keys"

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
+                }
+            }
+
```
Eshcar: add comment - this is the case where lastKey instanceof Chunk

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
                 nextChunk = hi != null ? skiplist.floorEntry(hi).getValue()
                         : skiplist.lastEntry().getValue();
-                if (nextChunk == null) {
-                    nextChunkIter = null;
-                } else {
+                if (nextChunk != null) {
                     nextChunkIter = hi != null ? nextChunk.descendingIter(hi, hiInclusive) : nextChunk.descendingIter();
                 }
             }
+            lastKeyBeforeChunkDelete = new AtomicReference<>(nextChunk);
```
Eshcar: CAS?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
+
+            if (nextIndex == Chunk.NONE) {
                 throw new NoSuchElementException();
-            findNextKeyInRange();
+            }
+
+
+            ByteBuffer bb = getKey(nextIndex, nextChunk);
+            Handle currentHandle = nextHandle;
+            setNextKey();
```
Eshcar: this can sign out from viewers
do you need to check this?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
+         * if such a key doesn't exists then next is set to Chunk.NONE and nextValue is set to null
+         */
+        private void setNextKey() {
+
+            Chunk<K, V> currentChunk = nextChunk;
+
+
             while (!nextChunkIter.hasNext()) { // chunks can have only removed keys
-                nextChunk = nextChunk.next.getReference(); // try next chunk
+                nextChunk.signoutIterator(this);
```
Eshcar: how can you be sure you already sign-in to the next chunk?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -827,159 +848,170 @@ public T next() {
...
-                next = nextChunkIter.next();
-                key = getKey(next, nextChunk);
+
+            //Must sign iterator after prev stages
+            if (currentChunk != nextChunk) {
+                nextChunk.signIterator(this);
+            }
+
+            //Check if released for frozen and store last key
+            if (nextChunk.state() == Chunk.State.RELEASED) {
```
Eshcar: shouldn't  we check this on every next() method?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -888,19 +891,19 @@ DescendingIter descendingIter(K from, boolean inclusive) {
...
-        AscendingIter(K from) {
+        AscendingIter(K from, boolean inclusive) {
             next = get(binaryFind(from), OFFSET_NEXT);
+
             int handle = get(next, OFFSET_HANDLE_INDEX);
-            while (next != Chunk.NONE && handle < 0) {
-//                while (next != Chunk.NONE && (handle < 0 || (handle > 0 && handles[handle].isDeleted()))) {
-                // if there is a next but it was removed then try next item
+            while (next != Chunk.NONE &&
+                    (compare(from,readKey(next)) > 0 ||
```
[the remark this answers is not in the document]
yonigottesman: Yes, ill fix

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -1104,6 +1105,19 @@ int size() {
     }
 
 
+    public void signIterator(InternalOakMap.Iter iterator) {
```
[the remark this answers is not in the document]
yonigottesman: ill fix

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: No, the iterator does : 1) signin chunk. 2) check if chunk released
If we put the release after set lastkey there could be the following error:
1) iterator moves to new chunk. (not yet signin chunk)
2) rebalancer reads viewingIterator list but its still empty
3) iterator continues signin (will be lost) and checks if chunk marked release returns false
4) rebalancer marks chunk released and deletes chunk and buffers
5) iterator tries to read released buffer and all hell breaks loose.

If we first release then either the iterator will see the chunk is released, or the rebalancer will see the iterator in the list

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: Ill make this code more clear. I can call getNext only once and its ok.

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: No need, the iterator is restarting

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: I dont understand you comment.
setNextKey can sigout and signin into a new chunk. why does this bother you?

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: There is no problem signing out even if your are not signing in.
This is exactly what will happen in this loop if we skip a few chunks

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: It is checked every next()

yonigottesman:
@Eshcar I agree

core/src/main/java/com/oath/oak/InternalOakMap.java
Eshcar: then this means the rebalancer code should
1) chunk release (currently line 172)
2) collect viewing iterators (currently line 171)
always reverse the order of the iterator at critical lines.

yonigottesman: The critical operations are 1) set/check lastKey 2) set/check release.
So leaving the release after collecting viewing iterators should be ok.

I think we should do the release first just to make the code clearer but i dont think its wrong this way.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -810,176 +883,213 @@ boolean tooHigh(Object key) {
...
 
         public final boolean hasNext() {
-            return next != Chunk.NONE;
+            return state.get().validState();
         }
 
         public T next() {
             try {
-                memoryManager.startOperation();
+                if (epochUsageCounter-- <= 0) {
```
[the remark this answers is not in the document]
yonigottesman: each iterator has its own epoch state. So im not sure how to move this to MemoryManager without having to access concurrent maps to get the state of a specific iterator.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -810,176 +883,213 @@ boolean tooHigh(Object key) {
...
-                    }
-                }
-                nextChunkIter = nextChunk.descendingIter(keySerializer.deserialize(serializedMinKey), false); // TODO check correctness
-            }
-            return true;
-        }
 
-        private void descend() {
-            if (!findPrevChunk()) {
+        // Long explanation here
```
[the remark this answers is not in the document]
yonigottesman: :(

PR #17, opened 2019-01-14
Adding synchrobench based benchmark

eranmeir:
- Separated benchmarks to two modules: jmh and synchrobench. Shuffled files to match
- Created synchrobench module to benchmarks subdirectory.

(Note: this PR doesn't include memory allocator contention fix)


yonigottesman:
Another more general question. Is it ok to merge someone elses project into ours like this?

benchmarks/jmh/pom.xml
```diff
@@ -6,16 +6,17 @@
         <artifactId>root</artifactId>
         <groupId>com.yahoo.oak</groupId>
         <version>0.1.4-SNAPSHOT</version>
```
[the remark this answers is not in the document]
eranmeir: I've modified the root pom to have `<dependencyManagement>` element so that submodules can include oak core dependency without explicit version.
Using properties for versioning is problematic. Instead, we can use versions-maven-plugin:
` mvn versions:set -DnewVersion=0.1.5-SNAPSHOT`

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBufferOak.java
```diff
@@ -0,0 +1,106 @@
...
+            return ret;
+        }
+
+        @Override
+        public int calculateSize(MyBuffer object) {
+            return object.buffer.capacity() + Integer.BYTES;
+        }
+    };
+
+    static OakComparator<MyBuffer> keysComparator = new OakComparator<MyBuffer>() {
```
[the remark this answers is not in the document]
eranmeir: Fixed, with a minor caveat: when a key is serialized the test puts the buffer capacity as the first integer in the buffer, so had to shift position before comparing.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
```diff
@@ -0,0 +1,125 @@
...
+    public OakMap() {
+        minKey = new MyBuffer(Integer.BYTES);
+        minKey.buffer.putInt(0, Integer.MIN_VALUE);
+        builder = new OakMapBuilder<MyBuffer, MyBuffer>()
+                .setKeySerializer(MyBufferOak.serializer)
+                .setValueSerializer(MyBufferOak.serializer)
+                .setMinKey(minKey)
+                .setComparator(MyBufferOak.keysComparator)
+                .setChunkBytesPerItem(Chunk.BYTES_PER_ITEM_DEFAULT)
+                .setChunkMaxItems(Chunk.MAX_ITEMS_DEFAULT)
```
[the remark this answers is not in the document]
eranmeir: Not fixing for now. Will add when implementing concrete workflows.

eranmeir:
> Another more general question. Is it ok to merge someone elses project into ours like this?

It uses Apache License 2.0, which is pretty permissive, according to [tldrlegal](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)):
> You can do what you like with the software, as long as you include the required notices. This permissive license contains a patent license from the contributors of the code.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBufferOak.java
```diff
@@ -0,0 +1,62 @@
...
+
+        @Override
+        public int compareSerializedKeys(ByteBuffer serializedKey1, ByteBuffer serializedKey2) {
+            return serializedKey1.compareTo(serializedKey2);
+        }
+
+        @Override
+        public int compareSerializedKeyAndKey(ByteBuffer serializedKey, MyBuffer key) {
+            int pos = serializedKey.position();
+            serializedKey.position(pos + Integer.BYTES);
```
yonigottesman: I think this is an error. the compareTo function already starts from position not from 0.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public OakMap() {
+        minKey = new MyBuffer(Integer.BYTES);
+        minKey.buffer.putInt(0, Integer.MIN_VALUE);
+        builder = new OakMapBuilder<MyBuffer, MyBuffer>()
+                .setKeySerializer(MyBufferOak.serializer)
+                .setValueSerializer(MyBufferOak.serializer)
+                .setMinKey(minKey)
+                .setComparator(MyBufferOak.keysComparator)
+                .setChunkBytesPerItem(Parameters.keySize)
```
yonigottesman: add Integer.BYTES ? we serialize the size in the buffer

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
```diff
@@ -0,0 +1,120 @@
...
+
+    }
+
+    @Override
+    public boolean ascendOak(K from, int length) {
+        int i;
+        try (com.oath.oak.OakMap<MyBuffer, MyBuffer> sub = oak.tailMap((MyBuffer) from, true)) {
+            OakBufferView<MyBuffer> oakView = sub.createBufferView();
+            i = 0;
+            OakIterator<OakRBuffer> iter = oakView.keysIterator();
```
yonigottesman: We should probably talk about this with the team. I think its not fair we are returning bufferview and not deserializing. Maybe we can have a command line option to use this or that and see the performance?

eranmeir: I agree. Suggest we open an issue and get to it in a different PR

PR #18, opened 2019-01-22
Mem fix + Benchmark fix/cleanup

eranmeir:
Some MemoryManager refactoring (no design change, just shuffled things around for testability).


benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBufferOak.java
```diff
@@ -69,7 +69,7 @@ public int compareSerializedKeyAndKey(ByteBuffer serializedKey, MyBuffer key) {
             int cap1 = serializedKey.getInt(pos1);
             int pos2 = key.buffer.position();
             int cap2 = key.buffer.capacity();
-            return compare(serializedKey, pos1 + Integer.BYTES, cap1, key.buffer, pos2, cap2);
+            return compare(serializedKey, pos1 + Integer.BYTES, cap1 - Integer.BYTES, key.buffer, pos2, cap2);
```
yonigottesman: Also in line 47 (cannot review it because it didnt change) before return 0 check if cap2 > i.
I realize its not important for this benchmark because all keys are 100, but if someone ever looks at this code to learn and copy how to use oak it should be right.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -37,11 +36,17 @@
     private final AtomicLong allocated = new AtomicLong(0);
 
     // constructor
-    // input param: memory capacity given to this Oak
-    OakNativeMemoryAllocator(long capacity) {
+    // input param: memory capacity given to this Oak. Uses default BlocksPool
+    public OakNativeMemoryAllocator(long capacity) {
+       this(capacity, BlocksPool.getInstance());
```
yonigottesman: Why not go all the way and remove this weird singleton and this constructor?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -71,16 +76,20 @@ public ByteBuffer allocate(int size) {
...
                 }
                 // does allocation of new block brings us out of capacity?
-                if ((blocks.size() + 1) * BlocksPool.BLOCK_SIZE > capacity) {
+                if ((blocks.size() + 1) * blocksProvider.blockSize() > capacity) {
                     throw new OakOutOfMemoryException();
                 } else {
-                    Block b = BlocksPool.getInstance().getBlock();
-                    this.blocks.add(b);
-                    this.currentBlock = b;
+                    synchronized (this) {
```
yonigottesman: Did you try a solution with CAS?

PR #20, opened 2019-01-28
Remove double indeirection

yonigottesman:
Remove the double indirection in the bytebuffers.
This means that all the bytebuffers point to the 100MB buffer and cannot touch the position or limit pointers. Also the capacity it the remaining (limit-position)


core/src/main/java/com/oath/oak/KeysManager.java
```diff
@@ -34,19 +33,15 @@ public int length() {
     }
 
     public void writeKey(K key, int ki) {
-        assert keys.position() == 0;
         ByteBuffer byteBuffer = keys.duplicate();
-        byteBuffer.position(ki);
-        byteBuffer.limit(ki + keySerializer.calculateSize(key));
-        byteBuffer = byteBuffer.slice();
-        assert byteBuffer.position() == 0;
+        byteBuffer.position(keys.position() + ki);
```
eranmeir: If we're duplicating for thread-safety (do we really need to?) then we should probably use `byteBuffer.position()` here, since  `keys.position()` may change:

```suggestion
        byteBuffer.position(byteBuffer.position() + ki);
```

core/src/main/java/com/oath/oak/KeysManager.java
```diff
@@ -34,19 +33,15 @@ public int length() {
...
+        byteBuffer.limit(keys.position() + ki + keySerializer.calculateSize(key));
```
eranmeir: Same here:
```suggestion
        byteBuffer.limit(byteBuffer.position() + keySerializer.calculateSize(key));
```

Also, not sure if for similar reasons we should use `keySerializer.calculateSize(`**`byteBuffer`**`))`, or if it will misbehave with new position/limit

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -69,15 +69,13 @@ void put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
             return;
         }
         int capacity = serializer.calculateSize(newVal);
-        if (this.value.capacity() < capacity) { // can not reuse the existing space
+        if (this.value.remaining() < capacity) { // can not reuse the existing space
             memoryManager.release(this.value);
             this.value = memoryManager.allocate(capacity);
         } else {
-            this.value.clear(); // zero the position
+
```
eranmeir: Remove redundant else

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -69,15 +69,13 @@ void put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
             return;
         }
         int capacity = serializer.calculateSize(newVal);
-        if (this.value.capacity() < capacity) { // can not reuse the existing space
+        if (this.value.remaining() < capacity) { // can not reuse the existing space
```
eranmeir: GitHub won't let me comment on unmodified code, so abusing this line.
In `increaseValueCapacity`, line 34 shouldn't it be `value.remaining()` instead of `value.capacity()`?

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -93,254 +91,126 @@ boolean compute(Consumer<OakWBuffer> computer, MemoryManager memoryManager) {
...
+        value.put(src, value.position() + offset, length);
         return this;
     }
 
     @Override
     public OakWBuffer put(byte[] src) {
-        assert writeLock.isHeldByCurrentThread();
-        value.put(src);
-        return this;
+        throw new NotImplementedException();
```
eranmeir: I suggest removing all not implemented methods from the relevant interface, instead of throwing exceptions (this relates to all `NotImplementedException()`s below).

PR #22, opened 2019-02-07
Synchrobench work

eranmeir:
- Bugfix
- Added bufferview (non-deserializing) option
- Update run script


yonigottesman:
looks good. Merge!

PR #23, opened 2019-02-10
Keys seperate allocations

yonigottesman:
allocate keys directly from ByeBuffer.allocateDirect() without going through the nativeallocator.
This way we can remove start() stop() operations because keys will be release by ByteBuffer framwork when the reference count is 0.

This is needed because druid keeps a reference to the key bytebuffer while iterating and it must not be deleted.


core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -796,14 +768,9 @@ public int getIndex() {
...
 
-        boolean validState() { return state == State.VALID_STATE;}
 
-        public static <K, V> IteratorState<K, V> endState() {return endState;}
-        public static <K, V> IteratorState<K, V> state(Chunk<K, V> nextChunk, Chunk.ChunkIter nextChunkIter, int nextIndex) {
-            return new IteratorState<>(nextChunk, nextChunkIter, nextIndex, State.VALID_STATE);
-        }
         public static <K, V> IteratorState<K, V> initState(Chunk<K, V> nextChunk, Chunk.ChunkIter nextChunkIter) {
-            return new IteratorState<>(nextChunk, nextChunkIter, Chunk.NONE, State.INIT_STATE);
+            return new IteratorState<>(nextChunk, nextChunkIter, Chunk.NONE);
```
eranmeir: Perhaps capitalize this method or rename it to emphasize this is a static factory method and not an ordinary method. Static factory methods [often](https://docs.oracle.com/javase/tutorial/datetime/overview/naming.html) start with `from` or `of`.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -927,90 +886,57 @@ private boolean updateEpoch() {
...
-            //Init state, not valid yet, must move forword
-            IteratorState<K, V> initState = IteratorState.initState(nextChunk, nextChunkIter);
-            state.set(initState);
-
-            lastKeyBeforeRelease = new AtomicReference<>(initState.getChunk());
-
-            // ok to signin init state, the rebalancer will ignore
-            initState.getChunk().signInIterator(this);
 
+            //Init state, not valid yet, must move forword
```
eranmeir:
```suggestion
            //Init state, not valid yet, must move forward
```

core/src/main/java/com/oath/oak/MemoryManager.java
```diff
@@ -6,147 +6,45 @@
...
-            this.timeStamps[i] = new AtomicLong(0);
-        }
-        this.releasedArray = new ArrayList<>(ThreadIndexCalculator.MAX_THREADS);
-        for (int i = 0; i < ThreadIndexCalculator.MAX_THREADS; i++) {
-            releasedArray.add(i, new LinkedList<>());
-        }
-        max = new AtomicLong(0);
-        releases = RELEASES_DEFAULT;
-        this.threadIndexCalculator = threadIndexCalculator;
+        keysMemoryAllocator = new DirectMemoryAllocator();
```
eranmeir: I think concerns are separated better if you inject `keysMemoryAllocator` as a constructor argument (and allocate a new `DirectMemoryAllocator` if null)

core/src/test/java/com/oath/oak/OakMemoryManagerTest.java
```diff
@@ -53,325 +53,4 @@ public void allocate() {
         assertEquals(4, bb.remaining());
         assertEquals(8, memoryManager.allocated());
     }
```
eranmeir: Why did you remove unit tests?

core/src/test/java/com/oath/oak/IteratorModificationTest.java
```diff
@@ -231,87 +233,55 @@ public void concurrentModificationTest() throws InterruptedException {
...
-//    }
+
+    @Test
+    public void valueDeleteTest() throws InterruptedException {
+
+        OakIterator<Map.Entry<String, String>> enrtyIterator = oak.entriesIterator();
+        assertTrue(enrtyIterator.hasNext());
+        oak.remove(generateString(0, KEY_SIZE));
+        assertTrue(enrtyIterator.hasNext());
+        assertEquals(null, enrtyIterator.next());
```
eranmeir:
```suggestion
        assertNull(enrtyIterator.next());
```

core/src/test/java/com/oath/oak/OakMemoryManagerTest.java
yonigottesman: because they dont mean anything anymore. they check when the release list is freed but in this new implementation there is no release list

PR #25, opened 2019-02-13
Remove interfaces

yonigottesman:
Remove redundant OakWBuffer


core/src/test/java/com/oath/oak/OakViewTests.java
```diff
@@ -0,0 +1,194 @@
...
+    @After
+    public void tearDown() {
+        oak.close();
+    }
+
+    @Test
+    public void uniTestOakRBuffer(){
+
+
+        Function<ByteBuffer, String> deserialize = (byteBuffer) -> {
```
eranmeir: Why not extract a proper method out of this?

core/src/test/java/com/oath/oak/OakViewTests.java
```diff
@@ -0,0 +1,194 @@
...
+
+            for(int i =0 ; i < testValBytes.length/Long.BYTES; ++i) {
+                assertEquals(testValBB.getLong(i), valBuffer.getLong(i));
+            }
+        }
+    }
+
+
+
+    @Test
```
eranmeir: Can use this:
```suggestion
    @Test(expected = ConcurrentModificationException.class)
```

and remove the try-catch-fail logic below

PR #28, opened 2019-03-12
Redesign puts

yonigottesman:


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -459,7 +461,7 @@ int linkEntry(int ei, boolean cas, K key) {
                       }
                     }
                   }
-                  return ei;
+                  return new LinkEntryResult(ei, true);
                 }
                 // CAS didn't succeed, try again
             } else {
```
Eshcar: why not return in else?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -404,7 +387,26 @@ int allocateEntryAndKey(K key) {
...
+
+        public LinkEntryResult(int ei, boolean newEntry) {
+            this.ei = ei;
+            this.newEntry = newEntry;
+        }
+    }
+
+    public LinkEntryResult linkEntry(int ei, boolean cas, K key) {
         int prev, curr, cmp;
         int anchor = -1;
```
Eshcar: what about comparing the anchor to key?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -629,46 +546,7 @@ private int getLastItemEntryIndex() {
...
-                    casSuccess = casPendingArray(i, null, FROZEN_OP_DATA);
-                }
-                // the frozen was set beforehand
-                // therefor a put could not publish or unpublish again
-                opData = pendingOps.get(i);
-                assert (casSuccess || opData == FROZEN_OP_DATA);
-            }
-
-        }
+        while (pendingOps.get() != 0);
```
Eshcar: exponential backoff??

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -432,7 +434,7 @@ int linkEntry(int ei, boolean cas, K key) {
```
Eshcar: change inner loop to while(curr!=NONE) ?

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -19,57 +18,97 @@
...
+        int capacity = serializer.calculateSize(newVal);
+        this.value = memoryManager.allocate(capacity);
+        serializer.serialize(newVal, this.value.slice());
     }
 
     boolean isDeleted() {
-        return deleted.get();
+        readLock.lock();
+        boolean retval = value == null;
+        readLock.unlock();
```
Eshcar: should move all lock-unlock to try-finally blocks

core/src/main/java/com/oath/oak/Chunk.java
yonigottesman: Because we do the loop again and try to find the entry that did get to the cas first

core/src/test/java/com/oath/oak/MultiThreadTest.java
```diff
@@ -206,6 +206,7 @@ public void run() {
                 oak.putIfAbsent(i, i);
                 value = oak.get(i);
                 //TODO YONIGO - value can be null because of remove
```
Eshcar: remove todo comment

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -109,7 +105,6 @@
           AtomicInteger externalSize,
           OakSerializer<K> keySerializer,
           OakSerializer<V> valueSerializer, ThreadIndexCalculator threadIndexCalculator) {
-        this.memoryManager = memoryManager;
```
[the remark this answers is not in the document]
yonigottesman: Not in the chunk. ;)

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -981,7 +787,7 @@ private void advanceState() {
         @Override
         public OakRBuffer next() {
             Handle handle = advance().getValue();
-            if (handle == null)
+            if (handle.isDeleted())
```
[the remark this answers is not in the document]
yonigottesman: Hopefully this will be figured out when we have the new interface.
Maybe we should just return a handle that is deleted...

core/src/main/java/com/oath/oak/Chunk.java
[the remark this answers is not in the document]
eranmeir: Perhaps it's better to keep a reference to it in the chunk, so that it can pass it on to handles when creating new ones. That way handles can have their own reference and shouldn't have to take a memory manager argument in some of their methods

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -156,7 +155,7 @@ long memorySize() {
         // will be redirected to help the rebalance procedure
         rebalancer.freeze();
 
-        Rebalancer.RebalanceResult result = rebalancer.createNewChunks(key, value, computer, op); // split or compact
+        rebalancer.createNewChunks(); // split or compact
         // if returned true then this thread was responsible for the creation of the new chunks
```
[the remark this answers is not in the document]
yonigottesman: the rebalance now doesn't do any op, so nothing to get lost

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -170,16 +169,11 @@ long memorySize() {
...
         engaged.forEach(Chunk::release);
 
-        if (result.success && result.putIfAbsent) {
-            size.incrementAndGet();
-        }
-
-        return result;
     }
 
     private void checkRebalance(Chunk c) {
```
yonigottesman: This function is almost never called. (only remove)
When should we check if a rebalance is needed? with new put implementation rebalance is done only when out of space. is this ok? @Eshcar @eranmeir @sanastas

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -332,280 +286,132 @@ void put(K key, V value) {
...
+        //attache handle to entry
+        c.pointToValue(ei, hi);
+
+        if (!c.publish()) {
+            rebalance(c);
+            return AbstractPutResult.REBALANCED;
+        }
+
+        //attache entry to chunk list
+        c.getHandle(ei).setValue(value, valueSerializer, memoryManager);
```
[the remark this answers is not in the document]
yonigottesman: yes. the entry isnt attached yet, so no one can access this handler yet

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -332,280 +286,132 @@ void put(K key, V value) {
...
+        AbstractPutResult retval;
+        if(!linkResult.isNewEntry()) {
+            //another thread inserted this key so we compute
+            c.getHandle(ei).remove(memoryManager);
+            if (operationIfFound.apply(c.getHandle(linkResult.getEi()))) {
+                c.getStatistics().incrementAddedCount();
+                size.incrementAndGet();
+                retval = AbstractPutResult.NEW_KEY_ADDED;
+            } else {
+                retval = AbstractPutResult.NEW_KEY_NOT_ADDED;
```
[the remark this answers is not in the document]
yonigottesman: I dont understand please elaborate

PR #30, opened 2019-03-18
Oak API Refactoring

eranmeir:
- OakMap implements ConcurrentNavigableMap
- Removed (Transform and Buffer) view classes
- Added basic unit tests covering API usage
- Updated tests and benchmarks to use new API
- Zero-Copy versions are exposed using the ZC() method
- Some interface methods are still not implemented


core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -252,9 +250,33 @@ public void putIfAbsentComputeIfPresent(K key, V value, Consumer<ByteBuffer> com
...
+    }
+
+
+    /* ---------------- ConcurrentNavigableMap API methods -------------- */
+
     /*-------------- SubMap --------------*/
-    // package visibility to be used by the views
-    public boolean inBounds(K key) {
+
+    boolean inBounds(Object key) {
```
Eshcar: why changed to Object and not Key?

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -391,59 +426,470 @@ public boolean inBounds(K key) {
...
+        return new Values<>(this);
+    }
+
+    /* ------ Zero-Copy API methods  ------ */
+
+    public ZeroCopyMap ZC() {
+        return new ZeroCopyMap(this);
+    }
+
+    public static class ZeroCopyMap<K, V>  {
```
Eshcar: Can this class move to a separate file? and anyway it shouldnt be in the middle of the class - this is very confusing
Shouldn't this extend OakMap or NavigableSomething? Or maybe have a common ancestor for both ZCMap and OakMap?

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -391,59 +426,470 @@ public boolean inBounds(K key) {
...
+
+    @Override
+    public Entry<K, V> pollLastEntry() {
+        throw new UnsupportedOperationException();
+    }
+
+    /* ---------------- View Classes -------------- */
+
+    static class KeySet<K> extends AbstractSet<K> implements NavigableSet<K> {
+
```
Eshcar: is there a way to handle this with inheritance instead of composition?

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -391,59 +426,470 @@ public boolean inBounds(K key) {
...
+        @Override
+        public Iterator<OakRBuffer> iterator() {
+            return m.valuesBufferIterator();
+        }
+
+        @Override
+        public int size() {
+            return m.size();
+        }
     }
```
Eshcar: will continue from here tomorrow

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -729,6 +741,81 @@ private Chunk findChunk(Object key) {
         return c;
     }
 
+    V replace(K key, V value, Function<ByteBuffer,V> valueDeserializeTransformer) {
```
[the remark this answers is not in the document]
eranmeir: Yes. Added two replace methods to the API required for implementing `ConcurrentMap`

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -9,19 +9,13 @@
 import javafx.util.Pair;
 
 import java.nio.ByteBuffer;
-import java.util.Comparator;
-import java.util.ConcurrentModificationException;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
+import java.util.*;
```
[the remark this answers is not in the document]
eranmeir: IDEA does that automatically :) Expanded back

core/src/main/java/com/oath/oak/OakMap.java
eranmeir: The standard Map interfaces take Object arguments (to preserve the substitution principle) and it trickled down here

core/src/main/java/com/oath/oak/OakMap.java
eranmeir: We can move this to a separate file and lose some type safety, since the type parameters K and V will not be bound. 
There are no common ancestors since our zero-copy implementation make some assumptions about in/out arguments to some methods that are in conflict with standard interfaces (for example, we use bytebuffers instead of keys/values in certain places).

core/src/main/java/com/oath/oak/OakMap.java
eranmeir: Could you elaborate? I'm not sure I understand what set you mean we should extend.

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -98,95 +98,62 @@ private OakMap(InternalOakMap internalOakMap, MemoryManager memoryManager,
...
-     * @param value value to be associated with the specified key
-     * @return {@code true} if there was no mapping for the key
-     * @throws IllegalArgumentException if the specified key or value is null
-     */
-    public boolean putIfAbsent(K key, V value) {
-        if (key == null || value == null)
-            throw new IllegalArgumentException();
-        if (!inBounds(key))
-            throw new IllegalArgumentException("The key is out of map bounds");
-        return internalOakMap.putIfAbsent(key, value);
```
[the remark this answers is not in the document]
eranmeir: There are checks in internal oak map, but I also added checks in the API wrappers themselves, since some requirements may be different (e.g, checking if a transformer is null).

README.md
```diff
@@ -127,92 +126,93 @@ OakMap<K,V> oak = builder.build();
 
 ### OakMap Methods
 You are welcome to take a look on the OakMap's [full API](https://github.com/yahoo/Oak/wiki/Full-API).
-OakMap's API is similar to the ConcurrentNavigableMap API; some non-standard API methods and special cases are discussed below.
+OakMap's API implements the ConcurrentNavigableMap interface. For improved performance, it offers additional non-standard zero-copy API methods that are discussed below.
 
-### OakBuffers
-OakMap provides two types of memory buffers: *OakRBuffer* (read-only) and *OakWBuffer* (read and write). These buffers support the standard Java API for read-only Java ByteBuffers and writable Java ByteBuffers, respectively.
+### OakRBuffers
+OakMap provides read-only memory buffers, *OakRBuffer*` that support the standard Java API for read-only Java ByteBuffers.
```
sanastas: I don't think it is a good idea to withdraw the OakWBuffer interface. This make it difficult to manage (reuse) the memory.

README.md
```diff
@@ -127,92 +126,93 @@ OakMap<K,V> oak = builder.build();
...
 
-An OakRBuffer can represent either a key or a value. The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Notice that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
+The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Note that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
+
+For backward compatibility with applications that are already based on the use of ByteBuffers, OakRBuffer provides the transform method that atomically applies a transformation to the underlying ByteBuffer. For a more comprehensive code example please refer to the [Usage](#usage) section.
```
sanastas: If use of ByteBuffer then only read-only ByteBuffer. It is not OK to give the direct ByteBuffer to the user. It may end up accessing wrong memory...

README.md
```diff
@@ -127,92 +126,93 @@ OakMap<K,V> oak = builder.build();
 
 ### OakMap Methods
 You are welcome to take a look on the OakMap's [full API](https://github.com/yahoo/Oak/wiki/Full-API).
```
sanastas: All the other wiki pages, have they been updated?

README.md
```diff
@@ -127,92 +126,93 @@ OakMap<K,V> oak = builder.build();
...
-However, note that each sub-map is in particular an OakMap and thus AutoCloseable and needs to be closed (explicitly or implicitly). Similarly, OakBufferView and OakTransformView are AutoCloseable objects and need to be closed. Again, close() can be invoked on different objects referring to the same underlying memory, but the final release will happen only once.
+Please pay attention that multiple views can be defined to the same underlying memory of OakMap. That is when other sub-maps are created. Do not worry, the true memory release will happen only when last of those views is closed.
+However, note that each sub-map is in particular an OakMap and thus AutoCloseable and needs to be closed (explicitly or implicitly). Again, close() can be invoked on different objects referring to the same underlying memory, but the final release will happen only once.
 
 ## Usage
 
 An Integer to Integer build example can be seen in [Code Examples](https://github.com/yahoo/Oak/wiki/Code-Examples). Here we illustrate individual operations.
```
sanastas: Also here check the link...

README.md
eranmeir: The wiki is not part of the codebase, so not part of the PR, but I will update.

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -7,16 +7,25 @@
...
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
 /**
  * A concurrent map implementation which supports off-heap memory.
  */
-public class OakMap<K, V> implements AutoCloseable {
+public class OakMap<K, V> extends AbstractMap<K, V> implements AutoCloseable, ConcurrentNavigableMap<K, V> {
```
yonigottesman: OakMap should not be autoclosable because then you have to call close() for every creation.
This is very confusing and even you in your API test didnt do it:
keySet = oak.subMap(3, 5).keySet(); - the submap function creates a new oak.
You will not be able to do nice things like:
oak.subMap(0, 100).tailMap(2).descendingMap();
this will create 3 calls to a new oakmap that will not be closed ever and the memory will never be free

core/src/main/java/com/oath/oak/ZeroCopyMap.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    Collection<OakRBuffer> values();
+
+    /**
+     * Returns a {@link Set} view of the serialized mappings contained in this
+     * map.
+     *
+     * @return a set view of the serialized mappings contained in this map
+     */
+    Set<Map.Entry<OakRBuffer, OakRBuffer>> entrySet();
```
yonigottesman: In druid i need to have an iterator on entries that has the values as ByteBuffers.
How is this possible in the new API? 
If we dont want to change anything the only way to do this is to add a "getByteBuffer()" function in OakRBuffer

core/src/main/java/com/oath/oak/OakMap.java
eranmeir: I tend to agree with removing the `AutoCloseable` interface. The current implementation is also not idempotent and double closing can cause problems. 

I suggest we discuss this further and maybe change it in a different pull request.

core/src/main/java/com/oath/oak/OakWBuffer.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Absolute <i>get</i> method.  Reads the byte at the given
+     * index.
+     *
+     * @param index The index from which the byte will be read
+     * @return The byte at the given index
+     * @throws IndexOutOfBoundsException If <tt>index</tt> is negative
+     *                                   or not smaller than the buffer's limit
+     */
+    byte get(int index) throws NullPointerException;
```
sanastas: I see both in (R and in W buffers) only absolute get and put methods (getting an index for position). What about relative get and put methods (getting nothing and using internal position)? This should be allowed only in W buffer as it updates the internals, but this is a valid part of ByteBuffer interface. Do I miss something?

README.md
```diff
@@ -127,42 +126,48 @@ OakMap<K,V> oak = builder.build();
...
  	- `void put(K key, V value)`
  	- `boolean putIfAbsent(K key, V value)`
  	- `void remove(K key)`
  	- `boolean computeIfPresent(K key, Consumer<OakWBuffer> computer)`
  	- `void putIfAbsentComputeIfPresent(K key, V value, Consumer<OakWBuffer> computer)`
-2. In contrast to the ConcurrentNavigableMap API, `void put(K key, V value)` does not return the value previously associated with the key, if key existed. Likewise, `void remove(K key)` does not return a boolean indicating whether key was actually deleted, if key existed.
-3. `boolean computeIfPresent(K key, Consumer<OakWBuffer> computer)` gets the user-defined computer function. The computer is invoked in case the key exists.
-The computer is provided with OakWBuffer, representing the serialized value associated with the key. The computer's effect is atomic, meaning that either all updates are seen by concurrent readers, or none are.
+3. In contrast to the ConcurrentNavigableMap API, `void put(K key, V value)` does not return the value previously associated with the key, if key existed. Likewise, `void remove(K key)` does not return a boolean indicating whether key was actually deleted, if key existed.
```
sanastas: Please stress that this is about ZeroCopyMap

README.md
```diff
@@ -127,42 +126,48 @@ OakMap<K,V> oak = builder.build();
...
 	
-	However, these direct methods return keys and/or values as Objects by applying deseriliazation (copy). This is costly,  and we strongly advice to use OakBufferView or OakTransformView to operate directly on the internal data representation.
-3. For further understanding of data retrieval via OakTransformView, please refer to the [Transformations](#transformations) section.
-4. Note that OakMap's iterators are `OakCloseableIterator` (extend AutoCloseable) so it is possible to reuse the memory previously referred by iterator. Be sure to use it within try-statement or call its close() method explicitly when iterator is no longer in use.
+	However, these direct methods return keys and/or values as Objects by applying deseriliazation (copy). This is costly,  and we strongly advice to use ZeroCopyMap to operate directly on the internal data representation.
+3. For examples of direct data manipulations, please refer to the [usage](#usage) section.
 
 ### Notes on data ingestion
-1. Data can be ingested and updated via the following five methods:
+1. Data can be ingested via the standard ConcurrentNavigableMap API.
```
sanastas: What about compute methods given by ConcurrentNavigableMap? Like:

default V computeIfPresent(K key,
                           BiFunction<? super K,? super V,? extends V> remappingFunction)

Do we support it anyhow?

README.md
```diff
@@ -171,48 +176,43 @@ This API looks for a key. If the key does not exist, it adds a new Serialized ke
...
 This memory (the entire given capacity) needs to be released later, thus OakMap implements AutoClosable. Be sure to use it within try-statement or better invoke OakMap's close() method when OakMap is no longer in use.
 
-Please pay attention that multiple views can be defined to the same underlying memory of OakMap. That is when other sub-maps and views (OakBufferView/OakTransformView) are created. Do not worry, the true memory release will happen only when last of those views is closed.
-However, note that each sub-map is in particular an OakMap and thus AutoCloseable and needs to be closed (explicitly or implicitly). Similarly, OakBufferView and OakTransformView are AutoCloseable objects and need to be closed. Again, close() can be invoked on different objects referring to the same underlying memory, but the final release will happen only once.
+Please pay attention that multiple views can be defined to the same underlying memory of OakMap. That is when other sub-maps are created. Do not worry, the true memory release will happen only when last of those views is closed.
```
sanastas: Did you mean ZeroCopyMaps instead of views? There are no views now...

README.md
```diff
@@ -127,42 +126,48 @@ OakMap<K,V> oak = builder.build();
...
 
-An OakRBuffer can represent either a key or a value. The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Notice that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
+An OakRBuffer can represent either a key or a value. The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Note that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
+
+For backward compatibility with applications that are already based on the use of ByteBuffers, Oak Buffers provide the transform method that atomically applies a transformation to the underlying ByteBuffer. For a more comprehensive code example please refer to the [usage](#usage) section.
```
sanastas: It is unclear (neither from here nor from usage section) that transform doesn't mutate the underlying buffer just creates a modified output to the user.

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -489,14 +487,13 @@ void writeValue(int hi, V value) {
      * turn has the value. For linkage this is an insert linearization point.
      * All the relevant data can be found inside opData.
      *
-     * point to value
-     * if unsuccessful this means someone else got to it first (helping rebalancer or other operation)
+     * if someone else got to it first (helping rebalancer or other operation), returns the old handle
      */
-    boolean pointToValue(OpData opData) {
+    Handle pointToValue(OpData opData) {
```
sanastas: I'll get deeper into it once the code is merged with the last (and final) Yoni changes

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -51,11 +51,11 @@ boolean remove(MemoryManager memoryManager) {
         return true;
     }
 
-    void put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
+    boolean put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
         writeLock.lock();
```
sanastas: If not to make Handle synchronization pluggable, let us at least encapsulate the Handle lock into some start/end wrap...

README.md
eranmeir: For non zero-copy operations we sometimes rely on default implementations in the map hierarchy.

Specifically, `computeIfPresent` has a default implementation in `ConcurrentMap` that uses `replace` and `remove`, which we do implement explicitly.

core/src/main/java/com/oath/oak/OakWBuffer.java
eranmeir: @yonigottesman changed the underlying implementation to avoid extra ByteBuffer slicing. The application now cannot assume position is 0 and it cannot modify the position. 

I also updated the README to state OakBuffers expose an API *similar* to the standard Java API.

README.md
eranmeir: > That is when other sub-maps are created.

Views here mean sub maps. Perhaps this should be more clearly indicated. I will change the wording.

core/src/main/java/com/oath/oak/Handle.java
eranmeir: I agree that we should revise our implementation to support pluggable Handle implementations.
I suggest we do it in a different PR, to keep the code review manageable.

PR #32, opened 2019-04-01
Put if absent removerebalne op

yonigottesman:
1. Fix putifAbsent, putIfabsentCompute bugs
2. Remove rebalance helpOp

### DIFF
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -64,8 +64,8 @@ public class Chunk<K, V> {
     private AtomicReference<Rebalancer> rebalancer;
     private final int[] entries;    // array is initialized to 0, i.e., NONE - this is important!
     private final KeysManager<K> keysManager;
-    private final Handle[] handles;
-    private AtomicReferenceArray<OpData> pendingOps;
+    private final Handle<V>[] handles;
+    private AtomicInteger pendingOps;
     private final AtomicInteger entryIndex;    // points to next free index of entry array
     final AtomicInteger keyIndex;    // points to next free index of key array
     private final AtomicInteger handleIndex;    // points to next free index of entry array
@@ -115,7 +115,7 @@ public class Chunk<K, V> {
         this.entryIndex = new AtomicInteger(FIRST_ITEM);
         this.handles = new Handle[maxItems + FIRST_ITEM];
         this.handleIndex = new AtomicInteger(FIRST_ITEM);
-        this.keysManager = new KeysManager<>(this.maxKeyBytes, memoryManager, keySerializer);
+        this.keysManager = new KeysManager<K>(this.maxKeyBytes, memoryManager, keySerializer);
         this.keyIndex = new AtomicInteger(FIRST_ITEM);
         this.sortedCount = new AtomicInteger(0);
         this.minKey = minKey;
@@ -125,7 +125,7 @@ public class Chunk<K, V> {
         else
             this.state = new AtomicReference<>(State.INFANT);
         this.next = new AtomicMarkableReference<>(null, false);
-        this.pendingOps = new AtomicReferenceArray<>(ThreadIndexCalculator.MAX_THREADS);
+        this.pendingOps = new AtomicInteger();
         this.rebalancer = new AtomicReference<>(null); // to be updated on rebalance
         this.statistics = new Statistics();
         this.comparator = comparator;
@@ -182,14 +182,6 @@ public class Chunk<K, V> {
         return unsafe.compareAndSwapInt(entries,
                 Unsafe.ARRAY_INT_BASE_OFFSET + (item + offset) * Unsafe.ARRAY_INT_INDEX_SCALE,
                 expected, value);
-    }
-
-    /**
-     * performs CAS from 'expected' to 'value' for field at specified offset of given item in
-     * pending array
-     */
-    private boolean casPendingArray(int item, OpData expected, OpData opData) {
-        return pendingOps.compareAndSet(item, expected, opData);
     }
 
     /**
@@ -252,7 +244,7 @@ public class Chunk<K, V> {
     /**
      * gets the value for the given item, or 'null' if it doesn't exist
      */
-    Handle getHandle(int entryIndex) {
+    Handle<V> getHandle(int entryIndex) {
 
         int hi = get(entryIndex, OFFSET_HANDLE_INDEX);
 
@@ -300,13 +292,13 @@ public class Chunk<K, V> {
         return null;
     }
 
-    static class LookUp {
+    static class LookUp<V> {
 
-        final Handle handle;
+        final Handle<V> handle;
         final int entryIndex;
         final int handleIndex;
 
-        LookUp(Handle handle, int entryIndex, int handleIndex) {
+        LookUp(Handle<V> handle, int entryIndex, int handleIndex) {
             this.handle = handle;
             this.entryIndex = entryIndex;
             this.handleIndex = handleIndex;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -599,7 +593,7 @@ private void setState(State state) {
     }
 
     void normalize() {
-        setState(State.NORMAL);
+        state.compareAndSet(State.INFANT, State.NORMAL);
```
Eshcar: good catch

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -621,51 +615,17 @@ private int getLastItemEntryIndex() {
         return entryIndex;
     }
 
+
+    public void freeHandle(int handleIndex) {
```
Eshcar: please add documentation - what exactly does this method do?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -621,51 +615,17 @@ private int getLastItemEntryIndex() {
...
+        handles[handleIndex].remove(memoryManager);
+    }
+
     /**
      * freezes chunk so no more changes can be done to it (marks pending items as frozen)
      */
     void freeze() {
         setState(State.FROZEN); // prevent new puts to this chunk
```
Eshcar: should we change this to cas as well? for example if the chunk if released no need to freeze right?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -283,45 +281,17 @@ private void updateIndexAndNormalize(List<Chunk<K, V>> engagedChunks, List<Chunk
...
-    private void rebalanceCompute(Chunk<K, V> c, K key, V value, Consumer<OakWBuffer> computer) {
-        Rebalancer.RebalanceResult result = rebalance(c, key, value, computer, Operation.COMPUTE);
-        assert result != null;
-        if (result.success) { // rebalance helped compute
-            return;
-        }
-        putIfAbsentComputeIfPresent(key, value, computer);
-    }
-
     private boolean rebalanceRemove(Chunk<K, V> c, K key) {
```
Eshcar: why do we keep this method? is it invoked anywhere?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -434,42 +410,51 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
             int prevEi = c.linkEntry(ei, true, key);
             if (prevEi != ei) {
-                Handle handle = c.getHandle(prevEi);
-                if (handle == null) {
-                    ei = prevEi;
-                    prevHi = c.getHandleIndex(prevEi);
-                } else {
+
+                prevHi = c.getHandleIndex(prevEi);
+                if (prevHi != -1 ) {
```
Eshcar: wait - so if you are changing this code does this mean it no longer follows the pseudocode in Hagar's thesis? so how do we know the proofs apply here? there is no straightforward way to "see" the code is correct

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -513,42 +497,49 @@ void putIfAbsentComputeIfPresent(K key, V value, Consumer<OakWBuffer> computer)
...
         c.writeValue(hi, value); // write value in place
 
         Chunk.OpData opData = new Chunk.OpData(Operation.COMPUTE, ei, hi, prevHi, computer);
 
         // publish put
-        if (!c.publish(opData)) {
-            rebalanceCompute(c, key, value, computer);
-            return;
+        if (!c.publish()) {
+            c.freeHandle(hi);
```
Eshcar: what happens if you do not free the handle - this is a memory leak?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -41,7 +41,7 @@
     // number of bytes allocated for this Oak among different Blocks
     // can be calculated, but kept for easy access
     private final AtomicLong allocated = new AtomicLong(0);
-
+    private final AtomicLong freeCounter = new AtomicLong(0);
```
Eshcar: please add documentation - what does it count? number of bytes?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -378,7 +372,7 @@ int allocateHandle() {
         if (hi + 1 > handles.length) {
             return -1;
         }
-        handles[hi] = new Handle();
+        handles[hi] = new Handle<>();
```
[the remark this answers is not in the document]
yonigottesman: whats wrong?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -529,7 +523,7 @@ Handle pointToValue(OpData opData) {
                     return pointToValue(opData);
                 }
             }
-            return null;
+            return h;
```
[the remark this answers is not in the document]
yonigottesman: Eran is right. It became a bug only after putifabsentcompute returns true/false

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: This code is actually closer to Hagars thesis.

core/src/main/java/com/oath/oak/InternalOakMap.java
yonigottesman: yes

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yonigottesman: Its just to make objects in free list unique.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -115,7 +115,7 @@ public ByteBuffer allocate(int size) {
     public void free(ByteBuffer bb) {
         allocated.addAndGet(-(bb.remaining()));
         if (stats != null) stats.release(bb);
-        freeList.add(new Pair<>(System.identityHashCode(bb), bb));
+        freeList.add(new Pair<>(freeCounter.getAndIncrement(), bb));
```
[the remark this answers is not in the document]
yonigottesman: No, there was no bug because every released bb is unique.
I found this was a performance issue, the identityHashCode function is heavy and its faster to just use an atomic counter

core/src/main/java/com/oath/oak/Chunk.java
yonigottesman: I think its pretty clear from the name of the function.

core/src/main/java/com/oath/oak/ZeroCopyMap.java
```diff
@@ -76,13 +76,15 @@
...
      * Else, updates the value for the specified key.
      *
-     * @param key      key with which the specified value is to be associated
-     * @param value    value to be associated with the specified key
-     * @param computer for computing the new value when the key is present
-     * @throws NullPointerException     if any of the parameters is null
+     * @param key         key with which the specified value is to be associated
+     * @param value       value to be associated with the specified key
+     * @param computer    for computing the new value when the key is present
+     * @return {@code true} if there was no mapping for the key
```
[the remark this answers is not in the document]
yonigottesman: I cannot find where in the README.

core/src/test/java/com/oath/oak/PutIfAbsentTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+
+public class PutIfAbsentTest {
+
+
```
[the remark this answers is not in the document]
yonigottesman: You are absolutely right.
But, we dont have the abstractions to mock and hand wire the bug

PR #34, opened 2019-04-28
Add unsafe memcopy

yonigottesman:


yonigottesman:
Must merge this before merge druid changes

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -85,7 +85,6 @@ boolean compute(Consumer<OakWBuffer> computer) {
     }
 
     ByteBuffer getSlicedByteBuffer() {
-        assert writeLock.isHeldByCurrentThread();
```
sanastas: Why to remove this? Is it under a read lock and read lock can not be checked? Please add a comment

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -68,4 +68,9 @@ public double getDouble(int index) {
         return transformer.apply(byteBuffer);
     }
 
+    @Override
+    public void unsafeBufferToIntArrayCopy(int srcPosition, int[] dstArray, int countInts) {
+        UnsafeUtils.unsafeBufferToArrayCopy(byteBuffer, srcPosition, dstArray, countInts);
```
sanastas: What if keys buffer is being copied during rebalance? Is it that source key buffer is immutable anyway? Keys are anyway immutable in current usage... Good to add a comment about that, to show that this was considered

core/src/main/java/com/oath/oak/UnsafeUtils.java
```diff
@@ -0,0 +1,55 @@
...
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
+        }
+        INT_ARRAY_OFFSET = unsafe.arrayBaseOffset(int[].class);
+        BYTE_ARRAY_OFFSET = unsafe.arrayBaseOffset(byte[].class);
+    }
+
+    private UnsafeUtils(){};
+
+    public static void unsafeBufferToArrayCopy(ByteBuffer srcByteBuffer, int position, int[] dstArray, int countInts) {
```
sanastas: Better to mention in the name that this is for int array and for any array

core/src/main/java/com/oath/oak/OakRBuffer.java
```diff
@@ -135,4 +135,12 @@
...
      */
     <T> T transform(Function<ByteBuffer, T> transformer);
+
+    /**
+     * Perform an unsafe memcopy from the underlying buffer to the dst int array.
+     * @param srcPosition The position in the underlying buffer.
+     * @param dstArray The int array to copy to.
+     * @param countInts The number of ints to copy.
+     */
+    void unsafeBufferToIntArrayCopy(int srcPosition, int[] dstArray, int countInts);
```
Eshcar: So this is a new api method to be used by the application? and you are using it in Druid to improve performance?
If so, it would be helpful to add a few sentences on when it is advisable to use this method. Either here or in the class documentation, or in both places.

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
yonigottesman: keys are immutable and the current mechanism for releasing a key is only when no references so no problem

core/src/main/java/com/oath/oak/UnsafeUtils.java
```diff
@@ -0,0 +1,55 @@
...
+            unsafeConstructor.setAccessible(true);
+            unsafe = unsafeConstructor.newInstance();
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
+        }
+        INT_ARRAY_OFFSET = unsafe.arrayBaseOffset(int[].class);
+        BYTE_ARRAY_OFFSET = unsafe.arrayBaseOffset(byte[].class);
+    }
+
+    private UnsafeUtils(){};
```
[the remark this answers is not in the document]
yonigottesman: effective java item 4 (you have until the end of the week before they kill safari for us)

core/src/main/java/com/oath/oak/OakRBuffer.java
yonigottesman: Currently used in druid. hasn't shown any major performance impact on the benchmarks we are running

PR #35, opened 2019-05-29
Netty allocator

yonigottesman:
**Issue #, if available:**

**Description of changes:**

I confirm that this contribution is made under the terms of your choice and that I have the authority necessary to make this contribution on behalf of its copyright owner.


core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -58,11 +61,11 @@ boolean put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager)
             return false;
         }
         int capacity = serializer.calculateSize(newVal);
-        if (this.value.remaining() < capacity) { // can not reuse the existing space
+        if (this.value.capacity() < capacity) { // can not reuse the existing space
             memoryManager.release(this.value);
             this.value = memoryManager.allocate(capacity);
         }
-        serializer.serialize(newVal, this.value.slice());
+        serializer.serialize(newVal, this.value.nioBuffer());
```
sanastas: I wonder about the usage of nioBuffer()... I wonder if this is what we want, and do we follow the original intent if switch to such usage...

This is from its description:

public abstract java.nio.ByteBuffer nioBuffer()
Exposes this buffer's readable bytes as an NIO ByteBuffer. The returned buffer either share or contains the copied content of this buffer, while changing the position and limit of the returned NIO buffer does not affect the indexes and marks of this buffer. This method is identical to buf.nioBuffer(buf.readerIndex(), buf.readableBytes()). This method does not modify readerIndex or writerIndex of this buffer. Please note that the returned NIO buffer will not see the changes of this buffer if this buffer is a dynamic buffer and it adjusted its capacity.
Throws:
java.lang.UnsupportedOperationException - if this buffer cannot create a ByteBuffer that shares the content with itself

core/src/main/java/com/oath/oak/MemoryManager.java
```diff
@@ -8,32 +8,30 @@
...
 
     public void close() {
-        valuesMemoryAllocator.close();
         keysMemoryAllocator.close();
     }
 
-    void release(ByteBuffer bb) {
-        valuesMemoryAllocator.free(bb);
+    void release(ByteBuf bb) {
+        bb.release();
```
sanastas: First, the allocation is done via allocator while release/free not. Second, the release is conducted under reference counting, and it looks like we do not increase the reference count under each access, so again I wonder if it is the correct usage...

boolean release()
Decreases the reference count by 1 and deallocates this object if the reference count reaches at 0.
Returns:
true if and only if the reference count became 0 and this object has been deallocated

https://netty.io/4.0/api/io/netty/util/ReferenceCounted.html

yonigottesman: Yes this is the way to use it. when the refcount gets to 0 the bytebuf is automatically released back to the pool. Its the user responsibility to decrease the refcount using the release function.
I dont think we currently need to increase refcount on each access because we have the handle lock.

core/src/main/java/com/oath/oak/Handle.java
yonigottesman: Take a look at the description here: https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html
You are write that this returns a sliced bytebuffer from readerIndex to writerIndex, thats why in line 37 i do value.writerIndex(value.capacity()); to move the writerIndex to the end.

sanastas: Why do I think that usage of nioBuffer() might not be correct/efficient is from the following parts of the description:

Exposes this buffer's **readable bytes** as an NIO ByteBuffer. The returned buffer either share **or contains the copied content** of this buffer, **while changing the position and limit of the returned NIO buffer does not affect the indexes and marks of this buffer**.

1. "readable bytes" what if nothing was yet written into this buffer and readable bytes are actually nothing?
2. "contains the copied content" if it goes to copy the content (even in some specific cases) it is not good solution for us
3. "changing the position and limit of the returned NIO buffer does not affect the indexes and marks of this buffer" it looks like the returned buffer was intend to be read-only and if we let user write into it (and we do), the write index will not be updated and user might not see the changes on next read...

PR #42, opened 2019-09-01
Bug fixes

eranmeir:
**UPDATED 2019-09-23**
- Bug fixes
  - putIfAbsent may return null instead of value for existing keys
  - Logical deletion in remove was not marked properly
  - Fix Non-ZC concurrent put
    - Move desrialization to an atomic value exchange method in Handle
  - Fix Non-ZC concurrent removals
    - Move desrialization to an atomic remove in Handle

- Other improvements
  - Some improvements to type safety in InternalOakMap and relatives (mostly trying to reduce fallout from mixed skiplist comparator and make comparator usage safer)
  - Refactor Integer OakMap reference implementation to a new class so that serializer and comparator can be reused
  - Pulled up Result class to top level so that it can be used in Handle as well
  - Fixed: OffHeapOakTest fails silently because JUnit doesn't fail on thread exceptions
    - Fail test when exceptions occur in threads

---

### DIFF
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -10,7 +10,6 @@ import sun.misc.Unsafe;
 
 import java.lang.reflect.Constructor;
 import java.nio.ByteBuffer;
-import java.util.Comparator;
 import java.util.EmptyStackException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -53,7 +52,7 @@ public class Chunk<K, V> {
     private final MemoryManager memoryManager;
     ByteBuffer minKey;       // minimal key that can be put in this chunk
     AtomicMarkableReference<Chunk<K, V>> next;
-    Comparator<Object> comparator;
+    OakComparator<K> comparator;
 
     // in split/compact process, represents parent of split (can be null!)
     private AtomicReference<Chunk<K, V>> creator;
@@ -94,7 +93,7 @@ public class Chunk<K, V> {
      * @param minKey  minimal key to be placed in chunk
      * @param creator the chunk that is responsible for this chunk creation
      */
-    Chunk(ByteBuffer minKey, Chunk<K, V> creator, Comparator<Object> comparator, MemoryManager memoryManager,
+    Chunk(ByteBuffer minKey, Chunk<K, V> creator, OakComparator<K> comparator, MemoryManager memoryManager,
           int maxItems, AtomicInteger externalSize,
           OakSerializer<K> keySerializer, OakSerializer<V> valueSerializer) {
         this.memoryManager = memoryManager;
@@ -151,13 +150,6 @@ public class Chunk<K, V> {
     void release() {
         // try to change the state
         state.compareAndSet(State.FROZEN, State.RELEASED);
-    }
-
-    /**
-     * compares ByteBuffer by calling the provided comparator
-     */
-    private int compare(Object k1, Object k2) {
-        return comparator.compare(k1, k2);
     }
 
     /**
@@ -320,7 +312,7 @@ public class Chunk<K, V> {
         // iterate until end of list (or key is found)
         while (curr != NONE) {
             // compare current item's key to searched key
-            cmp = compare(readKey(curr), key);
+            cmp = comparator.compareSerializedKeyAndKey(readKey(curr), key);
             // if item's key is larger - we've exceeded our key
             // it's not in chunk - no need to search further
             if (cmp > 0) {
@@ -370,14 +362,13 @@ public class Chunk<K, V> {
         int sortedCount = this.sortedCount.get();
         // if there are no sorted keys, or the first item is already larger than key -
         // return the head node for a regular linear search
-        if ((sortedCount == 0) || compare(readKey(FIRST_ITEM), key) >= 0) {
+        if ((sortedCount == 0) || comparator.compareSerializedKeyAndKey(readKey(FIRST_ITEM), key) >= 0) {
             return HEAD_NODE;
         }
 
         // optimization: compare with last key to avoid binary search
-        if (compare(readKey((sortedCount - 1) * FIELDS + FIRST_ITEM), key) < 0) {
+        if (comparator.compareSerializedKeyAndKey(readKey((sortedCount - 1) * FIELDS + FIRST_ITEM), key) < 0)
             return (sortedCount - 1) * FIELDS + FIRST_ITEM;
-        }
 
         int start = 0;
         int end = sortedCount;
@@ -385,7 +376,7 @@ public class Chunk<K, V> {
         while (end - start > 1) {
             int curr = start + (end - start) / 2;
 
-            if (compare(readKey(curr * FIELDS + FIRST_ITEM), key) >= 0) {
+            if (comparator.compareSerializedKeyAndKey(readKey(curr * FIELDS + FIRST_ITEM), key) >= 0) {
                 end = curr;
             } else {
                 start = curr;
@@ -445,7 +436,7 @@ public class Chunk<K, V> {
         return ei;
     }
 
-    int linkEntry(int ei, boolean cas, K key) {
+    int linkEntry(int ei, K key) {
         int prev, curr, cmp;
         int anchor = -1;
 
@@ -466,7 +457,7 @@ public class Chunk<K, V> {
                     break;
                 }
                 // compare current item's key to ours
-                cmp = compare(readKey(curr), key);
+                cmp = comparator.compareSerializedKeyAndKey(readKey(curr), key);
 
                 // if current item's key is larger, done searching - add between prev and curr
                 if (cmp > 0) {
@@ -482,33 +473,28 @@ public class Chunk<K, V> {
             // link to list between next and previous
             // first change this key's next to point to curr
             setEntryField(ei, OFFSET_NEXT, curr); // no need for CAS since put is not even published yet
-            if (cas) {
-                if (casEntriesArray(prev, OFFSET_NEXT, curr, ei)) {
-                    // Here is the single place where we do enter a new entry to the chunk, meaning
-                    // there is none else simultaneously inserting the same key
-                    // (we were the first to insert this key).
-                    // If the new entry's index is exactly after the sorted count and
-                    // the entry's key is greater or equal then to the previous (sorted count)
-                    // index key. Then increase the sorted count.
-                    int sortedCount = this.sortedCount.get();
-                    if (sortedCount > 0) {
-                        if (ei == (sortedCount * FIELDS + 1)) {
-                            // the new entry's index is exactly after the sorted count
-                            if (compare(
-                                    readKey((sortedCount - 1) * FIELDS + FIRST_ITEM), key) <= 0) {
-                                // compare with sorted count key, if inserting the "if-statement",
-                                // the sorted count key is less or equal to the key just inserted
-                                this.sortedCount.compareAndSet(sortedCount, (sortedCount + 1));
-                            }
+            if (casEntriesArray(prev, OFFSET_NEXT, curr, ei)) {
+                // Here is the single place where we do enter a new entry to the chunk, meaning
+                // there is none else simultaneously inserting the same key
+                // (we were the first to insert this key).
+                // If the new entry's index is exactly after the sorted count and
+                // the entry's key is greater or equal then to the previous (sorted count)
+                // index key. Then increase the sorted count.
+                int sortedCount = this.sortedCount.get();
+                if (sortedCount > 0) {
+                    if (ei == (sortedCount * FIELDS + 1)) {
+                        // the new entry's index is exactly after the sorted count
+                        if (comparator.compareSerializedKeyAndKey(
+                                readKey((sortedCount - 1) * FIELDS + FIRST_ITEM), key) <= 0) {
+                            // compare with sorted count key, if inserting the "if-statement",
+                            // the sorted count key is less or equal to the key just inserted
+                            this.sortedCount.compareAndSet(sortedCount, (sortedCount + 1));
                         }
                     }
-                    return ei;
                 }
-                // CAS didn't succeed, try again
-            } else {
-                // without CAS (used by rebalance)
-                setEntryField(prev, OFFSET_NEXT, ei);
+                return ei;
             }
+            // CAS didn't succeed, try again
         }
     }
 
@@ -523,7 +509,7 @@ public class Chunk<K, V> {
         handles[hi].setValue(byteBuffer);
     }
 
-    public int getMaxItems() {
+    int getMaxItems() {
         return maxItems;
     }
 
@@ -537,7 +523,7 @@ public class Chunk<K, V> {
     Handle pointToValue(OpData opData) {
 
         // try to perform the CAS according to operation data (opData)
-        if (pointToValueCAS(opData, true)) {
+        if (pointToValueCAS(opData)) {
             return null;
         }
 
@@ -545,7 +531,8 @@ public class Chunk<K, V> {
         Operation operation = opData.op;
 
         // the operation is remove, means we tried to change the handle index we knew about to -1
-        // the old handle index is no longer there so we have nothing to do
+        // the old handle index is no longer there so we have nothing to do. Note that in case of non-ZC removal the
+        // loop in remove() will lookup the key again so we will still return the previous value
         if (operation == Operation.REMOVE) {
             return null; // this is a remove, no need to try again and return doesn't matter
         }
@@ -584,21 +571,17 @@ public class Chunk<K, V> {
     /**
      * used by put/putIfAbsent/remove and rebalancer
      */
-    private boolean pointToValueCAS(OpData opData, boolean cas) {
-        if (cas) {
-            if (casEntriesArray(opData.entryIndex, OFFSET_HANDLE_INDEX, opData.prevHandleIndex, opData.handleIndex)) {
-                // update statistics only by thread that CASed
-                if (opData.prevHandleIndex < 0 && opData.handleIndex > 0) { // previously a remove
-                    statistics.incrementAddedCount();
-                    externalSize.incrementAndGet();
-                } else if (opData.prevHandleIndex > 0 && opData.handleIndex == -1) { // removing
-                    statistics.decrementAddedCount();
-                    externalSize.decrementAndGet();
-                }
-                return true;
+    private boolean pointToValueCAS(OpData opData) {
+        if (casEntriesArray(opData.entryIndex, OFFSET_HANDLE_INDEX, opData.prevHandleIndex, opData.handleIndex)) {
+            // update statistics only by thread that CASed
+            if (opData.prevHandleIndex < 0 && opData.handleIndex > 0) { // previously a remove
+                statistics.incrementAddedCount();
+                externalSize.incrementAndGet();
+            } else if (opData.prevHandleIndex > 0 && opData.handleIndex == -1) { // removing
+                statistics.decrementAddedCount();
+                externalSize.decrementAndGet();
             }
-        } else {
-            setEntryField(opData.entryIndex, OFFSET_HANDLE_INDEX, opData.handleIndex);
+            return true;
         }
         return false;
     }
@@ -667,8 +650,8 @@ public class Chunk<K, V> {
     }
 
 
-    public void freeHandle(int handleIndex) {
-        handles[handleIndex].remove(memoryManager);
+    void freeHandle(int handleIndex) {
+        handles[handleIndex].remove(memoryManager, null, null);
         handles[handleIndex] = null;
     }
 
@@ -688,7 +671,7 @@ public class Chunk<K, V> {
      * @param maxCapacity -- max number of entries "this" chunk can contain after copy
      * @return key index of next to the last copied item, NONE if all items were copied
      */
-    final int copyPartNoKeys(Chunk srcChunk, int srcEntryIdx, int maxCapacity) {
+    final int copyPartNoKeys(Chunk<K, V> srcChunk, int srcEntryIdx, int maxCapacity) {
 
         if (srcEntryIdx == HEAD_NODE) {
             return NONE;
@@ -849,14 +832,6 @@ public class Chunk<K, V> {
                 (numOfEntries * MAX_IDLE_ENTRIES_FACTOR > maxItems && numOfItems * MAX_IDLE_ENTRIES_FACTOR < numOfEntries);
     }
 
-    int getNumOfItems() {
-        return statistics.getCompactedCount();
-    }
-
-    int getNumOfEntries() {
-        return entryIndex.get() / FIELDS;
-    }
-
     /*-------------- Iterators --------------*/
 
     AscendingIter ascendingIter() {
@@ -900,9 +875,8 @@ public class Chunk<K, V> {
             int handle = getEntryField(next, OFFSET_HANDLE_INDEX);
 
             int compare = -1;
-            if (next != Chunk.NONE) {
-                compare = compare(from, readKey(next));
-            }
+            if (next != Chunk.NONE)
+                compare = (-1) * comparator.compareSerializedKeyAndKey(readKey(next), from);
 
             while (next != Chunk.NONE &&
                     (compare > 0 ||
@@ -910,9 +884,8 @@ public class Chunk<K, V> {
                             handle < 0)) {
                 next = getEntryField(next, OFFSET_NEXT);
                 handle = getEntryField(next, OFFSET_HANDLE_INDEX);
-                if (next != Chunk.NONE) {
-                    compare = compare(from, readKey(next));
-                }
+                if (next != Chunk.NONE)
+                    compare = (-1) * comparator.compareSerializedKeyAndKey(readKey(next), from);
             }
         }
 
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -529,9 +531,7 @@ Handle pointToValue(OpData opData) {
         int expectedHandleIdx = opData.handleIndex;
         int foundHandleIdx = getEntryField(opData.entryIndex, OFFSET_HANDLE_INDEX);
 
-        if (foundHandleIdx == expectedHandleIdx) {
-            return null; // someone helped
-        } else if (foundHandleIdx < 0) {
+        if (foundHandleIdx < 0) {
```
sanastas: I am not sure it is correct to remove the comparison and check that someone else helped so the found handle index value is as expected. The following checks like "else if (operation == Operation.PUT_IF_ABSENT)" assume the found handle index is different. Also it doesn't looks correct to recursively call for pointToValue() after all in this case...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -399,13 +399,17 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
 
         int ei = -1;
         int prevHi = -1;
+
+        // Is there already an entry associated with this key?
         if (lookUp != null) {
+            // There's an entry for this key, but it isn't linked to any handle (in which case prevHi will be < 0) or it's linked to a deleted handle that will then be indexed in prevHi (which will be > 0)
```
sanastas: too long line

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -546,36 +557,39 @@ V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
...
             }
 
-            if (logical) {
-                // we have marked this handle as deleted (successful remove)
+            if (logicallyDeleted) {
+                // This is the case where we logically deleted this entry (marked the handle as deleted), but someone reused
+                // the entry before we unlinked it. We have the previous value saved in v.
+                return v;
+            } else {
                 V vv = (transformer != null) ? (V) lookUp.handle.transform(transformer) : null;
```
sanastas: How do you know the value wouldn't be changed just after this line and before marked as deleted?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -546,36 +557,39 @@ V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
...
 
-                if (oldValue != null && !oldValue.equals(vv))
+                if (oldValue != null && !oldValue.equals(vv)) {
+                    // this is not the droid you're looking for
```
sanastas: What is "droid"?

core/src/main/java/com/oath/oak/InternalOakMap.java
sanastas: It can be changed also after oldValue equals vv check...

eranmeir:
These changes were made before we discussed fixing other, but related, linearization issues. I'm closing this PR as a more thorough fix is called for.

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -381,6 +381,7 @@ private int binaryFind(K key) {
      * @return result of CAS
      **/
     boolean publish() {
+        // TODO (eranmeir) we should probably rename this. It isn't really publishing, more like sempahore acquisition
```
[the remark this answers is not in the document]
eranmeir: removed

core/src/main/java/com/oath/oak/Chunk.java
eranmeir: reverted

core/src/main/java/com/oath/oak/InternalOakMap.java
eranmeir: https://youtu.be/532j-186xEQ?t=37

sanastas: Such an illiteracy and an incomprehension on my side! :)

eranmeir:
I just updated (force pushed) this PR so that it is aligned with recent changes @yoavz1997 included. Since this change list is getting quite large, this PR currently only fixes non-ZC put and remove. 

Note that it also partially addresses #52 - not by fixing the test, but at least making it propagate failure properly.

@sanastas @Eshcar would appreciate if you took another look

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -39,16 +39,18 @@ boolean isDeleted() {
         return deleted.get();
     }
 
-    boolean remove(MemoryManager memoryManager) {
+    <V> Result<V> remove(MemoryManager memoryManager, V oldValue, Function<ByteBuffer, V> transformer) {
         writeLock.lock();
-        if (isDeleted()) {
+        V vv = (transformer != null) ? transformer.apply(value) : null;
```
yoavz1997: Shouldn't the read be only after the deleted bit is checked?
It seems otherwise that you might read someone else's value, which may throw IndexOutOfBoundsException.

core/src/main/java/com/oath/oak/Rebalancer.java
```diff
@@ -16,14 +16,16 @@
 
     /*-------------- Constants --------------*/
 
-    private final int rebalanceSize;
-    private final double maxAfterMergePart;
-    private final double lowThreshold;
-    private final double appendThreshold;
+    private final int rebalanceSize = 2;
```
Eshcar: should these parameters be static and with UPPER CASE?

core/src/main/java/com/oath/oak/Result.java
```diff
@@ -0,0 +1,23 @@
+package com.oath.oak;
+
+class Result<V> {
```
Eshcar: class documentation please.
What does each boolean attribute represent?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -351,11 +366,14 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-            put(key, value, transformer);
-            return null;
+            return put(key, value, transformer);
         }
 
-        finishAfterPublishing(opData, c);
+        Handle oldHandle = finishAfterPublishing(opData, c);
+
+        if (oldHandle != null) {
+            return transformer != null ? oldHandle.transform(transformer) : null;
```
yoavz1997: There is still an atomicity problem here, since the value in oldHandle can be deleted meanwhile.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -541,38 +567,33 @@ V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
             throw new NullPointerException();
         }
 
-        boolean logical = true; // when logical is false, means we have marked the handle as deleted
-        Handle prev = null;
+        // when logicallyDeleted is true, it means we have marked the handle as deleted. Note that the entry may still be linked!
```
sanastas: Better to say, the entry will remain forever linked until rebalance happens. Remove doesn't disconnect entries from list

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -541,38 +567,33 @@ V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
...
+        boolean logicallyDeleted = false;
         V v = null;
 
         while (true) {
-
-            Chunk<K, V> c = findChunk(key); // find chunk matching key
+            Chunk c = findChunk(key); // find chunk matching key
```
yoavz1997: Chunk<K, V>

core/src/test/java/com/oath/oak/OffHeapOakTest.java
```diff
@@ -52,6 +56,7 @@ public void testThreads() throws InterruptedException {
         for (int i = 0; i < NUM_THREADS; i++) {
             threads.get(i).join();
         }
+        assertNull(threadException);
```
yoavz1997: Threads should be able to get ConcurrentModificationException in this test, since there are concurrent removals, iterations and rebalances.

yoavz1997: Other exceptions, on the other hand, should not be thrown.
Maybe check `assert threadException == null || threadException isinstanceof ConcurrentModificationException;`.

core/src/main/java/com/oath/oak/Handle.java
eranmeir: You're right, will fix

core/src/main/java/com/oath/oak/InternalOakMap.java
eranmeir: You're right, will fix

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -900,19 +875,17 @@ DescendingIter descendingIter(K from, boolean inclusive) {
             int handle = getEntryField(next, OFFSET_HANDLE_INDEX);
 
             int compare = -1;
-            if (next != Chunk.NONE) {
-                compare = compare(from, readKey(next));
-            }
+            if (next != Chunk.NONE)
+                compare = (-1) * comparator.compareSerializedKeyAndKey(readKey(next), from);
```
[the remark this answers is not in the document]
eranmeir: The reason is that the comparator for serialized key and (non-serialized) key has reverse order on the arguments than the order used here (you can notice the flipped calling order).
Note that the original comparator already has the sign flip multiplication in it, so this is the only way to negate it. 
As a side note, I think the comparator trick is unsafe and we should see if it can be refactored outside the scope of this PR.

core/src/main/java/com/oath/oak/IntegerOakMap.java
```diff
@@ -0,0 +1,56 @@
+package com.oath.oak;
+
+import java.nio.ByteBuffer;
+
+public class IntegerOakMap {
+
```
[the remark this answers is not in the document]
eranmeir: Yes. I refactored this out instead of justing returning an opaque `defaultBuilder` so that integer serializer and comparator code can be re-used.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -43,22 +43,43 @@
...
 
-        this.comparator = comparator;
+        this.comparator = oakComparator;
 
         this.minKey = ByteBuffer.allocate(this.keySerializer.calculateSize(minKey));
         this.minKey.position(0);
         this.keySerializer.serialize(minKey, this.minKey);
 
-        this.skiplist = new ConcurrentSkipListMap<>(this.comparator);
+        // This is a trick for letting us search through the skiplist using both serialized and unserialized keys.
```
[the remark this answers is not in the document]
eranmeir: We use this runtime coercion and type checking to allow comparison of both `K`-typed keys and serialized `ByteBuffer` keys. In our own code we know when to use which comparison, but for library skiplist, that is designed to compare a single key type, we need this trick to compare both.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -294,9 +315,7 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
         Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
         if (lookUp != null && lookUp.handle != null) {
-            V v = (transformer != null) ? lookUp.handle.transform(transformer) : null;
-            lookUp.handle.put(value, valueSerializer, memoryManager);
-            return v;
+            return lookUp.handle.exchange(value, transformer, valueSerializer, memoryManager);
```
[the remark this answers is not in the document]
eranmeir: bug fix

core/src/main/java/com/oath/oak/Rebalancer.java
eranmeir: Will change

core/src/main/java/com/oath/oak/Result.java
eranmeir: Will update documentation. This is basically a simple implementation of a sum type that can either hold a boolean value or a `V` value

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -436,14 +458,18 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
         }
 
         Handle oldHandle = finishAfterPublishing(opData, c);
+
+        // Keep result before freeing old handle (otherwise we may incorrectly return null for a deleted handle)
+        Result<V> result;
+        if (transformer == null)
+            result = Result.withFlag(oldHandle == null);
+        else
+            result = Result.withValue((oldHandle != null) ? oldHandle.transform(transformer) : null);
```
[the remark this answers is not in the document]
eranmeir: as I noted, this commit addresses only `put` and `remove`. `putIsAbsent` fix is not yet part of this commit.

core/src/main/java/com/oath/oak/InternalOakMap.java
eranmeir: You're right, will fix

core/src/test/java/com/oath/oak/OffHeapOakTest.java
eranmeir: I didn't actually fix the test, just made it so that it won't fail silently. 
@sanastas , if the fix is trivial please let me know and I'll include it

core/src/test/java/com/oath/oak/SingleThreadIteratorTest.java
```diff
@@ -25,7 +25,8 @@
 
     @Before
     public void init() {
-        OakMapBuilder<Integer, Integer> builder = OakMapBuilder.getDefaultBuilder()
+        int maxBytesPerChunkItem = 100;
```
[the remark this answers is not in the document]
eranmeir: You're right, will fix

core/src/test/java/com/oath/oak/SingleThreadTest.java
```diff
@@ -22,7 +22,8 @@
 
     @Before
     public void init() {
-        OakMapBuilder<Integer, Integer> builder = OakMapBuilder.getDefaultBuilder()
+        int maxBytesPerChunkItem = Integer.BYTES;
```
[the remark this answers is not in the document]
eranmeir: You're right, will fix

core/src/main/java/com/oath/oak/IntegerOakMap.java
```diff
@@ -0,0 +1,56 @@
+package com.oath.oak;
```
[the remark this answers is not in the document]
eranmeir: The "default" builder was part of the core package. I'll move this to be part of the test package.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -43,22 +43,43 @@
...
+        this.comparator = oakComparator;
 
         this.minKey = ByteBuffer.allocate(this.keySerializer.calculateSize(minKey));
         this.minKey.position(0);
         this.keySerializer.serialize(minKey, this.minKey);
 
-        this.skiplist = new ConcurrentSkipListMap<>(this.comparator);
+        // This is a trick for letting us search through the skiplist using both serialized and unserialized keys.
+        // Might be nicer to replace it with a proper visitor
+        Comparator mixedKeyComparator = (o1, o2) -> {
```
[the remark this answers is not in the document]
eranmeir: I didn't change the comparator, just moved it down to `InternalOakMap` and tried to keep the non-typesafe usage to a minimum.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -204,7 +225,7 @@ private void connectToChunkList(List<Chunk<K, V>> engaged, List<Chunk<K, V>> chi
                 break;
             }
             // chunk is not in list (someone else already updated list), so we're done with this part
-            if ((curr == null) || (prev == null)) {
+            if (curr == null) {
```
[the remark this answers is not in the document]
eranmeir: if `prev` was initialized with null (line 209) we would search from head (lines 212-220). At this point, either `curr` is `firstEngaged` and `prev` is null (because the while loop in line 216 never happened), and the if clause in line 223 holds, or curr is necessarily not null.

Also, what @yoavz1997 said: this is an improvement suggested by IntelliJ static analyzer.

core/src/main/java/com/oath/oak/InternalOakMap.java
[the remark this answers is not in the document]
eranmeir: You're right, will fix

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -390,22 +408,25 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
+        // Is there already an entry associated with this key?
         if (lookUp != null) {
-            assert lookUp.handle == null;
+            // There's an entry for this key, but it isn't linked to any handle (in which case prevHi will be < 0) or it's linked to a deleted handle that will then be indexed in prevHi (which will be > 0)
             ei = lookUp.entryIndex;
             assert ei > 0;
             prevHi = lookUp.handleIndex;
         }
 
+        // TODO - this can be the else clause of the previous if
```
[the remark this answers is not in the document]
eranmeir: done

core/src/main/java/com/oath/oak/InternalOakMap.java
eranmeir: fixed the comment

core/src/test/java/com/oath/oak/OffHeapOakTest.java
eranmeir: Since the fix requires a deeper dive into the test to see which exceptions are expected and which aren't, I'm leaving issue #52 open until the test and/or relevant code is actually fixed

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -534,38 +534,34 @@ Handle pointToValue(OpData opData) {
...
+                // retry to make a put and to attach the new handle
+                opData.prevHandleIndex = foundHandleIdx;
+                return pointToValue(opData);
             }
-            return h;
         }
-        // this is a put, try again
-        opData.prevHandleIndex = foundHandleIdx;
-        return pointToValue(opData);
+        // this is a put, try again (restart put)
```
yoavz1997: Maybe we should remove this comment since this particular `return false` also serves for putIfAbsentComputeIfPresent.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -457,19 +464,11 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-
-        // Keep result before freeing old handle (otherwise we may incorrectly return null for a deleted handle)
-        Result<V> result;
-        if (transformer == null)
-            result = Result.withFlag(oldHandle == null);
-        else
-            result = Result.withValue((oldHandle != null) ? oldHandle.transform(transformer) : null);
-
-        if (oldHandle != null) {
-            c.freeHandle(hi);
```
yoavz1997: Handle should still be freed before restarting.

core/src/main/java/com/oath/oak/Chunk.java
sanastas: I prefer to leave this comment here, because for PIACIP it is also for the case of possible put. May be this comment can be enhanced saying "this is put or putIfAbsentComputeIfPresent, try ..."

core/src/main/java/com/oath/oak/Chunk.java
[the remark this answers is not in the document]
sanastas: Can we please have a comment here (about the reason for -1)?

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -39,16 +39,24 @@ boolean isDeleted() {
...
     }
 
-    boolean remove(MemoryManager memoryManager) {
+    <V> Result<V> remove(MemoryManager memoryManager, V oldValue, Function<ByteBuffer, V> transformer) {
+        Result<V> res = Result.withFlag(false);
         writeLock.lock();
-        if (isDeleted()) {
+        try {
+            if (!isDeleted()) {
+                V vv = (transformer != null) ? transformer.apply(value) : null;
```
sanastas: Not a necessary, but still... Can there be an "empty transformer" that does nothing and returns null for ZC case? This way we may avoid countless checks for transformer being null...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -326,21 +351,22 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-                put(key, value, transformer);
-                return null;
+                return put(key, value, transformer);
             }
-            int prevEi = c.linkEntry(ei, true, key);
+            int prevEi = c.linkEntry(ei, key);
             if (prevEi != ei) {
                 ei = prevEi;
                 prevHi = c.getHandleIndex(prevEi);
+                if (prevHi != -1) {
```
sanastas: Please add a comment here, or at the beginning of the put, that we can swap the pointer to handle from one value to another only when we sure that previous handle was deleted (or its index was -1). This is an important invariant...

core/src/main/java/com/oath/oak/Handle.java
eranmeir: That's a good idea, although the checks are there because we return different *types* (`Result.withValue` vs `Result.withFlag`). A different abstraction like subclassing deeper for ZC/nonZC cases could fit.  I refrained from further abstraction to limit the size of this change and possible impact on performance.

core/src/main/java/com/oath/oak/Chunk.java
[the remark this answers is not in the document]
eranmeir: switched argument order and added a comment inside the mixedComparator to clarify things

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -370,22 +362,21 @@ private int binaryFind(K key) {
         int sortedCount = this.sortedCount.get();
         // if there are no sorted keys, or the first item is already larger than key -
         // return the head node for a regular linear search
-        if ((sortedCount == 0) || compare(readKey(FIRST_ITEM), key) >= 0) {
+        if ((sortedCount == 0) || comparator.compareKeyAndSerializedKey(key, readKey(FIRST_ITEM)) <= 0) {
```
sanastas: The opposite of ">=" is "<" and not "<=". Also in couple of lines bellow "<" was changed to ">", while the opposite of "<" is ">=". Are we sure that the condition of keys being equal are treated correctly here?

sanastas: Please verify it in all lines where comparator operands places where changes. As I see it in other lines also, just do not comment there as well...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -204,7 +226,7 @@ private void connectToChunkList(List<Chunk<K, V>> engaged, List<Chunk<K, V>> chi
                 break;
             }
             // chunk is not in list (someone else already updated list), so we're done with this part
-            if ((curr == null) || (prev == null)) {
+            if (curr == null) {
                 //TODO Never reached
```
sanastas: If already making all those changes. May be to remove this comment as well. If it would be so, ItelliJ would show it for sure...

core/src/test/java/com/oath/oak/OffHeapOakTest.java
```diff
@@ -16,21 +16,25 @@
...
     private final int NUM_THREADS = 31;
     private ArrayList<Thread> threads;
     private CountDownLatch latch;
     private int maxItemsPerChunk = 248;
+    private Exception threadException;
 
     @Before
     public void init() {
-        OakMapBuilder<Integer, Integer> builder = OakMapBuilder.getDefaultBuilder()
+        int maxBytesPerChunkItem = 100;
```
sanastas: where is it used?

sanastas:
I went through entire PR and do not have other comments.

core/src/main/java/com/oath/oak/Chunk.java
eranmeir: Note that we're not looking for the opposite condition. We still check the same semantic:  "Is the first item larger than or equals key". It's just that the sign is flipped because of argument order.

To make sure, I also tried running tests after replacing with exact opposites like you suggest and multiple tests fail.

core/src/main/java/com/oath/oak/InternalOakMap.java
eranmeir: removed comment

core/src/test/java/com/oath/oak/OffHeapOakTest.java
eranmeir: removed

core/src/main/java/com/oath/oak/Chunk.java
sanastas: OK, I agree

Issue #43, opened 2019-09-04
Relying on Reentrancy


yoavz1997:
https://github.com/yahoo/Oak/blob/d3691702b54010ff0b67ca9b3a3190dc9b903b64/core/src/main/java/com/oath/oak/InternalOakMap.java#L739-L756

https://github.com/yahoo/Oak/blob/d3691702b54010ff0b67ca9b3a3190dc9b903b64/core/src/main/java/com/oath/oak/Handle.java#L205-L218

In the replace functions, the write lock is acquired once when calling Handle::mutatingFunction and another time when calling Handle::put inside.
It works in the production code because the lock is reentrant, however, this assumption is limiting.

PR #44, opened 2019-09-04
Bug fix of issue #43

yoavz1997:
This is a fix to issue #43.
The changes include implementing a replacement to Handle::mutatingFunction (Handle::exchange and Handle::compareExchange) to acquire the writeLock only once, and not twice.
Also, most of the generics warnings are resolved (Handle and LookUp are not generic anymore).


core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -196,27 +200,33 @@ public ByteOrder order() {
...
+    <V> V exchange(V newValue, Function<ByteBuffer, V> valueDeserializeTransformer, OakSerializer<V> serializer, MemoryManager memoryManager) {
         try {
             writeLock.lock();
-            if (isDeleted()) {
-                // finally clause will handle unlock
+            if (isDeleted())
                 return null;
-            }
-            result = transformer.apply(getSlicedByteBuffer());
+            V v = valueDeserializeTransformer.apply(this.value);
```
Eshcar: so this is the core of the change bug fix (!)
here innerPut does not take a lock where previously `transformer.apply` used a lock
somewhat hard to notice this in the diff-clutter github created.

maybe a comment documenting `innerPut` and specifically mentioning it assumes a lock is taken by the invoking method can help

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -51,21 +51,25 @@ boolean remove(MemoryManager memoryManager) {
...
 
-    boolean put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
+    <V> boolean put(V newVal, OakSerializer<V> serializer, MemoryManager memoryManager) {
         writeLock.lock();
         if (isDeleted()) {
             writeLock.unlock();
             return false;
         }
+        innerPut(newVal, serializer, memoryManager);
+        writeLock.unlock();
```
eranmeir: I know this was in the code before, but this may be a good time to consider:
`innerPut` calls serializer methods, which we don't necessarily have control over, and potentially shuffles memory, which is also risky. I think we should check for exceptions and unlock if anything bad happens.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -280,7 +278,7 @@ private void updateIndexAndNormalize(List<Chunk<K, V>> engagedChunks, List<Chunk
         }
     }
 
-    private boolean rebalanceRemove(Chunk<K, V> c, K key) {
+    private boolean rebalanceRemove(Chunk<K, V> c) {
```
eranmeir: I would just remove this entire method

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -761,20 +749,7 @@ boolean replace(K key, V oldValue, V newValue, Function<ByteBuffer, V> valueDese
...
-
-            lookUp.handle.put(newValue, valueSerializer, memoryManager);
-
-            return true;
-        };
-
-        // res can be null if handle was deleted between lookup and the next call
-        Boolean res = (Boolean) lookUp.handle.mutatingTransform(replaceTransform);
-        return (res != null) ? res : false;
+        return lookUp.handle.compareExchange(oldValue, newValue, valueDeserializeTransformer, valueSerializer, memoryManager);
```
eranmeir: 👍

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -742,17 +739,8 @@ V replace(K key, V value, Function<ByteBuffer, V> valueDeserializeTransformer) {
...
-            V v = valueDeserializeTransformer.apply(bb);
-
-            lookUp.handle.put(value, valueSerializer, memoryManager);
-
-            return v;
-        };
-
         // will return null if handle was deleted between prior lookup and the next call
-        return (V) lookUp.handle.mutatingTransform(replaceTransform);
+        return lookUp.handle.exchange(value, valueDeserializeTransformer, valueSerializer, memoryManager);
```
eranmeir: 👍

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1208,12 +1183,12 @@ private Result(V value, boolean flag, boolean hasValue) {
 
         }
 
-        static Result withValue(Object value) {
-            return new Result(value, false, true);
+        static <V> Result<V> withValue(V value) {
```
eranmeir: 👍

core/src/main/java/com/oath/oak/Handle.java
```diff
@@ -196,27 +200,33 @@ public ByteOrder order() {
...
 
-    /**
-     * Applies a transformation under writers locking
-     *
-     * @param transformer transformation to apply
-     * @return Transformation result or null if value is deleted
-     */
-    <T> T mutatingTransform(Function<ByteBuffer, T> transformer) {
-        T result;
+    <V> V exchange(V newValue, Function<ByteBuffer, V> valueDeserializeTransformer, OakSerializer<V> serializer, MemoryManager memoryManager) {
```
eranmeir: I like the idea of encapsulating atomic exchange, but I think there's subtle semantic difference that is lost here. Specifically, these implementation of exchange and compareExchange are not zero copy. The original idea behind `mutatingTransform` was to be a potentially zero-copy transformation that takes a write lock. The thought was that going forward we may want to extend the zero-copy API to support more standard API methods while keeping ZC performance.

yoavz1997: We can perhaps pass the chunk's comparator instead of valueDeserializeTransformer to support a ZC exchange.

PR #48, opened 2019-09-11
Add StreamIterators

sanastas:
Please take a look on suggestion for additional type of iterators.


core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -970,6 +963,48 @@ private void initAfterRebalance() {
...
+                throw new NoSuchElementException();
+            }
+
+            Chunk.State chunkState = state.getChunk().state();
+
+            if (chunkState == Chunk.State.RELEASED) {
+                initAfterRebalance();
+            }
+
+            if (key != null) state.getChunk().setKeyRefer(state.getIndex(),key);
```
ebortnikov: In a  block please ...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1045,10 +1080,12 @@ private void advanceState() {
             int nextIndex = chunkIter.next();
             state.set(chunk, chunkIter, nextIndex);
 
-            ByteBuffer key = state.getChunk().readKey(state.getIndex());
-            if (!inBounds(key)) {
-                state = null;
-                return;
+            if (((hi != null) && !isDescending) || ((lo != null) && isDescending)) {
```
ebortnikov: Please comment that this is an optimization. 

IMHO, would be good to remove the parenthesis around (hi != null) and (lo != null)

core/src/main/java/com/oath/oak/NativeAllocator/Block.java
```diff
@@ -102,7 +102,7 @@ public ByteBuffer getReadOnlyBufferForThread(int position, int length) {
         bb.limit(position + length);
         bb.position(position);
         // on purpose not creating a ByteBuffer slice() here,
-        // slice() will be used only per demand when buffer is passed to the serializer
+        // slice() will be used only per demand when buffer is passed outside Oak
```
ebortnikov: Comment unclear

core/src/main/java/com/oath/oak/OakComparator.java
```diff
@@ -9,6 +9,7 @@
 
 import java.nio.ByteBuffer;
 
+// IMPORTANT: it shouldn't be assumed that any ByteBuffer position starts at zero
```
ebortnikov: Maybe say it in a positive way, e.g., ByteBuffer position might be any nonnegative integer

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public void setKeyLength(int keyLength) {
+        this.keyLength = keyLength;
+        this.bb = null;
+    }
+
+    @Override
+    public int capacity() {
+        return getTemporalPerThreadByteBuffer().capacity();
```
ebortnikov: Temporal --> Temporary?

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Override
+    public double getDouble(int index) {
+        return getTemporalPerThreadByteBuffer().getChar(index+keyPosition);
+    }
+
+    @Override
+    public <T> T transform(Function<ByteBuffer, T> transformer) {
+        // TODO: here we might need slice() or to be sure transformer doesn't assume starting at position 0
```
ebortnikov: Not clear. Does this indicate that there is some undone work? What is the meaning of this comment?

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        return transformer.apply(getTemporalPerThreadByteBuffer());
+    }
+
+    @Override
+    public void unsafeCopyBufferToIntArray(int srcPosition, int[] dstArray, int countInts) {
```
ebortnikov: Please comment

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        UnsafeUtils.unsafeCopyBufferToIntArray(getTemporalPerThreadByteBuffer(), srcPosition+keyPosition, dstArray, countInts);
+
+    }
+
+    private ByteBuffer getTemporalPerThreadByteBuffer() {
+        // TODO: how to check that the entire Oak wasn't already closed with its memoryManager
```
ebortnikov: 1. Please use the common terminology: Oak --> OakMap? 
2. Does the comment indicate a potential bug?

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -13,12 +13,16 @@
 
 public class OakRValueBufferImpl implements OakRBuffer {
```
ebortnikov: Maybe it would be better to have two OakRBuffer implementations: fixed-reference and non-fixed reference. It was good to have an impl with a fixed reference.

core/src/main/java/com/oath/oak/InternalOakMap.java
sanastas: This and other comments that are resolved without comment, means they are fixed

core/src/main/java/com/oath/oak/NativeAllocator/Block.java
sanastas: Please read two comment lines together. Is it still unclear?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -970,6 +963,48 @@ private void initAfterRebalance() {
...
+            Handle currentHandle = state.getChunk().getHandle(state.getIndex());
+            advanceState();
+            return currentHandle;
+        }
+
+        /**
+         * Advances next to higher key without creating a ByteBuffer for key.
+         * Do not get and return handle/value
+         */
+        void advanceStreamKeyOnly(OakRKeyReferenceImpl key) {
```
eranmeir: I'm not sure why we need two almost identical methods here. At worst, we can use the same method with a single extra key reference instance in values stream iterator, but I think we can think of some abstraction that can save even this single instance and reduce duplication.

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
sanastas: Applied the slice() method to be on the safe side

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
sanastas: Renamed+the problem was fixed

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
sanastas: The difference between two classes would be to small... I prefer to leave it like this, if OakRBuffer needs to be immutable just don't change it's buffer...

core/src/main/java/com/oath/oak/InternalOakMap.java
Eshcar: I agree
for example line 1004 need to check if key!=null like 982
either remove the version that does not return handle or have a single method with a boolean to indicate whether or not to return handle.

core/src/main/java/com/oath/oak/OakComparator.java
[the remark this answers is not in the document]
sanastas: added

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -13,12 +13,16 @@
...
 
-    private final Handle handle;
+    private Handle handle;
 
     OakRValueBufferImpl(Handle handle) {
         this.handle = handle;
     }
 
+    public void setHandle(Handle handle) {
```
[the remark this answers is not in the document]
sanastas: done

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -919,4 +943,59 @@ public int size() {
         }
     }
 
+    static final class KeyStreamBufferSet<K, V> extends AbstractSet<OakRBuffer> {
```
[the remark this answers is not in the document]
sanastas: I would better live with it for now...

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.oath.oak;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.function.Function;
+
+public class OakRKeyReferenceImpl implements OakRBuffer {
```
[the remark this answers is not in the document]
sanastas: done

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    // The OakRKeyReferBufferImpl user accesses OakRKeyReferBufferImpl
+    // as it would be a ByteBuffer with initially zero position.
+    // We translate it to the relevant ByteBuffer position, by adding keyPosition to any given index
+
+    OakRKeyReferenceImpl(MemoryManager memoryManager) {
+        this.memoryManager = memoryManager;
+    }
+
+    public void setBlockID(int blockID) {
```
Eshcar: Do you need to nullify bb if blockid is not changed?

If these 3 methods are always called together `setBlockID` `setKeyPosition` and `setKeyLength` then maybe have a single method with 3 parameters and invalidate bb just once.

opening an option to set only some of them is a potential for bugs

Eshcar: You also need to add the disclaimer that this code is for use only in special cases like flush where only a single reader go over the entries and it is not updated during the traversal.

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+package com.oath.oak;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.function.Function;
+
+public class OakRKeyReferenceImpl implements OakRBuffer {
+
+    private int blockID = 0;
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/test/java/com/oath/oak/OakViewTests.java
```diff
@@ -120,20 +120,41 @@ public void testBufferViewAPIs() {
             assertEquals(values[i], key);
         }
 
+        Iterator<OakRBuffer> keyStreamIterator = oak.zc().keyStreamSet().iterator();
```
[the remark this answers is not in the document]
sanastas: done

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -970,6 +963,48 @@ private void initAfterRebalance() {
             return new AbstractMap.SimpleImmutableEntry<>(bb, currentHandle);
         }
 
+        /**
+         * Advances next to higher entry without creating a ByteBuffer for key.
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -207,6 +205,27 @@ ByteBuffer readKey(int entryIndex) {
         return bb;
     }
 
+    /**
+     * Sets the given key reference (OakRKeyReferBufferImpl) given the entry index.
+     * There is no copy just a special ByteBuffer for a single key.
+     * The thread-local ByteBuffer can be reused by different threads, however as long as
+     * a thread is invoked the ByteBuffer is related solely to this thread.
+     */
+    void setKeyRefer(int entryIndex, OakRKeyReferenceImpl keyReferBuffer) {
```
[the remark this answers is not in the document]
sanastas: As it is hard to take the entries array calculations out of chunk I have created just a setter for all OakRKeyReferenceImpl fields together

core/src/main/java/com/oath/oak/InternalOakMap.java
sanastas: Left only one method

core/src/main/java/com/oath/oak/OakRKeyReferenceImpl.java
sanastas: fixed

PR #51, opened 2019-09-23
Remove Type Warnings and Formatting

yoavz1997:

### DIFF
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -7,6 +7,7 @@
 package com.oath.oak;
 
 import sun.misc.Unsafe;
+
 import java.lang.reflect.Constructor;
 import java.nio.ByteBuffer;
 import java.util.Comparator;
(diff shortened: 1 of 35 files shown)


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -45,24 +46,20 @@
     private static final double MAX_ENTRIES_FACTOR = 2;
     private static final double MAX_IDLE_ENTRIES_FACTOR = 5;
 
-    // when chunk is frozen, all of the elements in pending puts array will be this OpData
-    private static final OpData FROZEN_OP_DATA =
```
sanastas: Please do not remove this... I still would like to live with the possibility that rebalance will return to be non-blocking one day...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -507,20 +516,22 @@ int linkEntry(int ei, boolean cas, K key) {
...
+    public int getMaxItems() {
+        return maxItems;
+    }
 
     /**
      * Updates a linked entry to point to handle or otherwise removes such a link. The handle in
      * turn has the value. For linkage this is an insert linearization point.
      * All the relevant data can be found inside opData.
-     *
+     * <p>
```
sanastas: I am sorry, why "<p>" is here?

sanastas: /<p/>

sanastas:
I went over the change and it looks good to be. Just reference my two comments above. Thanks!

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -174,10 +173,10 @@ private boolean casEntriesArray(int item, int offset, int expected, int value) {
     /**
      * write key in slice
      **/
-    void writeKey(K key, int ei) {
+    private void writeKey(K key, int ei) {
```
eranmeir: I would leave this package private, since this is sort of an internal API (like `readKey`, etc.)

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -53,7 +53,7 @@
 
     // internal constructor, to create OakMap use OakMapBuilder
     OakMap(K minKey, OakSerializer<K> keySerializer, OakSerializer<V> valueSerializer, OakComparator<K> oakComparator,
-           int chunkMaxItems, int chunkBytesPerItem, MemoryManager mm, ThreadIndexCalculator threadIndexCalculator) {
+           int chunkMaxItems, MemoryManager mm, ThreadIndexCalculator threadIndexCalculator) {
```
eranmeir: You can remove `threadIndexCalculator` from here as well

core/src/main/java/com/oath/oak/Chunk.java
eranmeir: I don't think we should leave unused code in the codebase. If we need it back, that's what version control is for.

PR #53, opened 2019-09-26
Oak Without Handles

yoavz1997:


benchmarks/synchrobench/run.sh
```diff
@@ -3,10 +3,10 @@
 dir=`pwd`
 output=${dir}/output
 java=java
-jarfile="target/oak-benchmarks-synchrobench-0.1.4-SNAPSHOT.jar"
+jarfile="target/oak-benchmarks-synchrobench-0.1.6-SNAPSHOT.jar"
 
 thread="01 04 08 12"
-size="5000000"
+size="10000000"
```
sanastas: To change the path to jar is OK. But to change other parameters I am not sure... One one hand, anyway everyone should actually change it. One the other hand, our previous "regression" was without this change.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
```diff
@@ -13,441 +13,478 @@
...
+    /**
+     * Execute the main thread that starts and terminates the benchmark threads
+     *
+     * @throws InterruptedException
+     */
+    private void execute(int milliseconds, boolean maint)
+            throws InterruptedException {
+        printHeapStats("Before");
+        long startTime = System.currentTimeMillis();
+        long count = fill(Parameters.range, Parameters.size);
```
sanastas: Why do you count all operations (and not successes of putIfAbsent)? Isn't it just size? Why is it interesting to print it?

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
```diff
@@ -90,7 +90,7 @@ public void stopThread() {
     public void run() {
 
         boolean change = Parameters.change;
-        int size = 100;
+        int size = 10000;
```
sanastas: Why to change?

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
```diff
@@ -13,10 +13,10 @@
     private com.oath.oak.OakMap<MyBuffer, MyBuffer> oak;
     private OakMapBuilder<MyBuffer, MyBuffer> builder;
     private MyBuffer minKey;
-    private OakNativeMemoryAllocator ma;
+    public OakNativeMemoryAllocator ma;
```
sanastas: What is the need for this change?

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
```diff
@@ -93,7 +93,7 @@ public boolean descendOak(K from, int length) {
     private boolean createAndScanView(com.oath.oak.OakMap<MyBuffer, MyBuffer> subMap, int length) {
         Iterator iter;
         if (Parameters.zeroCopy) {
-            iter = subMap.zc().keySet().iterator();
+            iter = subMap.zc().keyStreamSet().iterator();
```
sanastas: Can you please add some new parameter tot he entire script? So we might run both regular iterator and the stream one?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -17,27 +18,45 @@
...
+import static com.oath.oak.NativeAllocator.OakNativeMemoryAllocator.INVALID_BLOCK_ID;
+import static com.oath.oak.UnsafeUtils.intsToLong;
+
 public class Chunk<K, V> {
 
     /*-------------- Constants --------------*/
 
-    static final int NONE = 0;    // constant for "no index", etc. MUST BE 0!
+    enum OFFSET {
+        NEXT(0), KEY_POSITION(3), VALUE_STATS(1), VALUE_POSITION(1),
```
sanastas: What are "STATS"? Position+length+BlockID? If yes, can we please use the word "reference"? Here and everywhere...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -17,27 +18,45 @@
...
+        VALUE_BLOCK_AND_LENGTH(2), VALUE_BLOCK(2), VALUE_LENGTH(2), KEY_BLOCK_AND_LENGTH(4),
+        KEY_BLOCK(4), KEY_LENGTH(4);
+
+        public final int value;
```
sanastas: Please add comments explaining what are those offsets (as there were comments in the original code). Also explain about this enum usage in general.

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -17,27 +18,45 @@
...
+    }
+
+    enum State {
+        INFANT,
+        NORMAL,
+        FROZEN,
+        RELEASED
+    }
 
+    static final int NONE = 0;    // constant for "no index", etc. MUST BE 0!
```
sanastas: What "index" is it referenced to? There can be multiple meanings... (I think this is about next-entry-index...)

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -47,28 +66,26 @@
...
     private static final Unsafe unsafe;
     private final MemoryManager memoryManager;
     ByteBuffer minKey;       // minimal key that can be put in this chunk
-    AtomicMarkableReference<Chunk<K,V>> next;
+    AtomicMarkableReference<Chunk<K, V>> next;
     Comparator<Object> comparator;
 
     // in split/compact process, represents parent of split (can be null!)
-    private AtomicReference<Chunk> creator;
+    private AtomicReference<Chunk<K, V>> creator;
```
sanastas: Aren't we done with those <K,V> ?!

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -124,34 +139,27 @@
...
-        FROZEN,
-        RELEASED
-    }
-
     static class OpData {
         final Operation op;
         final int entryIndex;
-        final int handleIndex;
-        int prevHandleIndex;
+        final long newValueStats;
```
sanastas: As I have commented already, please change the name "stats" everywhere...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -164,26 +172,31 @@ private int compare(Object k1, Object k2) {
...
+                Unsafe.ARRAY_INT_BASE_OFFSET + (item + offset.value) * Unsafe.ARRAY_INT_INDEX_SCALE,
                 expected, value);
     }
 
 
     /**
      * write key in slice
      **/
-    void writeKey(K key, int ei) {
+    private void writeKey(K key, int ei) {
```
sanastas: 👍

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -215,21 +227,20 @@ void setKeyRefer(int entryIndex, OakRKeyReference keyReferBuffer) {
...
     /**
      * release key in slice, currently not in use, waiting for GC to be arranged
      **/
     void releaseKey(int entryIndex) {
-        int blockID = getEntryField(entryIndex, OFFSET_KEY_BLOCK);
-        int keyPosition = getEntryField(entryIndex, OFFSET_KEY_POSITION);
-        int length = getEntryField(entryIndex, OFFSET_KEY_LENGTH);
-        ByteBuffer bb = memoryManager.getByteBufferFromBlockID(blockID, keyPosition,length);
-        Slice s = new Slice(blockID, bb);
+        int blockID = getEntryField(entryIndex, OFFSET.KEY_BLOCK);
```
sanastas: If we have here all those writeKey/readKey/setKeyRef/releaseKey AND values are now treated similarly as a keys, isn't it reasonable to have near writeValue/readValue/setValueRef/releaseValue or even same function to deal with key/value?

sanastas:
I am not done at all ! Just that is for today! Will continue :)

benchmarks/synchrobench/run.sh
yoavz1997: The size used in the paper is 10M. This way the right execution command is created.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
yoavz1997: We count the operations to measure the performance of filling up the index.
Since not every `putIfAbsent` succeeds, but it is still and operation that was executed, we count it.
Otherwise, it would not reflect the actual throughput, but rather the randomness of each run.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
yoavz1997: We wanted to run scans of 10K elements and not 100. 
I can change it back in the master it would be a 100.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
yoavz1997: In order to print the memory allocated there needs to be access to the memory allocator.
I guess a getter can be added instead if necessary.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OakMap.java
yoavz1997: Will do!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Will do!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: It refers to a null next-index (it is not my addition, it just moved around).

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: I did not rebase this branch of the master yet since more code is about to be merged soon.

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Will do!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: There is still a difference between keys and values.
First, keys do not change and thus reading them can be does non-atomically (should be kept that way to avoid unnecessary overhead when performing a lookup).
Second, values have a header to initialize and keys do not, and value are not directly written to an entry and keys do.

Regarding setKeyRef, I'm not sure how to use it for values.

I added releaseValue for consistency sake.

benchmarks/synchrobench/run.sh
sanastas: OK

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
sanastas: But fill() method does exactly Parameters.size operations in its for-loop. You have the number, why to count?

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
sanastas: Please remain with 100

core/src/main/java/com/oath/oak/Chunk.java
sanastas: Yes, I saw it is not your addition. Please correct the comment anyway, it is just couple of characters :)

core/src/main/java/com/oath/oak/Chunk.java
sanastas: OK

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
yoavz1997: It may do more `putIfAbsent` if the key it draws already exists

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
yoavz1997: Will do!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Will do!

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -247,68 +290,99 @@ ByteBuffer readMaxKey() {
...
     /**
      * gets the value for the given item, or 'null' if it doesn't exist
      */
-    Handle getHandle(int entryIndex) {
-
-        int hi = getEntryField(entryIndex, OFFSET_HANDLE_INDEX);
+    Slice getValueSlice(int entryIndex) {
 
+        long valueReference = getValueReference(entryIndex);
+        int[] valueArray = UnsafeUtils.longToInts(valueReference);
```
sanastas: Why valueArray is needed? Can't getValueReference signal anyhow that the value is not there? Like some invalid long?

sanastas: The INVALID_BLOCK_ID is also checked later inside buildValueSlice()...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -247,68 +290,99 @@ ByteBuffer readMaxKey() {
...
         }
     }
 
     /**
      * gets the value for the given item, or 'null' if it doesn't exist
      */
-    Handle getHandle(int entryIndex) {
-
-        int hi = getEntryField(entryIndex, OFFSET_HANDLE_INDEX);
+    Slice getValueSlice(int entryIndex) {
```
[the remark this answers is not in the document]
sanastas: Or as I see it you have also getValueReference. So to make it all similar: getValueReference, getValueSlice, and getValueByteBuffer

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -247,68 +290,99 @@ ByteBuffer readMaxKey() {
...
+        if ((valueArray[0] >>> VALUE_BLOCK_SHIFT) == INVALID_BLOCK_ID) {
+            return null;
         }
+        return new Slice(valueArray[0] >>> VALUE_BLOCK_SHIFT, valueArray[1], valueArray[0] & VALUE_LENGTH_MASK,
+                memoryManager);
+    }
+
+    // Assuming the reading of valuePosition and valueBlockAndLength is atomic!
+    long getValueReference(int entryIndex) {
+        int valuePosition, valueBlockAndLength;
```
sanastas: Aren't position+blockId+length updated atomically together with a long CAS?

sanastas: If yes, can't we read them atomically together as long?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -317,41 +391,49 @@ int getHandleIndex(int entryIndex) {
     LookUp lookUp(K key) {
         // binary search sorted part of key array to quickly find node to start search at
         // it finds previous-to-key so start with its next
-        int curr = getEntryField(binaryFind(key), OFFSET_NEXT);
-        int cmp;
+        int curr = getEntryField(binaryFind(key), OFFSET.NEXT);
+        int cmp = -1;
```
sanastas: Have you seen any problem or just common sense?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -317,41 +391,49 @@ int getHandleIndex(int entryIndex) {
...
-                // if keys are equal - we've found the item
+            }
+            // if keys are equal - we've found the item
             else if (cmp == 0) {
-                int hi = getEntryField(curr, OFFSET_HANDLE_INDEX);
-                if (hi < 0) return new LookUp(null, curr, -1);
-                Handle h = handles[hi];
-                if (h.isDeleted()) return new LookUp(null, curr, hi);
-                return new LookUp(h, curr, hi);
+                long valueReference = getValueReference(curr);
```
sanastas: Why not to use getValueSlice directly here?

sanastas: I see that you use them both. Why do you need them both? Slice for simplicity of usage and reference for any future CAS? If so please add a comment about that.

sanastas: The comment is more about LookUp class

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -412,28 +497,15 @@ void unpublish() {
...
-
     int allocateEntryAndKey(K key) {
         int ei = entryIndex.getAndAdd(FIELDS);
         if (ei + FIELDS > entries.length) {
             return -1;
         }
 
         // key and value must be set before linking to the list so it will make sense when reached before put is done
-        setEntryField(ei, OFFSET_HANDLE_INDEX, -1); // set value index to -1, value is init to null
+        setEntryField(ei, OFFSET.VALUE_POSITION, 0); // set value index to -1, value is init to null
```
sanastas: Why setting separately? Because the entry is not linked and still private? If so please comment

sanastas: Again, didn't come even close to finish the review. Got to linkEntry() of the Chunk. Will continue. This is mostly comment for myself :)

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Will be fixed!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: We can using unsafe. Will be fixed!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: We need the slice so we are able to write off-heap, and to signal that the value reference points to a deleted value (in case the slice is null and the reference is valid).
The reference is needed for the CAS later (e.g., when removing we need the old reference).

I think in the new version of @eranmeir (#42), where we only CAS `null` to a valid value, and a valid value to `null`, the reference is not always necessary (it is still needed for remove).

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: This can also be done using unsafe. Will be fixed!

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -494,44 +519,38 @@ boolean putIfAbsentComputeIfPresent(K key, V value, Consumer<OakWBuffer> compute
...
-            c.freeHandle(hi);
-            return false;
+        long res = finishAfterPublishing(opData, c);
+        if (res != DELETED_VALUE) {
+            c.releaseValue(newValueReference);
         }
+        return res == DELETED_VALUE;
     }
 
     V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
```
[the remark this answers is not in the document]
yoavz1997: This is a part of the Map API.
Remove should return the previous value.
In ZC map remove only returns a Boolean.

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Is it still relevant?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -679,89 +754,86 @@ void freeze() {
...
+            if (entriesToCopy > 0) {
                 for (int i = 0; i < entriesToCopy; ++i) {
                     int offset = i * FIELDS;
                     // next should point to the next item
-                    entries[sortedEntryIndex + offset + OFFSET_NEXT]
-                        = sortedEntryIndex + offset + FIELDS;
+                    entries[sortedEntryIndex + offset + OFFSET.NEXT.value]
+                            = sortedEntryIndex + offset + FIELDS;
 
-                    // copy the values of key position, key block index + key length => 2 integers via array copy
```
sanastas: Can you please return this comment and enhance it?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -472,122 +543,132 @@ int linkEntry(int ei, boolean cas, K key) {
...
     /**
      * write value in place
      **/
-    void writeValue(int hi, V value) {
-        assert hi >= 0 ;
-        ByteBuffer byteBuffer = memoryManager.allocate(valueSerializer.calculateSize(value));
-        // just allocated bytebuffer is ensured to have position 0
-        valueSerializer.serialize(value, byteBuffer.slice());
-        handles[hi].setValue(byteBuffer);
+    long writeValueOffHeap(V value) {
```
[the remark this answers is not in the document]
yoavz1997: In `writeKey`, the key reference is also written to the entry.
In `writeValueOffHeap`, the value is only written via the value reference, and the value reference is returned, not written to an entry.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -23,6 +21,18 @@
 
 public class OakNativeMemoryAllocator implements OakMemoryAllocator {
 
+    private static class FreeChuck {
```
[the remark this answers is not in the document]
yoavz1997: This code is relevant for utilizing the fact that the free-list is a skip list

core/src/main/java/com/oath/oak/Chunk.java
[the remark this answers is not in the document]
sanastas: It is OK that this is a different functionality, I am not saying to collide them, just use 'writeValue' as the function name

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -132,14 +139,12 @@ int entries() {
     }
 
 
-
-
-    private Rebalancer.RebalanceResult rebalance(Chunk<K, V> c) {
+    private boolean rebalance(Chunk<K, V> c) {
```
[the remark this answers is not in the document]
sanastas: Please add a comment explaining what return value means

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -297,9 +296,23 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
 
         Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp != null && lookUp.handle != null) {
-            V v = (transformer != null) ? (V) lookUp.handle.transform(transformer) : null;
-            lookUp.handle.put(value, valueSerializer, memoryManager);
+        // If there is a matching value reference for the given key, and it is not marked as deleted, then this put
+        // changes the slice pointed by this value reference.
+        if (lookUp != null && lookUp.valueSlice != null) {
+            V v = null;
```
sanastas: Will require a big reshaping while merging with current Eran's version...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -297,9 +296,23 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
 
         Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp != null && lookUp.handle != null) {
-            V v = (transformer != null) ? (V) lookUp.handle.transform(transformer) : null;
-            lookUp.handle.put(value, valueSerializer, memoryManager);
+        // If there is a matching value reference for the given key, and it is not marked as deleted, then this put
```
sanastas: I am skipping put() and will look again after merging

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -369,11 +375,18 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
             throw new NullPointerException();
         }
 
-        Chunk c = findChunk(key); // find chunk matching key
+        Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp != null && lookUp.handle != null) {
-            if (transformer == null) return Result.withFlag(false);
-            return Result.withValue(lookUp.handle.transform(transformer));
+        if (lookUp != null && lookUp.valueSlice != null) {
```
sanastas: Similar, skipping putIfAbsent() and will return after merge

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -614,10 +641,10 @@ OakRBuffer get(K key) {
 
         Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp == null || lookUp.handle == null) {
+        if (lookUp == null || lookUp.valueSlice == null) {
```
sanastas: Sometimes you compare lookUp.valueReference with DELETED_VALUE and sometimes lookUp.valueSlice with null... What is the rule? What means what? Must be consistent...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -627,26 +654,27 @@ OakRBuffer get(K key) {
...
             return null;
         }
 
-        T transformation = (T) lookUp.handle.transform(transformer);
-        return transformation;
+        AbstractMap.SimpleEntry<ValueUtils.ValueResult, T> res = ValueUtils.transform(lookUp.valueSlice, transformer);
```
sanastas: First, require a comment of what is going on here.
Second, when ValueUtils.transform may return RETRY? When it was deleted? If so why not to return NULL as it is deleted? Why to retry?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -627,26 +654,27 @@ OakRBuffer get(K key) {
...
+        if (res.getKey() == RETRY) {
+            return getValueTransformation(key, transformer);
+        }
+        return res.getValue();
+    }
 
+    private <T> T getValueTransformation(ByteBuffer key, Function<ByteBuffer, T> transformer) {
```
sanastas: Why this method is needed?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -657,7 +685,7 @@ ByteBuffer getKey(K key) {
 
         Chunk<K, V> c = findChunk(key);
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp == null || lookUp.handle == null || lookUp.entryIndex == -1) {
+        if (lookUp == null || lookUp.valueSlice == null || lookUp.entryIndex == -1) {
             return null;
         }
         return c.readKey(lookUp.entryIndex).slice();
```
sanastas: Where was/is it used? Why do we have to add slice() at the end here? Can you please add some comment?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -779,10 +825,11 @@ boolean replace(K key, V oldValue, V newValue, Function<ByteBuffer, V> valueDese
             return new AbstractMap.SimpleImmutableEntry<>(null, null);
         }
 
-        c.getHandle(prevIndex);
+        // TODO: No lock?
         return new AbstractMap.SimpleImmutableEntry<>(
                 keySerializer.deserialize(prevKey),
-                valueSerializer.deserialize(c.getHandle(prevIndex).getSlicedReadOnlyByteBuffer()));
+                valueSerializer.deserialize(ValueUtils.getActualValueBuffer(c.getValueSlice(prevIndex).getByteBuffer())
```
sanastas: getActualValueBuffer --> getValueBuffer

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -715,40 +743,58 @@ boolean computeIfPresent(K key, Consumer<OakWBuffer> computer) {
...
+        Chunk<K, V> c = findChunk(key); // find chunk matching key
         Chunk.LookUp lookUp = c.lookUp(key);
-        if (lookUp == null || lookUp.handle == null) return false;
+        if (lookUp == null || lookUp.valueSlice == null) {
+            return false;
+        }
 
-        return lookUp.handle.compute(computer);
+        ValueUtils.ValueResult res = ValueUtils.compute(lookUp.valueSlice, computer);
+        if (res == RETRY) {
```
sanastas: I don't know if you are going to continue with "your result" (ValueResult) after merging with Eran's last code. But Eran had a class Result. There need to be one "result" in the code after all...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -938,17 +986,17 @@ private void initAfterRebalance() {
                 initAfterRebalance();
             }
 
-            ByteBuffer bb = state.getChunk().readKey(state.getIndex()).slice();
-            Handle currentHandle = state.getChunk().getHandle(state.getIndex());
+            ByteBuffer bb = needsKey ? state.getChunk().readKey(state.getIndex()).slice() : null;
+            Slice currentValue = needsValue ? state.getChunk().getValueSlice(state.getIndex()) : null;
```
sanastas: If value is deleted getValueSlice will return null. Isn't it a problem? Didn't we say we need a loop here?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -959,15 +1007,16 @@ Handle advanceStream(OakRKeyReference key, boolean keyOnly) {
...
-            Handle currentHandle = null;
             if (key != null) {
-              state.getChunk().setKeyRefer(state.getIndex(),key);
+                state.getChunk().setKeyRefer(state.getIndex(), key);
             }
-            if (!keyOnly) {
-              currentHandle = state.getChunk().getHandle(state.getIndex());
+            if (value != null) {
+                if (!state.getChunk().setValueRefer(state.getIndex(), value)) {
+                    return null;
```
sanastas: Why null? What does it mean? First of all comment.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1161,22 +1213,30 @@ public T next() {
...
             ByteBuffer serializedKey = pair.getKey();
-            if (handle == null) {
+            Slice valueSlice = pair.getValue();
+            if (valueSlice == null) {
                 return null;
             }
-            handle.readLock();
-            if (handle.isDeleted()) {
-                handle.readUnLock();
+            ValueUtils.ValueResult res = ValueUtils.lockRead(valueSlice);
```
sanastas: I assume all that is going to change after merge with Eran... Eran introduced all those transformations going inside the handle, and you might have something parallel that can be used here...

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I looked through the code, and I never use the value reference in a condition.
Can you refer me to where you saw the use of value reference in a condition?

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: When iterating over Oak, in case one of the values moved, we need to find the new value.
This function uses a `ByteBuffer` instead of `K` to retrieve the value.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: This method is used in `getKeyTransformation` which is currently not used, though I assume it is used in Druid.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: My result has 3 values, `SUCCESS, FAILURE` and `RETRY` and Eran uses a boolean result.
I could try and consolidate them both into one.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I now added the loop, but there is still the problem of `hasNext` being incorrect because of this loop.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I added javadocs to ValueUtils.
When `RETRY` is returned it means that the value was moved, not deleted.

yoavz1997:
A cleaner version that can be merged into master is in #55

PR #54, opened 2019-10-06
Fixing Non-ZC putIfAbsent

eranmeir:
Problem: non-zc putIfAbsent might have returned null without inserting the new value if the handle was deleted between lookup and value deserialization.
Fix: When the method detects that a handle was deleted after successful lookup, it restarts.


sanastas:
The problem was fixed during other refactoring

PR #55, opened 2019-10-06
Oak with No Handles 2

yoavz1997:
A continuation of #53, now with no merge problems.
Some of the comments of #53 are not resolved yet and thus are not reflected in this PR.

-------------------------------


core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -0,0 +1,382 @@
...
+     * Marks a value as DELETED and frees its slice (not including the header).
+     *
+     * @param s             the value Slice (including the header)
+     * @param memoryManager the memory manager to which the slice is returned to
+     * @param oldValue      in case of a conditional remove, this is the value to which the actual value is compared to
+     * @param transformer   value deserializer
+     * @param <V>           the type of the value
+     * @return {@code SUCCESS} if the value was removed successfully,
+     * {@code FAILURE} if the value is already deleted, or if it does not equal to {@code oldValue}
+     * {@code RETRY} if the value was moved.
```
sanastas: If the only meaning of "RETRY" is that the value was moved, can we name it "MOVED"? This will make some code using this return value more clear...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -17,27 +16,78 @@
...
+
+    enum State {
+        INFANT,
+        NORMAL,
+        FROZEN,
+        RELEASED
+    }
+
+    static final int NONE = 0;    // an entry with NONE as its next pointer, points to a null entry
+    static final long DELETED_VALUE = 0;
```
sanastas: Should it be "deleted value" or maybe "invalid value"?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -17,27 +16,78 @@
...
 
-    private static final int FIELDS = 4;  // # of fields in each item of key array
-    private static final int OFFSET_NEXT = 0;
-    private static final int OFFSET_KEY_POSITION = 1;
-    private static final int OFFSET_KEY_LENGTH = 2;
-    private static final int OFFSET_HANDLE_INDEX = 3;
-    // key block is not used as an offset, rather as request differentiation,
-    // key block is part of key length integer, thus key length is limited to 65KB
-    private static final int OFFSET_KEY_BLOCK = 4;
+    private static final int FIELDS = 6;  // # of fields in each item of key array
```
sanastas: key array --> entries array

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -76,17 +124,6 @@
 
     /*-------------- Constructors --------------*/
 
-    // static constructor - access and create a new instance of Unsafe
```
sanastas: Currently I do not see why was it removed... Maybe just moved to the other place?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -122,25 +157,19 @@
         this.valueSerializer = valueSerializer;
     }
 
-    enum State {
```
sanastas: Where is it now?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -501,21 +589,26 @@ int linkEntry(int ei, K key) {
...
-        // just allocated bytebuffer is ensured to have position 0
-        valueSerializer.serialize(value, byteBuffer.slice());
-        handles[hi].setValue(byteBuffer);
+    long writeValue(V value) {
+        // the length of the given value plus its header
+        int valueLength = valueSerializer.calculateSize(value) + ValueUtils.VALUE_HEADER_SIZE;
+        Slice slice = memoryManager.allocateSlice(valueLength);
+        // initializing the header lock to be free
+        slice.initHeader();
+        // since this is a private environment we can only use slice, instead of duplicate and then slice
```
sanastas: There is an ambiguity between our class Slice and ByteBuffer method slice()... Can you please make it clearer?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -530,33 +623,35 @@ boolean pointToValue(OpData opData) {
...
+        long foundValueReference = getValueReference(opData.entryIndex);
+        int foundValueBlockAndLength = UnsafeUtils.longToInts(foundValueReference)[BLOCK_ID_LENGTH_ARRAY_INDEX];
 
-        if (foundHandleIdx == expectedHandleIdx) {
+        if (expectedValueReference == foundValueReference) {
             return true; // someone helped
-        } else if (foundHandleIdx < 0) {
-            // the handle was deleted, retry the attach
-            opData.prevHandleIndex = -1;
+        } else if (foundValueBlockAndLength == INVALID_BLOCK_ID) {
```
sanastas: This is the check whether value was deleted. And there are multiple ways to check whether value was deleted... E.G. when we are going to really delete it later we will try to assign value reference to be DELETED_VALUE... May be at least some explanatory comment here...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -569,12 +664,15 @@ boolean pointToValue(OpData opData) {
...
+                opData.newValueReference)) {
             // update statistics only by thread that CASed
-            if (opData.prevHandleIndex < 0 && opData.handleIndex > 0) { // previously a remove
+            int oldValueBlock = UnsafeUtils.longToInts(opData.oldValueReference)[BLOCK_ID_LENGTH_ARRAY_INDEX];
+            int newValueBlock = UnsafeUtils.longToInts(opData.newValueReference)[BLOCK_ID_LENGTH_ARRAY_INDEX];
+            if (oldValueBlock == INVALID_BLOCK_ID && newValueBlock != INVALID_BLOCK_ID) { // previously a remove
                 statistics.incrementAddedCount();
                 externalSize.incrementAndGet();
-            } else if (opData.prevHandleIndex > 0 && opData.handleIndex == -1) { // removing
+            } else if (oldValueBlock != INVALID_BLOCK_ID && newValueBlock == INVALID_BLOCK_ID) { // removing
```
sanastas: What is the order between "block ID set to invalid" and "vale reference set to DELETED"?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -722,37 +817,27 @@ final int copyPartNoKeys(Chunk<K, V> srcChunk, int srcEntryIdx, int maxCapacity)
...
                     // next should point to the next item
-                    entries[sortedEntryIndex + offset + OFFSET_NEXT]
+                    entries[sortedEntryIndex + offset + OFFSET.NEXT.value]
                             = sortedEntryIndex + offset + FIELDS;
 
-                    // copy the values of key position, key block index + key length => 2 integers via array copy
+                    // copy both the key and the value references (without the padding) => 4 integers via array copy
                     System.arraycopy(srcChunk.entries,  // source array
-                            entryIndexStart + offset + OFFSET_KEY_POSITION,
+                            entryIndexStart + offset + OFFSET.NEXT.value + 1,
```
sanastas: Why +1? For header? Explain in a comment

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -237,68 +292,111 @@ ByteBuffer readMaxKey() {
...
+    // Should be used with OFFSET.VALUE_REFERENCE and OFFSET.KEY_REFERENCE
+    private long getEntryFieldLong(int item, OFFSET offset) {
+        long arrayOffset = Unsafe.ARRAY_INT_BASE_OFFSET + (item + offset.value) * Unsafe.ARRAY_INT_INDEX_SCALE;
+        assert arrayOffset % 8 == 0;
+        return unsafe.getLongVolatile(entries, arrayOffset);
+    }
+
     /**
      * * sets the field of specified offset to 'value' for given item in entry array
+     * NOT ATOMIC!
```
sanastas: I think that an assignment of a machine word **is** atomic, but more likely doesn't propagate to other threads without additional memory fence...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -237,68 +292,111 @@ ByteBuffer readMaxKey() {
...
+                assert ((value << VALUE_BLOCK_SHIFT) >>> VALUE_BLOCK_SHIFT) == value; // value is up to 2^9
+                entries[item + offset.value] =
+                        (entries[item + offset.value]) | (value << VALUE_BLOCK_SHIFT);
+                return;
+            default:
+                entries[item + offset.value] = value;
         }
     }
 
+    private void setEntryFieldLong(int item, OFFSET offset, long value) {
```
sanastas: This one is indeed NOT ATOMIC :) Please comment

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -237,68 +292,111 @@ ByteBuffer readMaxKey() {
...
+        unsafe.putLongVolatile(entries, arrayOffset, value);
+    }
+
     /**
      * gets the value for the given item, or 'null' if it doesn't exist
      */
-    Handle getHandle(int entryIndex) {
-
-        int hi = getEntryField(entryIndex, OFFSET_HANDLE_INDEX);
+    Slice getValueSlice(int entryIndex) {
```
sanastas: Why both getValueSlice and buildValueSlice are needed? Can getValueSlice just do the job?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -320,34 +419,35 @@ LookUp lookUp(K key) {
             }
             // if keys are equal - we've found the item
             else if (cmp == 0) {
-                int hi = getEntryField(curr, OFFSET_HANDLE_INDEX);
-                if (hi < 0) {
-                    return new LookUp(null, curr, -1);
+                long valueReference = getValueReference(curr);
+                Slice valueSlice = buildValueSlice(valueReference);
+                if (valueSlice == null) {
```
sanastas: Please add a comment that value-slice being null means the value is deleted

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -320,34 +419,35 @@ LookUp lookUp(K key) {
...
+                    assert valueReference == 0;
+                    return new LookUp(null, valueReference, curr);
                 }
-                Handle h = handles[hi];
-                if (h.isDeleted()) {
-                    return new LookUp(null, curr, hi);
+                if (ValueUtils.isValueDeleted(valueSlice)) {
```
sanastas: Please add a comment that even if slice is not null the value can be still marked as deleted, what is checked here...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -320,34 +419,35 @@ LookUp lookUp(K key) {
...
+                curr = getEntryFieldInt(curr, OFFSET.NEXT);
             }
         }
         return null;
     }
 
     static class LookUp {
 
-        final Handle handle;
+        Slice valueSlice;
```
sanastas: For the sake of the future people who will work with this code. Please add a comments explaining that Slice actually is an object that references the value. Slice is needed for simplicity of accessing value. Also slice object equals null means the value is deleted. But valueReference is a serialized reference to a value compacted into long number. It is needed for future CAS if it will come.

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -410,28 +511,15 @@ void unpublish() {
...
     int allocateEntryAndKey(K key) {
         int ei = entryIndex.getAndAdd(FIELDS);
         if (ei + FIELDS > entries.length) {
             return -1;
         }
 
         // key and value must be set before linking to the list so it will make sense when reached before put is done
-        setEntryField(ei, OFFSET_HANDLE_INDEX, -1); // set value index to -1, value is init to null
+        // setting the value reference to DELETED_VALUE atomically
+        setEntryFieldLong(ei, OFFSET.VALUE_REFERENCE, DELETED_VALUE);
```
sanastas: Setting value reference to "deleted" straight before the assignment looks weird to me. I think we need DELETED_VALUE and INVALID_ VALUE and to use them accordingly. This would make the code much more understandable. I think I wrote about this already somewhere...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -501,21 +589,26 @@ int linkEntry(int ei, K key) {
     /**
      * write value in place
      **/
-    void writeValue(int hi, V value) {
-        assert hi >= 0;
-        ByteBuffer byteBuffer = memoryManager.allocate(valueSerializer.calculateSize(value));
-        // just allocated bytebuffer is ensured to have position 0
-        valueSerializer.serialize(value, byteBuffer.slice());
-        handles[hi].setValue(byteBuffer);
+    long writeValue(V value) {
```
sanastas: explain that this method returns valueReference

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -501,21 +589,26 @@ int linkEntry(int ei, K key) {
...
+        // the length of the given value plus its header
+        int valueLength = valueSerializer.calculateSize(value) + ValueUtils.VALUE_HEADER_SIZE;
+        Slice slice = memoryManager.allocateSlice(valueLength);
```
sanastas: Please add a comment that allocateSlice does not creates a new ByteBuffer object as part of the Slice, the ByteBuffer object is created later when provided to value serializer. If it indeed works as I think...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -530,33 +623,35 @@ boolean pointToValue(OpData opData) {
...
+        // the old value reference is no longer there so we have nothing to do. Note that in case of non-ZC removal the
         // loop in remove() will lookup the key again so we will still return the previous value
         if (operation == Operation.REMOVE) {
             return true; // this is a remove, no need to try again and return doesn't matter
         }
 
         // the operation is either NO_OP, PUT, PUT_IF_ABSENT, COMPUTE
-        int expectedHandleIdx = opData.handleIndex;
-        int foundHandleIdx = getEntryField(opData.entryIndex, OFFSET_HANDLE_INDEX);
+        long expectedValueReference = opData.newValueReference;
```
sanastas: Why expectedValueReference is getting newValueReference? Most likely it is just bad naming of variables... However, opData has old and new valueReferences so the expected should be old... Please check and explain in a code comment

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -436,39 +440,38 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
             }
             int prevEi = c.linkEntry(ei, key);
             if (prevEi != ei) {
-                // something changed this entry right before we linked it.
-                prevHi = c.getHandleIndex(prevEi);
-                if (prevHi != -1) {
+                // some other thread linked its entry with the same key.
+                oldReference = c.getValueReference(prevEi);
+                if (oldReference != DELETED_VALUE) {
```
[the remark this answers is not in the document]
yoavz1997: I need the reference for the subsequent CAS, but I changed the condition to look at the slice itself.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
```diff
@@ -132,18 +133,20 @@ public void run() {
                     bench.computeOak(key);
                     numSize++;
                 } else {
-                    if (bench.computeIfPresentOak(key)) {
-                        numSize++;
-                    } else {
-                        failures++;
-                    }
+                    MyBuffer newKey = new MyBuffer(Parameters.keySize);
```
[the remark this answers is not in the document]
yoavz1997: I took this code for PIACIP invocation from Yoni. I'm not sure why we need to allocate a new key.

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: I used the unsafe instance from UnsafeUtils instead of creating a new instance.

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: I moved it to be under the OFFSET enum

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: I used putLongVolatile which makes it atomic

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Do you mean call both `getValueSlice`, or remove `buildValueSlice`.
If you meant the latter, I disagree since one takes an entryIndex and the other a reference.
Sometimes a reference is first read and later if needs be a slice is create, and both are needed (to read from the off-heap and CAS an entry).
If this is the case, we cannot read the entry twice since the slice could have changed in the mean time.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
```diff
@@ -13,441 +13,480 @@
...
+    }
+
+    /**
+     * Execute the main thread that starts and terminates the benchmark threads
+     *
+     * @throws InterruptedException
+     */
+    private void execute(int milliseconds, boolean maint)
+            throws InterruptedException {
+        printHeapStats("Before");
```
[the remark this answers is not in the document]
yoavz1997: We always print these statistics.
They do not affect the measurements since the GC is done before the time measurement begins and after it ends.

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: I do not understand what you mean in this comment by "... does not create a new ByteBuffer ... the ByteBuffer is created later ...".

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Previously there were `handleIndex` and `prevHandleIndex`. There is a condition checking whether someone else did the CAS for us when comparing `expectedValueReference` and `foundValueReference`.
To avoid naming problems I inlined this variable, and now the condition is `if (opData.newValueReference == foundValueReference)`.

core/src/main/java/com/oath/oak/Chunk.java
sanastas: OK, leave it as is

core/src/main/java/com/oath/oak/NativeAllocator/BlocksPool.java
```diff
@@ -45,15 +54,22 @@ static BlocksPool getInstance() {
         return instance;
     }
 
+    static void setBlockSize(int blockSize) {
```
sanastas: Why is it needed? For tests? Please add some comment

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/contention/benchmark/Test.java
```diff
@@ -13,441 +13,480 @@
...
+            switch (benchType) {
+                case OAKMAP:
+                    MyBuffer key = new MyBuffer(Parameters.keySize);
+                    key.buffer.putInt(0, v);
+                    MyBuffer val = new MyBuffer(Parameters.valSize);
+                    val.buffer.putInt(0, v);
+                    if (oakBench.putIfAbsentOak(key, val)) {
+                        i--;
+                    }
+                    operations++;
```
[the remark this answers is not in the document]
sanastas: "counting all the putIfAbsent invocations"

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -152,10 +158,15 @@ int entries() {
     }
 
 
-    private Rebalancer.RebalanceResult rebalance(Chunk<K, V> c) {
+    /**
+     * @param c - Chunk to rebalance
+     * @return - returns {@code true} if this thread managed to CAS to newChunk list of rebalance (the result of
+     * {@code rebalancer.createNewChunks}). {@code false} otherwise.
+     */
+    private boolean rebalance(Chunk<K, V> c) {
```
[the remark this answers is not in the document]
sanastas: yes

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -315,15 +320,15 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-            Result<V> res = lookUp.handle.exchange(value, transformer, valueSerializer, memoryManager);
-            if (res.hasValue) {
-                return res.value;
-            } else if (res.flag) {
-                // Exchange was successful, but we don't care about value, i.e., this is a ZC scenario
-                return null;
+        // If there is a matching value reference for the given key, and it is not marked as deleted, then this put
+        // changes the slice pointed by this value reference.
+        if (lookUp != null && lookUp.valueSlice != null) {
+            AbstractMap.SimpleEntry<ValueUtils.ValueResult, V> res = ValueUtils.exchange(c, lookUp, value,
```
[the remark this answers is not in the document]
sanastas: Will enhance Eran's result

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -33,10 +43,14 @@
...
-            Comparator.<Map.Entry<Long, Slice>, Integer>comparing(p -> p.getValue().getByteBuffer().remaining())
-            .thenComparing(Map.Entry::getKey);
-    private final ConcurrentSkipListSet<Map.Entry<Long,Slice>> freeList = new ConcurrentSkipListSet<>(comparator);
+    private final ConcurrentSkipListSet<FreeChuck> freeList = new ConcurrentSkipListSet<>((x, y) -> {
+        if (x.length == y.length) {
+            return (int) (x.id - y.id);
+        }
+        return (int) (x.length - y.length);
+    });
+    private final FreeChuck[] dummies = new FreeChuck[ThreadIndexCalculator.MAX_THREADS];
```
sanastas: Please add a comment what are those "dummies" for ...

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -57,83 +71,83 @@
...
         this.capacity = capacity;
+        for (int i = 0; i < ThreadIndexCalculator.MAX_THREADS; i++) {
+            dummies[i] = new FreeChuck(-1, 0, null);
+        }
     }
 
     @Override
     public ByteBuffer allocate(int size) {
-      return allocate(size, true);
+        return allocateSlice(size).getByteBuffer();
```
sanastas: I allocate() method in use anywhere? Is it @Override (I think we changed the interface)?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -57,83 +71,83 @@
...
-            if (bb.remaining() >= size && freeList.remove(kv)) {
-              if (stats != null) stats.reclaim(size);
-              return bb;
+    @Override
+    public Slice allocateSlice(int size) {
+
+        FreeChuck myDummy = dummies[threadIndexCalculator.getIndex()];
+        while (!freeList.isEmpty()) {
+            myDummy.length = size;
+            FreeChuck bestFit = freeList.higher(myDummy);
```
sanastas: Please add a comment explaining the algorithm. I assume you are looking for first fit in size in LogN...

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -57,83 +71,83 @@
...
+            if (bestFit == null) {
+                break;
+            }
+            if (bestFit.slice.getByteBuffer().remaining() > (RECLAIM_FACTOR * size)) {
```
sanastas: Are ByteBuffers clear/rewind before getting into the freeList so the remaining() is their full size...?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -33,10 +43,14 @@
     private final AtomicInteger idGenerator = new AtomicInteger(1);
 
     // free list of Slices which can be reused - sorted by buffer size, then by unique hash
-    private Comparator<Map.Entry<Long, Slice>> comparator =
-            Comparator.<Map.Entry<Long, Slice>, Integer>comparing(p -> p.getValue().getByteBuffer().remaining())
-            .thenComparing(Map.Entry::getKey);
-    private final ConcurrentSkipListSet<Map.Entry<Long,Slice>> freeList = new ConcurrentSkipListSet<>(comparator);
+    private final ConcurrentSkipListSet<FreeChuck> freeList = new ConcurrentSkipListSet<>((x, y) -> {
```
sanastas: Please comment how it is sorted by size. From small to big or the opposite...

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
sanastas: Also I am confused now with this RECLAIM_FACTOR (currently 2). Does it mean that if we got a buffer but it twice bigger than we need we do not take it? If so, first please comment with explanation. Second, if so, why is it named RECLAIM_FACTOR?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -57,83 +71,83 @@
...
+        while (!freeList.isEmpty()) {
+            myDummy.length = size;
+            FreeChuck bestFit = freeList.higher(myDummy);
+            if (bestFit == null) {
+                break;
+            }
+            if (bestFit.slice.getByteBuffer().remaining() > (RECLAIM_FACTOR * size)) {
+                break;     // all remaining buffers are too big
+            }
+            if (freeList.remove(bestFit)) {
```
sanastas: freeList is a concurrent structure, is it that many threads can get the same bestFit chunk as an output of line 106? Is it that only one will succeed to remove the bestFit and get the allocation? Others will allocate a new space, although they might have yet another allocation from freeList, am I wrong?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -142,43 +156,53 @@ public Slice allocateSlice(int size) {
     @Override
     public void free(ByteBuffer bb) {
         allocated.addAndGet(-(bb.remaining()));
-        if (stats != null) stats.release(bb);
-        freeList.add(
-            new AbstractMap.SimpleImmutableEntry<>(
-                freeCounter.getAndIncrement(), new Slice(INVALID_BLOCK_ID,bb)));
+        if (stats != null) {
```
sanastas: Do we still use free() of ByteBuffers?

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -142,43 +156,53 @@ public Slice allocateSlice(int size) {
...
-    if (stats != null) stats.release(slice.getByteBuffer());
-    freeList.add(new AbstractMap.SimpleImmutableEntry<>(freeCounter.getAndIncrement(), slice));
-  }
+    @Override
+    public void freeSlice(Slice slice) {
+        allocated.addAndGet(-(slice.getByteBuffer().remaining()));
+        if (stats != null) {
+            stats.release(slice.getByteBuffer());
+        }
+        freeList.add(new FreeChuck(freeCounter.getAndIncrement(), slice.getByteBuffer().remaining(), slice));
```
sanastas: I am still worried from where do we get this ByteBuffer and can it's position be not 0? If the ByteBuffer is specially made only for release, please comment about this...

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
sanastas: If this free/allocate are here just because of the interface, maybe to throw an unimplemented exception? Just as a sanity check...

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -0,0 +1,384 @@
...
+    enum ValueResult {
+        SUCCESS, FAILURE, MOVED
+    }
+
+    private static final int LOCK_MASK = 0x3;
+    public static final int VALUE_HEADER_SIZE = 4;
+
+    private static Unsafe unsafe = UnsafeUtils.unsafe;
+
+    //One duplication
```
sanastas: Where this comment should belong?

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * Similar to #getValueByteBufferNoHeaderPrivate(ByteBuffer bb), without the assumption of a thread local
+     * environment.
+     * Creates two new objects instead of one.
+     *
+     * @param bb the Byte Buffer of the value
+     * @return the byte buffer of the value without its header.
+     */
+    static ByteBuffer getValueByteBufferNoHeader(ByteBuffer bb) {
```
sanastas: Where is it needed?

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -0,0 +1,384 @@
...
+            bb = s.getByteBuffer();
+            bb.putInt(bb.position(), LOCKED.value);
+            int valueBlockAndLength =
+                    (s.getBlockID() << VALUE_BLOCK_SHIFT) | ((capacity + VALUE_HEADER_SIZE) & VALUE_LENGTH_MASK);
+            assert chunk.casEntriesArrayLong(lookUp.entryIndex, Chunk.OFFSET.VALUE_REFERENCE, lookUp.valueReference,
+                    UnsafeUtils.intsToLong(valueBlockAndLength, bb.position()));
+            chunk.unpublish();
+        }
+        ByteBuffer dup = getValueByteBufferNoHeader(bb);
+        // It seems like I have to change bb here or else I expose the header to the user
```
sanastas: Remove "It seems like I"

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -0,0 +1,384 @@
...
+     * @param lookUp        has the Slice of the value and the entry index
+     * @param newVal        the new value to write
+     * @param serializer    value serializer
+     * @param memoryManager the memory manager to free a slice with is not needed after the value moved
+     * @param <V>           the type of the value
+     * @return {@code SUCCESS} if the value was written off-heap successfully,
+     * {@code FAILURE} if the value is deleted (cannot be overwritten),
+     * {@code RETRY} if the value was moved, or if the chuck is frozen/released (prevents the moving of the value).
+     */
+    public static <V> ValueResult put(Chunk<?, V> chunk, Chunk.LookUp lookUp, V newVal, OakSerializer<V> serializer,
```
sanastas: Is it used only for putIfPresent? Right? Please comment

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yoavz1997: The buffer's position is usually not 0

sanastas: From where do we get this BB?

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -13,29 +13,31 @@
 
 public class OakRValueBufferImpl implements OakRBuffer {
 
-    private Handle handle;
+    private ByteBuffer bb;
```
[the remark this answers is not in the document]
yoavz1997: Yes, all the Byte Buffers except the ones passed to the user functions always include the header.

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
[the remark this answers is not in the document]
yoavz1997: OakRValueBufferImpl is used with regular iterators, and OakRReference with Stream Iterators.

core/src/main/java/com/oath/oak/ValueUtils.java
yoavz1997: It is not used, so I deleted it.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yoavz1997: It is now deprecated. It is used in some tests so I left it in the code.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yoavz1997: Thread re-execute this loop. I added this explanation in a comment.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yoavz1997: It is deprecated, but not deleted since it is used in some tests.

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
yoavz1997: From the user, when it frees its buffer.

core/src/main/java/com/oath/oak/OakWBufferImpl.java
```diff
@@ -8,112 +8,123 @@
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.function.Function;
 
 public class OakWBufferImpl implements OakWBuffer {
 
-    private final Handle handle;
+    private final ByteBuffer bb;
```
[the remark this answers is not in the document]
yoavz1997: I explained the setting in which OakWBuffers can be used (where a write lock is already acquired) in a comment.

core/src/main/java/com/oath/oak/Slice.java
```diff
@@ -19,11 +21,19 @@ public Slice(int blockID, ByteBuffer buffer) {
         this.buffer = buffer;
     }
 
+    public Slice(int blockID, int position, int length, MemoryManager memoryManager) {
+        this(blockID, memoryManager.getByteBufferFromBlockID(blockID, position, length).duplicate());
```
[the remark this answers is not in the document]
yoavz1997: I removed this duplicate()

PR #56, opened 2019-10-07
Synchrobench compare bug

yoavz1997:
The main fix is in `MyBufferOak`.
The arguments' order was wrong and resulted in a `NullPointerException` when running synchrobench.

The other code is PIACIP which is needed for the one of the paper's tests.


benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBufferOak.java
```diff
@@ -69,7 +69,7 @@ public int compareKeyAndSerializedKey(MyBuffer key, ByteBuffer serializedKey) {
             int cap1 = key.buffer.capacity();
             int pos2 = serializedKey.position();
             int cap2 = serializedKey.getInt(pos1);
-            return compare(serializedKey, pos1 + Integer.BYTES, cap1, key.buffer, pos2, cap2);
+            return compare(key.buffer, pos1, cap1, serializedKey, pos2 + Integer.BYTES, cap2);
```
sanastas: Initially serializedKey came with pos1 + Integer.BYTES, now it comes with pos2. Can you explain why is it correct?

sanastas:
I have reviewed the new code and the changes looks good to me. I am going to approve the merge to free way for other commits. If Eran has any comments we might add the fixes just after.

PR #58, opened 2019-10-17
Squashed mm2

yoavz1997:


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -387,13 +392,68 @@ long getValueReference(int entryIndex) {
...
+            int offHeapVersion = operator.getOffHeapVersion(valueSlice);
+            casEntriesArrayInt(lookUp.entryIndex, OFFSET.VALUE_VERSION, entryVersion, offHeapVersion);
+            lookUp.version = offHeapVersion;
+            return offHeapVersion;
+        } finally {
+            unpublish();
+        }
+    }
+
+    NovaValueUtils.NovaResult finalizeDeletion(LookUp lookUp) {
```
sanastas: Yet another Result? Is there an option to collide it all together?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -387,13 +392,68 @@ long getValueReference(int entryIndex) {
...
+        return valueReference;
+    }
+
+    int completeLinking(LookUp lookUp) {
+        int entryVersion = lookUp.version;
+        // no need to complete a thing
+        if (entryVersion > INVALID_VERSION) {
+            return entryVersion;
+        }
+        if (!publish()) {
```
sanastas: Publish without parameters... It is already so far from what we had... Why to publish if nobody can help? It is not the problem of this code I am just frustrated... I hope I will be able to make the rebalance non-blocking again...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -427,18 +487,23 @@ LookUp lookUp(K key) {
...
                 }
-                if (ValueUtils.isValueDeleted(valueSlice)) {
+                NovaValueUtils.NovaResult result = operator.isValueDeleted(valueSlice, version[0]);
+                if (result == TRUE) {
                     // There is a deleted value associated with the given key
-                    return new LookUp(null, valueReference, curr);
+                    return new LookUp(null, valueReference, curr, version[0]);
                 }
-                return new LookUp(valueSlice, valueReference, curr);
+                // If result == RETRY, I ignore it, since it will be discovered later down the line as well
```
sanastas: Please rephrase without "I"...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -465,11 +530,13 @@ else if (cmp == 0) {
          */
         final long valueReference;
         final int entryIndex;
+        int version;
```
sanastas: Excellent comments about slice and version, please explain about version as well....

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -639,72 +708,26 @@ int getMaxItems() {
...
-        // or PIACIP compute happened, which should return false
-        return false;
-    }
-
-    /**
-     * used by put/putIfAbsent/remove and rebalancer
-     */
-    private boolean pointToValueCAS(OpData opData) {
-        if (casEntriesArrayLong(opData.entryIndex, OFFSET.VALUE_REFERENCE, opData.oldValueReference,
+    NovaValueUtils.NovaResult linkValue(OpData opData) {
```
sanastas: Does linkValue() replace pointToValue()? It is hard to find resemblance....

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -639,72 +708,26 @@ int getMaxItems() {
...
-                // previously a remove
-                statistics.incrementAddedCount();
-                externalSize.incrementAndGet();
-            } else if (opData.oldValueReference != INVALID_VALUE_REFERENCE && opData.newValueReference == INVALID_VALUE_REFERENCE) {
-                // removing
-                statistics.decrementAddedCount();
-                externalSize.decrementAndGet();
-            }
-            return true;
+            return FALSE;
```
sanastas: How does it returns FALSE? Is NovaValueUtils.NovaResult an enum?

core/src/main/java/com/oath/oak/Chunk.java
sanastas: Please explain it to me personally

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -196,13 +205,12 @@ boolean casEntriesArrayLong(int item, OFFSET offset, long expected, long value)
...
     }
 
-
     /**
      * write key in slice
      **/
     private void writeKey(K key, int ei) {
         int keySize = keySerializer.calculateSize(key);
-        Slice s = memoryManager.allocateSlice(keySize);
+        Slice s = memoryManager.allocateSlice(keySize, true);
```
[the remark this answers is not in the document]
yoavz1997: I currently keep track of the number of keys and values allocated.
I used it when measuring the memory consumption.
Should I remove it?

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Yes, I'll do that!

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: Yes

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -104,7 +103,11 @@ void close() {
         // once reference count is zeroed, the map meant to be deleted and should not be used.
         // reference count will never grow again
         if (res == 0) {
-            memoryManager.close();
+            try {
+                memoryManager.close();
+            } catch (IOException e) {
```
sanastas: Why/what exception can happen here?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -238,6 +237,7 @@ private void connectToChunkList(List<Chunk<K, V>> engaged, List<Chunk<K, V>> chi
             }
             // chunk is not in list (someone else already updated list), so we're done with this part
             if (curr == null) {
+                //TODO Never reached
```
sanastas: Why to add it again?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
 
-        Chunk<K, V> c = findChunk(key); // find chunk matching key
-        Chunk.LookUp lookUp = c.lookUp(key);
-        // If there is a matching value reference for the given key, and it is not marked as deleted, then this put
-        // changes the slice pointed by this value reference.
-        if (lookUp != null && lookUp.valueSlice != null) {
-            Result<V> res = ValueUtils.exchange(c, lookUp, value, transformer, valueSerializer, memoryManager);
-            if (res.operationResult == SUCCESS) {
-                return res.value;
+        while (true) {
```
sanastas: Is it all that you replaced the recursive call with a loop? Why so?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-            rebalance(c.creator());
-            return put(key, value, transformer);
-        }
-        if (state == Chunk.State.FROZEN || state == Chunk.State.RELEASED) {
-            rebalance(c);
-            return put(key, value, transformer);
-        }
+            // if chunk is frozen or infant, we can't add to it
+            // we need to help rebalancer first, then proceed
+            if (inTheMiddleOfRebalance(c) || finalizeDeletion(c, lookUp)) {
```
sanastas: 👍

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
-            ei = c.allocateEntryAndKey(key);
-            if (ei == INVALID_ENTRY_INDEX) {
-                rebalance(c);
-                return put(key, value, transformer);
-            }
-            int prevEi = c.linkEntry(ei, key);
-            if (prevEi != ei) {
-                ei = prevEi;
-                oldReference = c.getValueReference(prevEi);
-                // We can use the old reference *only* when it is DELETED_VALUE,
```
sanastas: Move this comment back

core/src/main/java/com/oath/oak/Chunk.java
yoavz1997: This Result replaces the Result of ValueUtils (It is used inside the compound class Result<V>), so no need to consolidate it.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
+                c.releaseValue(newValueReference);
+                c.unpublish();
+            } else {
+                c.unpublish();
+                checkRebalance(c);
+                return null;
+            }
         }
-
-        return null;
```
sanastas: Don't we need to return null at the end?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
+                }
+                continue;
+            }
 
+            // if chunk is frozen or infant, we can't add to it
+            // we need to help rebalancer first, then proceed
+            if (inTheMiddleOfRebalance(c) || finalizeDeletion(c, lookUp)) {
+                continue;
+            }
+            int oldVersion = lookUp == null ? INVALID_VERSION : -Math.abs(lookUp.version);
```
sanastas: Please add comment explaining why do we take the minus of the absolute value. Why minus indeed?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -317,250 +340,240 @@ V put(K key, V value, Function<ByteBuffer, V> transformer) {
...
                 }
             }
-        }
 
-        long newValueReference = c.writeValue(value); // write value in place
+            int[] version = new int[1];
+            long newValueReference = c.writeValue(value, version); // write value in place
 
-        Chunk.OpData opData = new Chunk.OpData(Operation.PUT_IF_ABSENT, ei, newValueReference, oldReference, null);
+            Chunk.OpData opData = new Chunk.OpData(ei, INVALID_VALUE_REFERENCE, newValueReference, oldVersion,
```
sanastas: Who use this op data ?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -639,72 +714,26 @@ int getMaxItems() {
...
-            return true;
+            return FALSE;
+        }
+        casEntriesArrayInt(opData.entryIndex, OFFSET.VALUE_VERSION, opData.oldVersion, opData.newVersion);
+        if (opData.oldValueReference == INVALID_VALUE_REFERENCE) {
+            assert opData.newValueReference != INVALID_VALUE_REFERENCE;
+            statistics.incrementAddedCount();
+            externalSize.incrementAndGet();
+        } else {
+            assert false;
```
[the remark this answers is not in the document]
yoavz1997: I removed it from the code.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: When stating that MemoryManager implements Closable, it told me to add this try-catch clause.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I think it is left over from merging branches, I'll remove it.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: It is easier on the call stack (less memory and less nested calls), and it is easier to debug if necessary (can follow the loop manually instead of having a StackOverflowException).

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I added an if-else clause.
In case the CAS succeed (the result is TRUE), we return ``null``.
Otherwise, the value is freed, the chunk is rebalanced, and the operation is restarted.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: I added comments where they are needed.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: The linking operation in line 490

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -574,100 +615,141 @@ V remove(K key, V oldValue, Function<ByteBuffer, V> transformer) {
...
-            return null;
+        while (true) {
+            Chunk<K, V> c = findChunk(key); // find chunk matching key
+            Chunk.LookUp lookUp = c.lookUp(key);
+
+            if (lookUp != null && lookUp.valueSlice != null) {
+                if (updateVersionAfterLinking(c, lookUp)) {
+                    continue;
+                }
+                NovaValueUtils.NovaResult res = operator.compute(lookUp.valueSlice, computer, lookUp.version);
```
sanastas: Operator appears to be the "handle" to make the operations on the existing values like compute/put/exchange/etc. However this is very opaque. Especially because it is one operator providing "all the services". It would be much clearer if we have "something" related to a specific value on which compute/put/exchange/etc. operations can be applied. It can be LookUp, Slice, ...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -786,12 +850,13 @@ boolean replace(K key, V oldValue, V newValue, Function<ByteBuffer, V> valueDese
                 return false;
             }
 
-            ValueUtils.ValueResult result = ValueUtils.compareExchange(c, lookUp, oldValue, newValue,
-                    valueDeserializeTransformer, valueSerializer,
-                    memoryManager);
-            if (result != MOVED) {
-                return result == SUCCESS;
+            // res can be null if handle was deleted between lookup and the next call
```
sanastas: Remove "handle", also please go over all the code and look for "handle"

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -986,18 +1073,30 @@ private void initAfterRebalance() {
...
+            int[] valueVersion = new int[1];
+            Map.Entry<Integer, Long> value = null;
             if (needsValue) {
-                currentValue = state.getChunk().getValueSlice(state.getIndex());
-                if (currentValue == null || ValueUtils.isValueDeleted(currentValue)) {
+                valueReference = state.getChunk().getValueReferenceAndVersion(state.getIndex(), valueVersion);
+                if (valueReference != INVALID_VALUE_REFERENCE) {
+                    valueVersion[0] = state.getChunk().completeLinking(new LookUp(null, valueReference,
+                            state.getIndex(), valueVersion[0]));
+                    // TODO: check the first clause
```
sanastas: What about TODO?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -973,8 +1057,11 @@ private void initAfterRebalance() {
         /**
          * Advances next to higher entry.
          * Return previous index
+         *
+         * @return The first long is the key's reference, the integer is the value's version and the second long is
+         * the value's reference. If {@code needsValue == false}, then the value of the map entry is {@code null}.
          */
-        Map.Entry<ByteBuffer, Slice> advance(boolean needsKey, boolean needsValue) {
+        Map.Entry<Long, Map.Entry<Integer, Long>> advance(boolean needsValue) {
```
sanastas: Map Entry on top of the other Map Entry? Here it looks to me a good place to add a class... Or can we use the existing Result?

core/src/main/java/com/oath/oak/MemoryManager.java
```diff
@@ -1,56 +1,23 @@
...
-    }
+    boolean isClosed();
 
-    // how many memory is allocated for this OakMap
-    public long allocated() {
-        return valuesMemoryAllocator.allocated();
-    }
+    long allocated();
 
-    // allocateSlice is used when the blockID (of the block from which the ByteBuffer is allocated)
```
sanastas: Please bring back those or other comments explaining the interface...

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -97,14 +99,14 @@ public OakNativeMemoryAllocator(long capacity) {
 
     @Deprecated
     public ByteBuffer allocate(int size) {
-        return allocateSlice(size).getByteBuffer();
+        return allocateSlice(size, true).getByteBuffer();
```
sanastas: Is it used only for tests? Please say so. Can you use KEY enum instead of true?

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: The inner map entry is the version of the value and its reference.
I could create a class which is a pair of int and long if needs be.

core/src/main/java/com/oath/oak/MemoryManager.java
yoavz1997: These comments do not add any information. We only allocate slices, and we use slices for both keys and values.

core/src/main/java/com/oath/oak/InternalOakMap.java
yoavz1997: Will be fixed (I also added a comment)

core/src/main/java/com/oath/oak/OakBlockMemoryAllocator.java
```diff
@@ -6,27 +6,32 @@
...
 
+import java.nio.ByteBuffer;
+
 // this is the interface to be implemented to replace the OakNativeMemoryAllocator
 // this is about allocation of a new ByteBuffer (which need to be DirectByteBuffer
 // in order to continue supporting off-heap)
 // allocator is also getting a ByteBuffer to reuse the memory, given ByteBuffer is
 // no longer in use by any thread
-public interface OakMemoryAllocator {
+public interface OakBlockMemoryAllocator {
```
sanastas: In the comment above explain about Slice

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -526,10 +522,10 @@ public OakRBuffer get(K key) {
             return m.internalOakMap.get(key);
         }
 
-        public void remove(K key) {
+        public boolean remove(K key) {
```
sanastas: Is it ZC API or regular ConcurrentNavigableMap interface?

core/src/main/java/com/oath/oak/OakMapBuilder.java
```diff
@@ -43,6 +44,7 @@ public OakMapBuilder() {
         this.chunkMaxItems = Chunk.MAX_ITEMS_DEFAULT;
         this.memoryCapacity = MAX_MEM_CAPACITY;
         this.memoryAllocator = null;
+        this.operator = new ValueUtilsImpl();
```
sanastas: I do not think builder should create operator... If it is one per Oak it should be in InternalOakMap or a singleton

core/src/main/java/com/oath/oak/OakMap.java
```diff
@@ -52,12 +52,12 @@
 
     // internal constructor, to create OakMap use OakMapBuilder
     OakMap(K minKey, OakSerializer<K> keySerializer, OakSerializer<V> valueSerializer, OakComparator<K> oakComparator,
-           int chunkMaxItems, MemoryManager mm) {
+           int chunkMaxItems, MemoryManager mm, ValueUtils operator) {
```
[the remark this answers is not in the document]
yoavz1997: It is not a singleton if it does not have to be. Currently it is a singleton since it has no state, but if in the future some implementation would add a state, then it would help.
Moreover, it becomes more testable this way since mocks can be used if necessary.

core/src/main/java/com/oath/oak/OakMap.java
yoavz1997: In the regular API, remove returns the previously associated value with the given key.
In the ZC API, I changed the signature to return ``boolean`` instead of plain ``void``

core/src/main/java/com/oath/oak/OakMapBuilder.java
yoavz1997: I think that just as the memory manager is created in the Builder, so should the valueOperator.
I think, however, that there should be a method to set the valueOperator, so making the access to the off-heap lock-free would be seamless.

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -11,115 +11,147 @@
...
 
+import static com.oath.oak.Chunk.BLOCK_ID_LENGTH_ARRAY_INDEX;
+import static com.oath.oak.Chunk.POSITION_ARRAY_INDEX;
+import static com.oath.oak.Chunk.VALUE_BLOCK_SHIFT;
+import static com.oath.oak.Chunk.VALUE_LENGTH_MASK;
+import static com.oath.oak.ValueUtils.ValueResult.*;
+
+// remove header
 public class OakRValueBufferImpl implements OakRBuffer {
+    private long valueReference;
```
sanastas: Please add a word why each field is needed. E.g. keyreference is needed in order to retry (I assume...)

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -134,31 +166,47 @@ public double getDouble(int index) {
...
+        end(s);
     }
 
-    private void start() {
-        // if the value was moved a ConcurrentModificationException is thrown
-        ValueUtils.ValueResult res = ValueUtils.lockRead(bb);
-        if (res != ValueUtils.ValueResult.SUCCESS) {
+    private void start(Slice valueSlice) {
+        ValueUtils.ValueResult res = operator.lockRead(valueSlice, version);
+        if (res == FALSE) {
```
sanastas: How lock can return false? If value was deleted? Please comment

core/src/main/java/com/oath/oak/OakWBuffer.java
```diff
@@ -22,14 +22,6 @@
      */
     int capacity() throws NullPointerException;
 
-    /**
-     * Going to be deprecated!!!!!!!!!
-     *
-     * @return the actual ByteBuffer
-     */
-    @Deprecated
-    ByteBuffer getByteBuffer();
```
sanastas: IT WONT COMPILE WITH DRUID!!!! PUT IT BACK! We should first resolve how to work without in in Druid....

core/src/main/java/com/oath/oak/OldMemoryManager.java
```diff
@@ -0,0 +1,56 @@
...
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.oath.oak;
+
+import com.oath.oak.NativeAllocator.OakNativeMemoryAllocator;
+import java.nio.ByteBuffer;
+
+
+public class OldMemoryManager {
```
sanastas: What does it mean "OldMemoryManager"? Name it anyhow different. Handle-based-Meory-Manager? It would be the best to use it for "Empty Memory Manager" as Eddie asked for...

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -1,368 +1,58 @@
...
     enum ValueResult {
-        SUCCESS, FAILURE, MOVED
-    }
-
-    private static final int LOCK_MASK = 0x3;
-    public static final int VALUE_HEADER_SIZE = 4;
-
-    private static Unsafe unsafe = UnsafeUtils.unsafe;
-
-    /**
```
sanastas: I do not understand why do you remove the comments when moving to interface... Each interface must be easily understandable. In my philosophy there is no such a thing as "not needed comment", even if a comment is obvious. Unless the comment is wrong, but then just correct the comment... Please add comments to the interface

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -1,368 +1,58 @@
...
-        if (lockRead(bb) != SUCCESS) {
-            return null;
-        }
+    int getOffHeapVersion(Slice s);
 
-        T transformation = transformer.apply(getValueByteBufferNoHeader(bb).asReadOnlyBuffer());
-        unlockRead(bb);
-        return transformation;
-    }
+    int NOVA_HEADER_SIZE = 4;
```
sanastas: Should it be named with "NOVA". ValueUtils should work regardless to memory management...

yoavz1997: This is the size of the version. It does not have to necessarily be connected to Nova, but it has to be connected to versions (as does ValueUtils). I renamed it to ``VERSION_SIZE``.

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
yoavz1997: I added comments to the API of ValueUtils (in this case, the documentation of lockRead explains in what case FALSE can be returned).

core/src/main/java/com/oath/oak/OakWBuffer.java
yoavz1997: I put it back

core/src/main/java/com/oath/oak/ValueUtilsImpl.java
```diff
@@ -0,0 +1,329 @@
...
+        }
+        computer.accept(new OakWBufferImpl(s, this));
+        unlockWrite(s);
+        return TRUE;
+    }
+
+    @Override
+    public <V> Result<V> remove(Slice s, MemoryManager memoryManager, int version, V oldValue,
+                                Function<ByteBuffer, V> transformer) {
+        // No need to check the old value
```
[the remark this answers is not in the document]
yoavz1997: What locks?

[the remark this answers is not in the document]
yoavz1997: What locks?

PR #64, opened 2019-11-09
Removing ThreadLocal from ThreadIndexCalculator

sanastas:
Removing usage of ThreadLocal from ThreadIndexCalculator. The tests are passing. Synchrobench zero-copy gets-only benchmark show a small 5% performance improvement.

### DIFF
--- a/core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
+++ b/core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
@@ -1,41 +1,62 @@
 
 package com.oath.oak;
 
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class ThreadIndexCalculator {
 
     public static final int MAX_THREADS = 32;
     private static final int INVALID_THREAD_ID = -1;
-    private ThreadLocal<Integer> local = ThreadLocal.withInitial(() -> INVALID_THREAD_ID);
-    private AtomicInteger[] indices = new AtomicInteger[MAX_THREADS];
+    // Long for correctness and anti false-sharing
+    private AtomicLong[] indices = new AtomicLong[MAX_THREADS];
 
     private ThreadIndexCalculator() {
         for (int i = 0; i < MAX_THREADS; ++i) {
-            indices[i] = new AtomicInteger(INVALID_THREAD_ID);
+            indices[i] = new AtomicLong(INVALID_THREAD_ID);
         }
     }
 
+    private int getExistingIndex(long threadID){
+      int iterationCnt = 0;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -1,41 +1,62 @@
...
         }
     }
 
+    private int getExistingIndex(long threadID){
+      int iterationCnt = 0;
+      int currentIndex = ((int)threadID) % MAX_THREADS;
+      long currentThreadID = indices[currentIndex].get();
+      while (currentThreadID != threadID) {
+        if (currentThreadID == INVALID_THREAD_ID) {
+          return -1*currentIndex;
```
Eshcar: what does a negative number indicates?

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -1,41 +1,62 @@
...
+          // due to multiplying by -1 check this special array element
+          if (tid == indices[0].get()) {
+            return threadIdx;
+          }
         }
-        int tid = (int) Thread.currentThread().getId();
-        int i = tid % MAX_THREADS;
+        int i = threadIdx*-1;
         while (!indices[i].compareAndSet(INVALID_THREAD_ID, tid)) {
             //TODO get out of loop sometime
```
Eshcar: why do you do the CAS externally and not in the `getExistingIndex` method? it will also save you the redundant 0-case check

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
```diff
@@ -1,41 +1,62 @@
...
-        local.set(i);
         return i;
     }
 
     public void releaseIndex() {
-        indices[local.get()].set(INVALID_THREAD_ID);
-        local.set(INVALID_THREAD_ID);
+        long tid = Thread.currentThread().getId();
+        int index = getExistingIndex(tid);
+        assert index >=0 ;
```
Eshcar: you may want to throw exception to be able to quickly identify bugs in which a thread releases its index twice.

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
sanastas: negative output indicates that a new index need to be created for this thread id
a comment is added

core/src/main/java/com/oath/oak/ThreadIndexCalculator.java
sanastas: getExistingIndex is meant to check for an *existing* index. It used for example in releaseIndex and doesn't need to create a new index there.

PR #67, opened 2019-12-25
Fixing a bug in Descending Scan that the same keys were returned multiple times

sanastas:
Added a test that failed due to existing problem. The test doesn't fail after the fix. All tests pass. Performance evaluation show no degradation. 

### DIFF
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -1185,6 +1185,8 @@ public class Chunk<K, V> {
                     if (comparator.compareSerializedKeys(tmpBBprevAnchorKey, readSecondKey(next)) > 0) {
                         stack.push(next);
                         next = getEntryFieldInt(next, OFFSET.NEXT);
+                    } else {
+                        break;
                     }
                 }
             }
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -6,6 +6,7 @@
 
 package com.oath.oak;
 
+import com.oath.oak.NativeAllocator.OakNativeMemoryAllocator;
 import sun.misc.Unsafe;
 
 import java.nio.ByteBuffer;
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -241,7 +242,29 @@ ByteBuffer readKey(int entryIndex) {
         int keyPosition = keyArray[POSITION_ARRAY_INDEX];
         int length = keyArray[BLOCK_ID_LENGTH_ARRAY_INDEX] & KEY_LENGTH_MASK;
 
-        return memoryManager.getByteBufferFromBlockID(blockID, keyPosition, length);
+        return memoryManager.getByteBufferFromBlockID(blockID, keyPosition, length,
+            OakNativeMemoryAllocator.FIRST_THREAD_BUFFER);
```
Eshcar: is this first/second thread buffer related to the descending scan bug?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -241,7 +242,29 @@ ByteBuffer readKey(int entryIndex) {
...
+    ByteBuffer readSecondKey(int entryIndex) {
+        if (entryIndex == Chunk.NONE) {
+            return null;
+        }
+
+        long keyReference = getKeyReference(entryIndex);
+        int[] keyArray = longToInts(keyReference);
+        int blockID = keyArray[BLOCK_ID_LENGTH_ARRAY_INDEX] >> KEY_BLOCK_SHIFT;
+        int keyPosition = keyArray[POSITION_ARRAY_INDEX];
+        int length = keyArray[BLOCK_ID_LENGTH_ARRAY_INDEX] & KEY_LENGTH_MASK;
```
Eshcar: code duplication 256-264 should go through 3rd method that gets first/second as parameter

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -1156,10 +1175,28 @@ private void findNewNextInStack() {
             }
         }
 
+        private void pushToStack(boolean compareWithPrevAnchor) {
```
Eshcar: please add documentation (explaining the steps)
it is difficult to reason about why this code is correct

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -1156,10 +1175,28 @@ private void findNewNextInStack() {
...
+            while (next != Chunk.NONE) {
```
Eshcar: check if changing this to next != prevAnchor will save the keys bytebuffer comparisons

PR #68, opened 2020-01-16
Desc scan v03

sanastas:
New fix for the correctness of descending scans


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -1107,6 +1103,8 @@ public int next() {
         private final K from;
         private boolean inclusive;
 
+        static final int SKIP_ENTRIES_FOR_BIGGER_STACK = 1; // 1 is the lower possible value
```
Eshcar: lowest(?) possible value

PR #69, opened 2020-01-21
CORRECTNESS: Making OAK robust to changes of the buffer byte order by the user

liran-funaro:
Making OAK robust to changes of the buffer byte order by the user

* Fix endianness in conversion between int array and long.
  Separate two different cases that were using the same method incorrectly.
* The version is now written to the buffer in the ValueUtils class, instead of in the NovaManager class.
* MemoryManager interface now exposes the current version. Future: should consider moving the version logic from the memory manager.
* All writes/reads of version, lock state and any header data are now handled via the "unsafe" interface and uses the native byte order.
* Fix OakViewTests: The test wrongly assumed the byte representation of the internal buffers (specifically the endianness). Fix test to be compatible to any byte representation.

### DIFF
--- a/core/src/main/java/com/oath/oak/ValueUtilsImpl.java
+++ b/core/src/main/java/com/oath/oak/ValueUtilsImpl.java
@@ -7,8 +7,6 @@ import java.nio.ByteBuffer;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import static com.oath.oak.Chunk.VALUE_BLOCK_SHIFT;
-import static com.oath.oak.Chunk.VALUE_LENGTH_MASK;
 import static com.oath.oak.ValueUtilsImpl.LockStates.DELETED;
 import static com.oath.oak.ValueUtilsImpl.LockStates.FREE;
 import static com.oath.oak.ValueUtilsImpl.LockStates.LOCKED;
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/oath/oak/ValueUtilsImpl.java
```diff
@@ -94,15 +89,17 @@ public void unsafeBufferToIntArrayCopy(ByteBuffer bb, int srcPosition, int[] dst
...
         Slice s = lookUp.valueSlice;
-        putInt(s, getLockLocation(), MOVED.value);
+        setLockState(s, MOVED);
         memoryManager.releaseSlice(s);
         s = memoryManager.allocateSlice(capacity + getHeaderSize(), MemoryManager.Allocate.VALUE);
-        putInt(s, getLockLocation(), LOCKED.value);
+        initHeader(s, memoryManager.getCurrentVersion(), LOCKED);
         int valueBlockAndLength =
                 (s.getBlockID() << VALUE_BLOCK_SHIFT) | ((capacity + getHeaderSize()) & VALUE_LENGTH_MASK);
+        // TODO Liran: should this really be inside an assert?!
```
sanastas: Before the merge please check that no "TODO Liran" comments are inside :)

sanastas:
The change looks good. Will be overwrote a bit with the coming refactoring.

PR #70, opened 2020-01-21
DESIGN: Use loop instead of redundant recursion to retry a transform operation

liran-funaro:

### DIFF
--- a/core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
+++ b/core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
@@ -182,7 +182,8 @@ public class OakRValueBufferImpl implements OakRBuffer {
         if (transformer == null) {
             throw new NullPointerException();
         }
-        while (true) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -182,14 +182,16 @@ public double getDouble(int index) {
...
             throw new NullPointerException();
         }
-        Result<T> result = valueOperator.transform(getValueSlice(), transformer, version);
-        if (result.operationResult == FALSE) {
-            throw new ConcurrentModificationException();
-        } else if (result.operationResult == RETRY) {
-            lookupValueReference();
-            transform(transformer);
+        while (true) {
+            Result<T> result = valueOperator.transform(getValueSlice(), transformer, version);
```
sanastas: 1) Why the PR name starts with "correctness"? :)
2) Please add an assertion for infinite loop

PR #71, opened 2020-01-21
PERF: Removing reference to block array in the memory allocator after closing

liran-funaro:
This will prevent unwanted access to the released block which will obviate the “isclosed()” check.

### DIFF
--- a/core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
+++ b/core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
@@ -204,7 +204,7 @@ public class OakNativeMemoryAllocator implements OakBlockMemoryAllocator {
             return;
         }
 
-        // Release the hold of the block array and return it the the provider.
+        // Release the hold of the block array and return it the provider.
         Block[] b = blocksArray;
         blocksArray = null;
 
--- a/core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
+++ b/core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
@@ -40,7 +40,7 @@ public class OakNativeMemoryAllocator implements OakBlockMemoryAllocator {
     public static final int INVALID_BLOCK_ID = 0;
 
     // mapping IDs to blocks allocated solely to this Allocator
-    private final Block[] blocksArray;
+    private Block[] blocksArray;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -203,8 +203,16 @@ public void close() {
         if (!closed.compareAndSet(false, true)) {
             return;
         }
+
+        // Release the hold of the block array and return it the the provider.
```
sanastas: typo: double "the"

core/src/main/java/com/oath/oak/NativeAllocator/OakNativeMemoryAllocator.java
```diff
@@ -203,8 +203,16 @@ public void close() {
...
+        Block[] b = blocksArray;
+        blocksArray = null;
+
+        // Reset "closed" to apply a memory barrier before actually returning the block.
```
sanastas: Good one 👍

core/src/main/java/com/oath/oak/OakRReference.java
```diff
@@ -124,12 +123,6 @@ public void unsafeCopyBufferToIntArray(int srcPosition, int[] dstArray, int coun
     }
 
     private ByteBuffer getTemporaryPerThreadByteBuffer() {
-        // need to check that the entire OakMap wasn't already closed with its memoryManager
```
sanastas: Please add a comment explaining the possibility of memory manager being closed and the fact that it will cause a NPE

liran-funaro:
Here is one example of performance improvement (Persist Benchmark from Druid incremental index).

Oak master
-------------------
```
Benchmark                        (indexType)  (rollup)  (rollupOpportunity)  (rowsPerSegment)  (schema)  Mode  Cnt         Score         Error  Units
IndexPersistBenchmark.persistV9          oak     false                 none           2000000     rollo  avgt   10  40381584.856 ? 1826817.619  us/op
```

Oak release block
-------------------
```
Benchmark                        (indexType)  (rollup)  (rollupOpportunity)  (rowsPerSegment)  (schema)  Mode  Cnt         Score         Error  Units
IndexPersistBenchmark.persistV9          oak     false                 none           2000000     rollo  avgt   10  39500250.052 ? 1925702.509  us/op
```

sanastas:
I am merging this pull request

PR #72, opened 2020-01-21
PERF: Avoid creating multiple instances of map entry when using a stream iterator

liran-funaro:

### DIFF
--- a/core/src/main/java/com/oath/oak/InternalOakMap.java
+++ b/core/src/main/java/com/oath/oak/InternalOakMap.java
@@ -1403,7 +1403,7 @@ class InternalOakMap<K, V> {
         }
     }
 
-    class EntryStreamIterator extends Iter<Map.Entry<OakRBuffer, OakRBuffer>> {
+    class EntryStreamIterator extends Iter<Map.Entry<OakRBuffer, OakRBuffer>> implements Map.Entry<OakRBuffer, OakRBuffer> {
 
         private OakRReference key = new OakRReference(memoryManager, KEY_HEADER_SIZE);
         private OakRReference value = new OakRReference(memoryManager, valueOperator.getHeaderSize());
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1418,7 +1418,22 @@ else if (res.operationResult == RETRY) {
             if (value == null) {
                 return null;
             }
-            return new AbstractMap.SimpleImmutableEntry<>(key, value);
+            return this;
```
sanastas: I would prefer a more comprehensive solution :) Like enhancement of the IterItem to extend Map.Entry... So it can be also used in the future cases...

But this works as well :)

liran-funaro:
Here is one example of performance improvement (Top N Benchmark from Druid incremental index).
Notice that we see an improvement in the first test.

Oak master
-------------------
```
Benchmark                                  (indexType)  (numSegments)  (rowsPerSegment)        (schemaAndQuery)  (threshold)  Mode  Cnt        Score       Error  Units
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000                 basic.A           10  avgt  100  6218370.345 ? 42883.469  us/op
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000       basic.numericSort           10  avgt  100  1337594.476 ? 38590.743  us/op
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000  basic.alphanumericSort           10  avgt  100  1219775.948 ?  9310.287  us/op
```

Oak Stream Iter
--------------------
```
Benchmark                                  (indexType)  (numSegments)  (rowsPerSegment)        (schemaAndQuery)  (threshold)  Mode  Cnt        Score       Error  Units
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000                 basic.A           10  avgt  100  6068787.596 ? 38920.114  us/op
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000       basic.numericSort           10  avgt  100  1376169.398 ? 62738.640  us/op
TopNBenchmark.querySingleIncrementalIndex          oak              1           2000000  basic.alphanumericSort           10  avgt  100  1206796.378 ? 12070.627  us/op
```

sanastas:
Thank you! I see the Druid improvement. I believe we should see also improvement in Oak's benchmark related to entry stream scan. Do we have the results of Oak's benchmarks? I cannot find them... If there is nothing surprising in the Oak's benchmarks and we do see improvement in the entry stream scan, I will merge the pull request.

PR #73, opened 2020-01-21
Fix bug: getByteBuffer() returned the a buffer that is positioned at the header instead of the data

liran-funaro:
Also, parameterize the benchmark output path.

### DIFF
--- a/core/src/main/java/com/oath/oak/OakWBufferImpl.java
+++ b/core/src/main/java/com/oath/oak/OakWBufferImpl.java
@@ -14,15 +14,15 @@ import java.nio.ByteOrder;
 public class OakWBufferImpl implements OakWBuffer {
 
     private ByteBuffer bb;
-    private final ValueUtils valueOperator;
 
     OakWBufferImpl(Slice s, ValueUtils valueOperator) {
-        this.bb = s.getByteBuffer();
-        this.valueOperator = valueOperator;
-    }
+        ByteBuffer buff = s.getByteBuffer();
+        int initPos = buff.position();
+        buff.position(initPos + valueOperator.getHeaderSize());
+        this.bb = buff.slice();
 
-    private int valuePosition() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/oath/oak/OakWBufferImpl.java
```diff
@@ -14,15 +14,15 @@
...
-    private final ValueUtils valueOperator;
 
     OakWBufferImpl(Slice s, ValueUtils valueOperator) {
-        this.bb = s.getByteBuffer();
-        this.valueOperator = valueOperator;
-    }
+        ByteBuffer buff = s.getByteBuffer();
+        int initPos = buff.position();
+        buff.position(initPos + valueOperator.getHeaderSize());
+        this.bb = buff.slice();
```
sanastas: I am against using slice() method here. We had sliced the block's ByteBuffer previously and we intentionally removed this code, as it caused lots of small objects floating around for example in compute-heavy workload. We've seen big performance improvement due to this change. OakWBuffer is there to protect all the accesses. I would suggest to move the position and slice() only in getByteBuffer() method.

core/src/main/java/com/oath/oak/OakWBufferImpl.java
```diff
@@ -14,15 +14,15 @@
...
 
-    private int valuePosition() {
-        return bb.position() + valueOperator.getHeaderSize();
+        // It is important to return the original buffer to its original state.
```
sanastas: If you never need to do anything with the header of the value, why it is not enough to change the BB position to be after the header for as long as OakWBuffer is in use? For other accesses this thread's block's position will be reset anyway... What do I miss?

sanastas:
A general comment: please do run Oak's regression benchmarks for each suggested change separately. After we are both satisfied with the code review the pull request will be merged only after we see no degradation in our regression benchmarks. Thanks!

PR #74, opened 2020-01-21
PERF: Store a temporary buffer in OakRReference to avoid looking for one for each request

liran-funaro:


sanastas:
I am sorry, I do not see what exactly is the gain of this tmp addition. What are we saving? Only the access to ThreadIndexCalculator? When TIC is not built on Atomic it should be not so much.

This fix limits Oak abilities in general, and we can get to it only if we see big change in performance, let say above 15%. Let us apply all other changes then run again with and without temp buffer and see how much do we gain.

liran-funaro:
Irrelevant at this point

sanastas:
I am closing this PR, because new OakBuffer interface will soon be out, as discussed internally.

sanastas:
OK. Liran just closed it :)

PR #76, opened 2020-01-28
CORRECTNESS: Validate keys' order only if from/to keys are available (not null)

liran-funaro:


sanastas:
The change looks good to me.

PR #81, opened 2020-03-09
Use predefined, fixed, upper and lower thresholds for the block pool

liran-funaro:
Use predefined, fixed, upper and lower thresholds for the block pool

Defined three new constants:
1. `PRE_ALLOC_BLOCKS`: Number of memory blocks to be pre-allocated.
2. `NEW_ALLOC_BLOCKS`: Number of memory blocks to be allocated at once when not enough blocks available.
3. `HIGH/LOW_RESERVED_BLOCKS`: Number of memory blocks to reserve in the pool for future use.

Modify `checkOakCapacity()` test to handle correctly any of the above parameters.


sanastas:
The code looks good to me. My biggest concern is about performance benchmark that we are going to see.

PR #82, opened 2020-03-09
Introduce new direct access API for OakRBuffer

liran-funaro:
The new API includes two methods:
1. getByteBuffer(): returns a ByteBuffer object that is bounded to the scope of this key/value
Used for faster repeated access to the buffer's data.

2. address(): return the memory address of the direct byte buffer at the relevant position
Used for an even faster access to the buffer's data (via Java's unsafe API) and without the need to instantiate a new ByteBuffer object.


liran-funaro:
Replaced by Oak's unsafe interface

PR #80, opened 2020-03-09
PERF: using direct field accessors to key/value reference data instead of instantiating new int[] array

liran-funaro:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #85, opened 2020-03-19
Introduce a new unsafe interface

liran-funaro:
And cleanup the safe interface from unsafe methods


core/src/main/java/com/oath/oak/OakUnsafeRef.java
```diff
@@ -0,0 +1,28 @@
+package com.oath.oak;
+
+import java.nio.ByteBuffer;
+import java.util.function.Function;
+
+public interface OakUnsafeRef {
```
ebortnikov: - This requires a chunk of external documentation. We must explain that there is an explicit tradeoff between safety and performance. 
   The user must understand that he is on his own with synchronization and bound checking. Preferably, give a couple of examples of what could go wrong. 
- Why is the name OakUnsafeRef? Shouldn't we be following the "Buffer" language everywhere?

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -198,12 +196,31 @@ public double getDouble(int index) {
...
-        valueOperator.unsafeBufferToIntArrayCopy(dup, srcPosition, dstArray, countInts);
-        end(s);
+    public ByteBuffer getByteBuffer() {
+        ByteBuffer buff = getValueSlice().getByteBuffer();
+        buff = buff.asReadOnlyBuffer();
+        int position = valuePosition();
+        int limit = position + capacity();
+        buff.position(position);
+        buff.limit(limit);
+        return buff.slice();
```
ebortnikov: Is this where the new object is created? 

In general, I'd suggest quantifying the decision to expose offsets - complexity vs performance. As long as the performance is acceptable (primarily queries) I'd recommend to simplify the programming model and not expose the complexity of working with offsets.

core/src/main/java/com/oath/oak/OakWBuffer.java
```diff
@@ -15,12 +15,6 @@
  */
 public interface OakWBuffer {
```
ebortnikov: Can we improve the lack of symmetry in naming: OakRKeyBufferImpl vs OakWBufferImpl. "W" should be for values, right? Or we'll have some other "R" for read-only values?

core/src/main/java/com/oath/oak/OakRReference.java
```diff
@@ -25,7 +26,7 @@
  * before each access since it can only be used without other concurrent writes in the background.
  * */
 
-public class OakRReference implements OakRBuffer {
+public class OakRReference implements OakRBuffer, OakUnsafeRef {
```
ebortnikov: Need to find a better name for this class, like OakReadOnlyBuffer maybe?

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -84,8 +86,27 @@ public double getDouble(int index) {
     }
 
     @Override
-    public void unsafeCopyBufferToIntArray(int srcPosition, int[] dstArray, int countInts) {
-        UnsafeUtils.unsafeCopyBufferToIntArray(getKeyBuffer().slice(), srcPosition, dstArray, countInts);
+    public ByteBuffer getByteBuffer() {
```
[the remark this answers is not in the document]
liran-funaro: DirectBuffer only supports a limited interface:
```java
public interface DirectBuffer {
    public long address();
    public Object attachment();
    public Cleaner cleaner();
}
```
The full implementation that extends ByteBuffer is in DirectByteBuffer class, but it is a private class so we cannot use it in our interface.

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -84,8 +86,27 @@ public double getDouble(int index) {
...
+        ByteBuffer buff = getKeyBuffer();
+        buff = buff.asReadOnlyBuffer();
+        buff.position(initialPosition);
+        return buff.slice();
     }
 
+    @Override
+    public int getOffset() {
```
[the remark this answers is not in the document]
liran-funaro: Yes

core/src/main/java/com/oath/oak/OakRReference.java
```diff
@@ -112,14 +113,29 @@ public double getDouble(int index) {
...
 
-    // the method provides an optimal way to copy data as array of integers from the key's buffer
-    // srcPosition - index from where to start copying the internal key buffer,
-    // srcPosition = 0 if to start from the beginning
     @Override
-    public void unsafeCopyBufferToIntArray(int srcPosition, int[] dstArray, int countInts) {
-        UnsafeUtils.unsafeCopyBufferToIntArray(getTemporaryPerThreadByteBuffer().slice(),
-                srcPosition + headerSize, dstArray, countInts);
+    public ByteBuffer getByteBuffer() {
+        ByteBuffer buff = getTemporaryPerThreadByteBuffer().asReadOnlyBuffer();
```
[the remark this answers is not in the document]
liran-funaro: I agree. I suggest to group many renaming like this and do a different PR for renaming ONLY.

core/src/main/java/com/oath/oak/OakRReference.java
```diff
@@ -112,14 +113,29 @@ public double getDouble(int index) {
...
+    }
+
+    @Override
+    public long getAddress() {
+        ByteBuffer buff = getTemporaryPerThreadByteBuffer();
+        long address = ((DirectBuffer) buff).address();
+        return address + headerSize + position;
     }
 
     private ByteBuffer getTemporaryPerThreadByteBuffer() {
```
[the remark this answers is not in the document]
liran-funaro: See a similar comment for OakRKeyValueImpl.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
liran-funaro: Regarding the second bullet, notice that this interface does not have any similarity to Java's buffers interface.
It only references to the buffer/address of the data. Hence the name.

core/src/main/java/com/oath/oak/OakWBuffer.java
liran-funaro: R: read only
W: wriyable
The read buffers have different implementation for key/value, but the write buffer have only one implementation for both.

core/src/main/java/com/oath/oak/OakWBufferImpl.java
```diff
@@ -159,4 +151,30 @@ public OakWBuffer putDouble(int index, double value) {
         bb.putDouble(dataPos + index, value);
         return this;
     }
+
+    @Override
```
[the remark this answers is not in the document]
liran-funaro: Yes

core/src/main/java/com/oath/oak/OakRReference.java
liran-funaro: I agree. I suggest to group many renaming like this and do a different PR for renaming ONLY.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * @return the key/value length
+     */
+    int getLength();
+
+    /**
+     * @return the exact address of the underlying buffer in the position of the key/value
+     */
+    long getAddress();
+
```
sanastas: It should have transform() method as well.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
sanastas: I am also supporting to remain with "Buffer" name! Yoav have changed OakRKeyBuffer for OakRReference and broke this rule which was a mistake. It should all be Buffers or all be References! But as currently the majority are Buffers this should be also just Unsafe Buffer. Afterwards (or already here) we can change all the naming in a consistent way. Because in "safe" variation the class is more similar to buffer, I vote for buffer-ed name. Why someone would cast Buffer to Reference? To cast Buffer to UnsafeBuffer is more reasonable....

liran-funaro: I agree that casting buffer to ref is nonintuitive.
But calling something a buffer when its API does not resemble a buffer in any way is confusing.
I prefer using the reference notation, but if we want to stick with Java's internal conventions, we could mimic the `DirectBuffer` naming rule and call this `OakUnsafeDirectBuffer`.
Agreed?

core/src/main/java/com/oath/oak/OakUnsafeRef.java
liran-funaro: Note that at this point, `OakBuffer`s `transform()` method acts on a `ByteBuffer`, so adding a `transform()` method here does not make much sense.
In addition, thinking forward, wouldn't it be easier to just call the `transform()` method of the `OakBuffer` and then cast the input buffer to `OakUnsafeRef` instead of having to deal with two different kinds of transform method types?

core/src/main/java/com/oath/oak/OakRReference.java
sanastas: I agree that this name is bad. I think it should be OakRKeyBufferImpl (or similar). I also think that naming of the Oak output buffers only can be held under this PR.

core/src/main/java/com/oath/oak/OakRReference.java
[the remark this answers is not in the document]
sanastas: This name indeed can be changed later

core/src/main/java/com/oath/oak/OakRReference.java
sanastas: Especially not to add the NEW Reference name...

core/src/main/java/com/oath/oak/OakUnsafeRef.java
sanastas: OakUnsafeDirectBuffer is better than OakUnsafeRef for sure!

core/src/main/java/com/oath/oak/OakUnsafeRef.java
sanastas: 1. I understand that currently you can manage without transform here, but if already adding this "OakUnsafeDirectBuffer" I believe it should be the full interface.
2. Transform with safe/unsafe buffer is much more comfortable. If you do the cast once you can work with your new UnsafeBuffer only. Otherwise you need to jungle between Safe/Unsafe buffers.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
sanastas: Other concern: you disregard the OakR and OakW! Although they are similar (put aside transform) one should return read-only buffer and the other should return writable buffer... What about that?

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -230,4 +219,33 @@ private void lookupValueReference() {
         valueReference = lookUp.valueReference;
         version = lookUp.version;
     }
+
+    /*-------------- OakUnsafeRef --------------*/
+
+    @Override
+    public ByteBuffer getByteBuffer() {
```
sanastas: This is actually the only place I see where we currently rely on good usage of the user. This ByteBuffer should not be accessed while it is written by another thread. Should be noted somewhere...

core/src/main/java/com/oath/oak/OakRBuffer.java
```diff
@@ -133,13 +133,4 @@
      * @return The return value of the transform
      */
     <T> T transform(Function<ByteBuffer, T> transformer);
-
```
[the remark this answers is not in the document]
liran-funaro: This is unsafe because the user should be aware that the endianness of this data will not be compatible with the buffer endianness as used by, for example, `getInt()`.
This is extremely problematic as the user cannot modify the endianness of the buffer without using the unsafe interface.
Regarding the lock, the user can still use the unsafe interface inside a locked context and implement its own array copy method.

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
liran-funaro: This is now noted in the `OakUnsafeDirectBuffer` interface internal documentation.

core/src/main/java/com/oath/oak/OakWBufferImpl.java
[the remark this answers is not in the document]
liran-funaro: This is now documented in `OakUnsafeDirectBuffer` documentation.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
liran-funaro: This is now noted inside the internal documentation of `OakUnsafeDirectBuffer` that depending on the context, `getByteBuffer()` might return a read-only buffer.

core/src/main/java/com/oath/oak/OakUnsafeRef.java
ebortnikov: In general, agree with both statements. 
1. It's better to provide one consistent way of doing things. 
2. However, the Oak{R,W}Buffer world must remain self-sufficient. If a transform() there is not just a syntactic sugar, we should support it too. That might be in conflict with (1).

liran-funaro: Adding a transform here is syntactic sugar.

Instead of writing:
```java
OakRBuffer b = ...;
b.transform(rb -> {
    OakUnsafeDirectBuffer urb = (OakUnsafeDirectBuffer) rb;
    long address = urb.getAddress();
   ...
});
```
You'll be able to write:
```java
OakRBuffer b = ...;
OakUnsafeDirectBuffer ub = (OakUnsafeDirectBuffer) b;
ub.transform(urb -> {
    long address = urb.getAddress();
   ...
});
```
* Notice that the casting was moved outside of the transform lambda expression.

I think this API should remain separate from the buffer API that is used for regular access.
That is, its only purpose should be getting the underlying data.
Any "Oak" related logic and interface should be used from the regular buffer interface.

core/src/test/java/com/oath/oak/UnsafeUtilsTest.java
```diff
@@ -53,8 +53,12 @@ public void testUnsafeCopy() {
             int size = entry.getKey().getInt(0);
             int[] dstArrayValue = new int[size];
             int[] dstArrayKey = new int[size];
-            entry.getKey().unsafeCopyBufferToIntArray(Integer.BYTES, dstArrayKey, size);
-            entry.getValue().unsafeCopyBufferToIntArray(Integer.BYTES, dstArrayValue, size);
+            OakUnsafeRef keyRef = ((OakUnsafeRef) entry.getKey());
+            OakUnsafeRef valueRef = ((OakUnsafeRef) entry.getValue());
```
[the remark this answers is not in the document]
liran-funaro: I removed the `unsafeCopyBufferToIntArray()` code from Oak entirly.
Also removed this test.

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -6,86 +6,149 @@
...
-    OakRKeyBufferImpl(long keyReference, MemoryManager memoryManager) {
-        this.keyReference = keyReference;
+    OakRKeyBufferImpl(MemoryManager memoryManager, int headerSize) {
         this.memoryManager = memoryManager;
-        this.initialPosition = UnsafeUtils.longToInts(keyReference)[POSITION_ARRAY_INDEX];
+        this.headerSize = headerSize;
     }
 
-    private ByteBuffer getKeyBuffer() {
-        int[] keyArray = UnsafeUtils.longToInts(keyReference);
```
Eshcar: I see, since we are creating an array of ints (object) anyway, no point in storing the data in a single long

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -6,86 +6,149 @@
...
+ * */
+
+public class OakRKeyBufferImpl implements OakRBuffer, OakUnsafeDirectBuffer {
+
+    private int blockID = OakNativeMemoryAllocator.INVALID_BLOCK_ID;
+    private int position = -1;
+    private int length = -1;
     private final MemoryManager memoryManager;
-    private final int initialPosition;
+    private final int headerSize;
```
Eshcar: capital letters for constant?

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -6,86 +6,149 @@
...
+
+    // The OakRReference user accesses OakRReference as it would be a ByteBuffer with initially zero position.
+    // We translate it to the relevant ByteBuffer position, by adding keyPosition and the header size to any given index
 
-    OakRKeyBufferImpl(long keyReference, MemoryManager memoryManager) {
-        this.keyReference = keyReference;
+    OakRKeyBufferImpl(MemoryManager memoryManager, int headerSize) {
```
Eshcar: do you have different sizes of headers? it is not fixed size?

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
```diff
@@ -6,86 +6,149 @@
...
     @Override
     public int capacity() {
-        return UnsafeUtils.longToInts(keyReference)[BLOCK_ID_LENGTH_ARRAY_INDEX] & KEY_LENGTH_MASK;
+        return getTemporaryPerThreadByteBuffer().capacity();
     }
 
     @Override
     public byte get(int index) {
-        return getKeyBuffer().get(index + initialPosition);
+        return getTemporaryPerThreadByteBuffer().get(index + headerSize + position);
```
Eshcar: method 
int getIndexInByteBuffer(int index) {return headerSize + index + position;}

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
```diff
@@ -197,15 +195,6 @@ public double getDouble(int index) {
         throw new RuntimeException("Transform failed: reached retry limit (1024).");
     }
 
-    @Override
-    public void unsafeCopyBufferToIntArray(int srcPosition, int[] dstArray, int countInts) {
-        Slice s = getValueSlice();
-        start(s);
```
Eshcar: So Liran suggests that start and end will be called by the user when necessary, and it is not supported internally

core/src/main/java/com/oath/oak/OakUnsafeDirectBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+ * 2. Data boundaries: when using this interface, Oak will not alert the developer regarding any out of boundary access.
+ *    Thus, the developer should use getOffset() and getLength() to obtain the data boundaries and carefully access
+ *    the data. Writing data out of these boundaries might result in corrupted data, a crash or a deadlock.
+ *
+ * To use this interface, the developer should cast Oak's buffer (OakRBuffer/OakWBuffer) to this interface,
+ * similarly to how Java's internal DirectBuffer is used. For example:
+ * void foo(OakRBuffer b) {
+ *     long address = ((OakUnsafeDirectBuffer) b).getAddress();
+ *     ...
+ * }
```
Eshcar: Nice documentation.
When all PRs are resolved this should also go into the wiki, with code examples of using the interface, e.g., the way you are using it in Druid

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
Eshcar: Should Liran become a Member of the Oak community ? I think he should :)

core/src/test/java/com/oath/oak/UnsafeUtilsTest.java
```diff
@@ -12,134 +12,6 @@
 
 public class UnsafeUtilsTest {
 
-
-    @Test
-    public void testUnsafeCopy() {
```
Eshcar: strange that we have such high test coverage for a piece of code that is actually redundant ?

core/src/main/java/com/oath/oak/OakRKeyBufferImpl.java
liran-funaro: This is solved by PR https://github.com/yahoo/Oak/pull/80

core/src/main/java/com/oath/oak/OakRValueBufferImpl.java
liran-funaro: No, to the contrary. This PR does not change this behavior.
Specifically here, I removed `unsafeCopyBufferToIntArray()` entirely. The other methods (e.g., `getInt()`) still uses start/end.

core/src/test/java/com/oath/oak/UnsafeUtilsTest.java
liran-funaro: It was used in Druid, but now we realized that such unsafe behavior should be implemented by the user if needed and not be given as a service.

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -247,7 +247,7 @@ ByteBuffer readKey(int entryIndex) {
      * The thread-local ByteBuffer can be reused by different threads, however as long as
      * a thread is invoked the ByteBuffer is related solely to this thread.
      */
-    void setKeyRefer(int entryIndex, OakRReference keyReferBuffer) {
+    void setKeyRefer(int entryIndex, OakRBufferStream keyReferBuffer) {
```
sanastas: I am sorry I am against the "OakRBufferStream" name... Why stream? It is used also in EntryIterator and in KeyIterator (not only Key/EntryStreamIterator). This is just the buffer for read access for keys, so any combination of Oak-R-Key-Buffer (with or without Impl) should work

core/src/main/java/com/oath/oak/OakUnsafeDirectBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.function.Function;
+
+/**
+ * This interface allows high performance access to the underlying data of Oak.
+ * To achieve that, it sacrifices safety, so it should be used only if you know what you are doing.
+ * Misuse of this interface might result in corrupted data, a crash or a deadlock.
+ *
+ * Specifically, the developer should be concerned by two issues:
+ * 1. Concurrency: using this interface inside the context of serialize, compute, compare and transform is thread safe.
+ *    In other contexts (e.g., zero copy API), the developer should ensure that there is no concurrent access to this
```
sanastas: I wouldn't mention here "zero copy API", but you can speak about get/iterator output.

core/src/main/java/com/oath/oak/OakUnsafeDirectBuffer.java
sanastas: Please add also about read-only and writable buffers when cast OakRBuffer or OakWBuffer respectively

sanastas: I see that you speak about it bellow, but I think it should be correlated with cast...

core/src/main/java/com/oath/oak/OakWBufferImpl.java
```diff
@@ -6,12 +6,14 @@
...
 
+import sun.nio.ch.DirectBuffer;
+
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 
 // An instance of OakWBufferImpl is only used when the write lock of the value referenced by it is already acquired.
 // This is the reason no lock is acquired in each access.
-public class OakWBufferImpl implements OakWBuffer {
+public class OakWBufferImpl implements OakWBuffer, OakUnsafeRef {
```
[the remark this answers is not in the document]
liran-funaro: This is a discussion for future PR.

core/src/main/java/com/oath/oak/OakUnsafeDirectBuffer.java
liran-funaro: fixed

core/src/main/java/com/oath/oak/Chunk.java
liran-funaro: fixed

core/src/main/java/com/oath/oak/OakUnsafeDirectBuffer.java
liran-funaro: fixed

PR #86, opened 2020-03-19
Bug fix: OakRValueBufferImpl will fail if the value moved to a different ByteBuffer

liran-funaro:
Please also see the added unit-test that reproduces this bug.

Note that the latest PR (https://github.com/yahoo/Oak/pull/84) only escalates this problem.
In the old code, it will only fail if Oak replaced the value's ByteBuffer.
In the latest PR, it will fail for any moved value.


sanastas:
Can you please rebase it to the newly update master? If yes, I can merge and incorporate it with my PR. So we have this fix inside as well...

sanastas:
I looked over the code and the fix looks good for me

liran-funaro:
Greate. It is now rebased

PR #87, opened 2020-03-22
Introduce a new safer interface

liran-funaro:
The new interface does not expose the internal ByteBuffer to the user outside of the `OakUnsafeRef` interface.
This PR might be better compared to PR https://github.com/yahoo/Oak/pull/85 to avoid duplicated comments.


sanastas:
I wonder why do you continue to add changes to this PR...? Just saw that you added the README here. Isn't it something that is not going to be committed?

PR #88, opened 2020-03-29
New oak buffer API documentation

liran-funaro:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,17 @@
 # Oak
-> Oak (Off-heap Allocated Keys) is a scalable, concurrent, in-memory Key Value (KV) map.
+> Oak (Off-heap Allocated Keys) is a scalable, concurrent, in-memory Key-Value (KV) map.
 
-OakMap is a concurrent Key-Value Map that may keep all keys and values off-heap. This enables working with bigger heap sizes than JVM's managed heap.
-OakMap implements the industry standard Java8 ConcurrentNavigableMap API. It provides strong (atomic) semantics for read, write, and read-modify-write, as well as (non-atomic) range query (scan) operations, both forward and backward.
+OakMap is a concurrent Key-Value Map that keeps all keys and values off-heap. This allows storing more data (up to 3 times more data) compare to using the standard JVM heap management, albeit using the same memory footprint.
+OakMap implements the industry-standard Java8 ConcurrentNavigableMap API. It provides strong (atomic) semantics for read, write, and read-modify-write operations, as well as (non-atomic) range query (scan) operations, both forward and backward.
 OakMap is optimized for big keys and values, in particular, for incremental maintenance of objects (update in-place).
 It is faster and scales better with additional CPU cores than the popular Java ConcurrentNavigableMap [ConcurrentSkipListMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentSkipListMap.html).
 
 ## Why OakMap?
-1. OakMap provides great performance: it employs fine-grain synchronization, and thus scales well with numbers of threads; it also achieves cache-friendliness by avoiding memory fragmentation. [Performance evaluation](https://github.com/yahoo/Oak/wiki/Performance).
+1. OakMap provides great performance: it employs fine-grain synchronization, and thus scales well with numbers of threads; it also achieves cache-friendliness by avoiding memory fragmentation (see [performance evaluation](https://github.com/yahoo/Oak/wiki/Performance)).
 2. OakMap takes keys and the data off-heap, and thus allows working with a huge heap (RAM) -- even more than 50G -- without JVM GC overheads.
-	- To support off-heap, OakMap has embedded, efficient, epoch-based memory management that mostly eliminates JVM GC overheads.
-4. OakMap provides a rich API for **atomic** accesses to data. For example, OakMap supports atomic compute() -- in place computations on existing keys -- whereas the current Java ConcurrentSkipListMap implementation does not guarantee the atomicity of compute(). OakMap’s update operations (such as put and compute) take user-provided lambda functions for easy integration in diverse use cases.
-5. Descending Scans: OakMap expedites descending scans without additional complexity. In our experiments, OakMap’s descending scans are 4.8x faster than ConcurrentSkipListMap’s, and perform similarly to their ascending counterparts. [Performance evaluation](https://github.com/yahoo/Oak/wiki/Performance).
+   - To support off-heap, OakMap has embedded, efficient, epoch-based memory management that mostly eliminates JVM GC overheads.
+4. OakMap provides a rich API for **atomic** accesses to data. For example, OakMap supports atomic compute() -- in place computations on existing keys -- whereas the current Java ConcurrentSkipListMap implementation does not guarantee the atomicity of `compute()`. OakMap’s update operations (such as put and compute) take user-provided lambda functions for easy integration in diverse use cases.
+5. Descending Scans: OakMap expedites descending scans without additional complexity. In our experiments, OakMap’s descending scans are 4.8x faster than ConcurrentSkipListMap’s, and perform similarly to their ascending counterparts (see [performance evaluation](https://github.com/yahoo/Oak/wiki/Performance)).
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,31 +111,49 @@ After a Key-Value pair is inserted into OakMap, it is kept in a serialized (buff
...
 
-This is how to create the comparator class in your code:
+*Note 1*: Oak use dedicated objects to access off-heap memory: `OakReadBuffer` and `OakWriteBuffer`.
+See [Oak Buffers](#oak-buffers) for more information.
+
+*Note 2*: `OakReadBuffer` and `OakWriteBuffer` should not be stored for future use.
+They are valid only in the context of these methods (`compareKeys()`/`compareSerializedKeys()/compareSerializedKeyAndKey()`).
+Using these buffers outside their intended context may yield unpredicted results.
+
+For example, the implementation of key serializer for an application that use integer as keys might like this:
```
sanastas: ... key *comparator* for an ...

README.md
```diff
@@ -92,31 +111,49 @@ After a Key-Value pair is inserted into OakMap, it is kept in a serialized (buff
...
+  }
+
+  int compareSerializedKeyAndKey(OakReadBuffer serializedKey1, Integer key2) {
+    return serializedKey1.getInt(0) - key2;
+  }
+}
 ```
 
 ### Builder
 We provide an example how to create OakMapBuilder and OakMap. For a more comprehensive code example please refer to the [Usage](#usage) section.
```
sanastas: I see that you updated Usage section as well which is great. How can I comment on that content? Is it going to be just copied to main repository as it is not part of PR, I think...?

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
 You are welcome to take a look on the OakMap's [full API](https://github.com/yahoo/Oak/wiki/Full-API).
```
sanastas: It almost a must to update *full API* section as well... This will exactly show the API...

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
...
-An OakRBuffer can represent either a key or a value. The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Note that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
-
-For backward compatibility with applications that are already based on the use of ByteBuffers, Oak Buffers provide the transform method that atomically applies a transformation to a *read-only* instance of the underlying ByteBuffer. For a more comprehensive code example please refer to the [usage](#usage) section. 
+### Oak Buffers
+Oak use dedicated buffer objects to access off-heap memory. 
+These buffers cannot be instantiated by the user and are always supplied to the user by Oak.
+Their interfaces are:
+ 1. `OakReadBuffer`: scoped, read-only access to the data
+ 2. `OakWriteBuffer`: scoped, read and write access to the data
+ 3. `OakScopeFreeBuffer`: scope-free, read-only access to the data
```
sanastas: I see that you decided to use 3 types of buffers instead of old two. Which is may be good, instead of using "same named" OakReadBuffer in different circumstances... However, if so, I would suggest similar names just with some negation as part of the name. For example: internal-external, restricted-free, bounded-unbounded, closed-open, ...

How about OakReadRestrictBuffer and OakReadFreeBuffer?

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
...
-
-An OakRBuffer can represent either a key or a value. The OakRBuffer's user can use the standard interface of a *read-only* ByteBuffer, for example, `int getInt(int index)`, `char getChar(int index)`, `limit()`, etc. Note that ConcurrentModificationException can be thrown as a result of any OakRBuffer method in case the mapping is concurrently deleted.
-
-For backward compatibility with applications that are already based on the use of ByteBuffers, Oak Buffers provide the transform method that atomically applies a transformation to a *read-only* instance of the underlying ByteBuffer. For a more comprehensive code example please refer to the [usage](#usage) section. 
+### Oak Buffers
+Oak use dedicated buffer objects to access off-heap memory. 
+These buffers cannot be instantiated by the user and are always supplied to the user by Oak.
```
sanastas: What about minKey? Didn't you wanted it to be OakReadBuffer? Will we serialize from K ourselves?

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
...
+They mimic the standard interface of Java's `ByteBuffer`, for example, `int getInt(int index)`, `char getChar(int index)`, `capacity()`, etc. 
+
+Scoped buffers (`OakReadBuffer` and `OakWriteBuffer`) can only be used in the scope of the callback method they were first introduced to the user. 
+Such callback method might be the application's serializer and comparator, or a lambda function that can read/store/update the data (see [full API](https://github.com/yahoo/Oak/wiki/Full-API) for more information).
+This access reduces unnecessary copies and deserialization of the underlying data.
+In their intended context, the user does not need to worry about concurrent accesses and memory management.
+Using these buffers outside their intended context may yield unpredicted results, e.g., reading non-consistent data and/or irrelevant data.
+
+The third buffer (`OakScopeFreeBuffer`) is scope free, i.e., its usage is not scope limited and may be stored for future use.
+Some OakMap's methods returns this buffer to avoid copying the data and instead the user can access to the underlying memory buffer directly (lazy evaluation).
```
sanastas: All zero-copy interface OakMap's methods returns this buffer...

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
...
+While the scoped buffers are mutually exclusive, when using `OakScopeFreeBuffer`, the same memory might be access by a concurrent update operations.
```
sanastas: "mutually exclusive" is not clear enough, explain that this is about synchronization of the data access/update. In addition multiple read accesses can happen simultaneously, so not exactly mutually exclusive. May be just say: "data accesses are synchronized..."

README.md
```diff
@@ -128,15 +165,36 @@ OakMap<K,V> oak = builder.build();
...
+Thus, the reader may encounter different values -- and even value deletions -- when accessing `OakScopeFreeBuffer` multiple times.
+Specifically, when trying to access a deleted mapping via a `OakScopeFreeBuffer`, `ConcurrentModificationException` will be thrown.
+This is of course normal behavior for a concurrent map that avoids copying.
+To allow complex, multi-value atomic operations on the data, `OakScopeFreeBuffer` provides a `transform()` method that allow the user to apply a transformation function atomically on a read-only, scoped version of the buffer (`OakReadBuffer`). 
+See the [zero-copy API](#zero-copy-api) for more information.
+
+For backward compatibility with applications that are already based on the use of `ByteBuffer`, Oak's buffers also implements a dedicated unsafe interface `OakUnsafeDirectBuffer`.
```
sanastas: More elaborate explanations on OakUnsafeDirectBuffer are required here or in additional section. The *full API* doesn't explain it, *full API* are just methods. Even if it would something need to be here as well.

README.md
```diff
@@ -274,7 +331,7 @@ try (OakMap sub = oak.subMap(from, false, to, true)) {
 ##### Direct ByteBuffer Transformations
 
 ```java
-Function<ByteBuffer, String> toStrings = e -> String.valueOf(e.getInt(0));
+Function<OakReadBuffer, String> toStrings = e -> String.valueOf(e.getInt(0));
 
 Iterator<String> iter = oak.zc().values().stream().map(v -> v.transform(toStrings)).iterator();
 while (iter.hasNext()) {
```
sanastas: 1) There need to be OakUnsafeDirectBuffer usage example.
2) Stream iterators need to be added... although not part of this effort. If you prefer not to do it here please at least open an issue.

README.md
```diff
@@ -58,29 +58,48 @@ OakMap requires multiple parameters to be defined for the builder, as shall be e
...
+  int calculateSize(T data);
 }
 ```
 
-This is how to create those classes in your code:
+*Note 1*: Oak use dedicated objects to access off-heap memory: `OakReadBuffer` and `OakWriteBuffer`.
+See [Oak Buffers](#oak-buffers) for more information.
+
+*Note 2*: `OakReadBuffer` and `OakWriteBuffer` should not be stored for future use.
+They are valid only in the context of these methods (`serialize()`/`deserialize()`).
```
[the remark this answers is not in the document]
liran-funaro: This is OakScopeFreeBuffer. See "Oak Buffers" section.

README.md
liran-funaro: In the current implementation, the user supplies a min key as K.
We use the user's serialize function to convert it to a buffer.
Nothing changes there. Just that serialize will not be called with OakWriteBuffer.
Then we disable the write access to the buffer to prevent the user from changing it.

README.md
liran-funaro: The way I see it, we escalate the user required proficiency, where higher level of proficiency using more descriptive names, and lower levels use minimalistic naming.
For example: proficiency levels
0. Use only scoped buffers: `Oak{R/W}Buffer`
1. Use scope free buffers: `OakScopeFreeBuffer`
2. Use underlying data: `OakUnsafeDirectBuffer`

According to the logic you suggested in your comment, regular buffers should be called `OakSafeBoundedReadBuffer`.
The internal class names could (and should) be more descriptive as the user is not exposed to them.

Similar naming convention in Java:
ByteBuffer -> DirectBuffer (DirectByteBuffer and HeapByteBuffer are internal)

Note that we might have chosen here one of two approaches:
1. Reduced restrictions added to the name
2. Increased restrictions added to the name
But we don't have unscoped OakWBuffer so (2) wouldn't make sense for us.

README.md
```diff
@@ -58,29 +58,48 @@ OakMap requires multiple parameters to be defined for the builder, as shall be e
...
+Using these buffers outside their intended context may yield unpredicted results.
+
+For example, the implementation of key serializer for an application that use integer as keys might like this:
```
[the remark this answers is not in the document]
liran-funaro: fixed

README.md
```diff
@@ -58,29 +58,48 @@ OakMap requires multiple parameters to be defined for the builder, as shall be e
...
 
 ```java
-public class OakKeySerializerImplementation implements OakSerializer<K>
-{...}
+public class MyAppKeySerializer implements OakSerializer<Integer> {
+  void serialize(Integer key, OakWriteBuffer serializedKey) {
+    // We store the value at the first position of the off-heap buffer.
+    storage.putInt(0, value);
```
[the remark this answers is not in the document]
liran-funaro: fixed

README.md
liran-funaro: fixed

README.md
liran-funaro: fixed

README.md
liran-funaro: fixed

README.md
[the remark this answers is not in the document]
sanastas: Yes, I didn't understand it at the beginning

README.md
sanastas: I thought you updated the wiki

README.md
sanastas: Important not to forget when committing. Maybe to add Performance Evaluation at the same time.

README.md
sanastas: User with lowest level of proficiency may not use zero copy at all. If zero copy is used then Oak{R/W}Buffer AND OakScopeFreeBuffer need to be used. So I do not see here higher level of proficiency... Both OakScopeFreeBuffer and Oak{R/W}Buffer are the same level...

README.md
liran-funaro: fixed

README.md
```diff
@@ -1,17 +1,17 @@
...
 	- To support off-heap, OakMap has embedded, efficient, epoch-based memory management that mostly eliminates JVM GC overheads.
 4. OakMap provides a rich API for **atomic** accesses to data. For example, OakMap supports atomic compute() -- in place computations on existing keys -- whereas the current Java ConcurrentSkipListMap implementation does not guarantee the atomicity of compute(). OakMap’s update operations (such as put and compute) take user-provided lambda functions for easy integration in diverse use cases.
-5. Descending Scans: OakMap expedites descending scans without additional complexity. In our experiments, OakMap’s descending scans are 4.8x faster than ConcurrentSkipListMap’s, and perform similarly to their ascending counterparts. [Performance evaluation](https://github.com/yahoo/Oak/wiki/Performance).
+5. Descending Scans: OakMap expedites descending scans without additional complexity. In our experiments, OakMap’s descending scans are 4.8x faster than ConcurrentSkipListMap’s, and perform similarly to their ascending counterparts (see [performance evaluation](https://github.com/yahoo/Oak/wiki/Performance)).
```
sanastas, line 14: In recent experiments "OakMap’s descending scans are 8x faster than ConcurrentSkipListMap’s" we just need to update the Oak wiki performance. Will happen after release, I believe...

sanastas: But please update to 8

PR #89, opened 2020-03-29
Add common package for basic java primitives Oak specialization

liran-funaro:
This is a preliminary step to reduce the footprint of future API changes.


core/src/main/java/com/oath/oak/common/OakCommonFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import com.oath.oak.common.intbuffer.OakIntBufferSerializer;
+import com.oath.oak.common.integer.OakIntComparator;
+import com.oath.oak.common.integer.OakIntSerializer;
+import com.oath.oak.common.string.OakStringComparator;
+import com.oath.oak.common.string.OakStringSerializer;
+
+import java.nio.ByteBuffer;
+import java.util.Comparator;
+
+public class OakCommonFactory {
```
sanastas: 1) This is something to be used by external user as well, right? Should he get the "normal" builder from here as well? Though this is only for "specific" builders...?
2) OakCommonFactory --> at least OakMapCommonFactory, but still it is confusing because it sounds like something creating "Oaks" and not OakMapBuilders... Bbuilder is also kind of factory. May be OakMapBuilderFactory? Then coming to my first question from where to take the "normal/unspecific" OakMapBuilder...?
3) This change should be better reviewed by Eshcar as well. She is better in software engineering staff... But generally looks good to me.

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBuffer.java
```diff
@@ -1,18 +1,118 @@
 package com.oath.oak.synchrobench.maps;
 
 
+import com.oath.oak.OakComparator;
+import com.oath.oak.OakSerializer;
+import com.oath.oak.common.intbuffer.OakIntBufferComparator;
+import com.oath.oak.common.intbuffer.OakIntBufferSerializer;
+
 import java.nio.ByteBuffer;
```
Eshcar, line 9: all these changes and new methods in this class are scoped only to synchrobench testing framework

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/MyBuffer.java
```diff
@@ -1,18 +1,118 @@
...
+    public static int compareBuffers(MyBuffer buffer1, MyBuffer buffer2) {
+        return compareBuffers(buffer1.buffer, buffer2.buffer);
+    }
+
+    public static int compareBuffers(MyBuffer key1, ByteBuffer key2) {
+        return compareBuffers(key1.buffer, key1.buffer.position(), key1.buffer.capacity(),
+            key2, key2.position(), key2.capacity());
+    }
+
+
```
Eshcar, line 70: essentially imported MyBufferOak to this class

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/maps/OffHeapList.java
```diff
@@ -35,20 +35,18 @@ public OffHeapList() {
...
-                            ((Slice) key2).getByteBuffer());
+                    return MyBuffer.compareBuffers((MyBuffer) o1, ((Slice) key2).getByteBuffer());
                 }
 
             } else if (o2 instanceof MyBuffer) {
                 Cell cell1 = (Cell) o1;
                 Object key1 = cell1.key.get();
                 if (key1 instanceof MyBuffer) {
-                    return MyBufferOak.keysComparator.compareKeys((MyBuffer) key1, (MyBuffer) o2);
+                    return MyBuffer.defaultComparator.compareKeys((MyBuffer) key1, (MyBuffer) o2);
```
Eshcar, line 47: why call compare keys and not compare Buffers here?

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -19,97 +19,28 @@
...
 public class Chunk<K, V> {
+    static final int NONE_NEXT = 0;    // an entry with NONE_NEXT as its next pointer, points to a null entry
 
     /*-------------- Constants --------------*/
 
-    /***
-     * This enum is used to access the different fields in each entry.
-     * The value associated with each entry signifies the offset of the field relative to the entry's beginning.
-     */
-    enum OFFSET {
```
Eshcar, line 29: where is this OFFSET going? It also changes location in Anastasia's PR

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -119,19 +50,15 @@
     // chunk can be in the following states: normal, frozen or infant(has a creator)
     private final AtomicReference<State> state;
     private AtomicReference<Rebalancer<K, V>> rebalancer;
-    private final int[] entries;    // array is initialized to 0, i.e., NONE - this is important!
+    private final EntrySet<K,V> entrySet;
```
Eshcar, line 53: wasn't this part of previous PR?

sanastas:
This is regarding your main comment. I hope you are reviewing **only** "liran-funaro:pr-common-primitives" commit. You shouldn't see any of  "Anastasia's PR". The reason is that this PR is based on Anastasia's branch (not yet committed) for easier future merge.

core/src/main/java/com/oath/oak/Chunk.java
sanastas: yes it was

core/src/main/java/com/oath/oak/Chunk.java
sanastas: I am referring only to comments caused by my PR. Yes it is part of  "Anastasia's PR". It was moved to EntrySet.

core/src/main/java/com/oath/oak/common/integer/OakIntSerializer.java
```diff
@@ -0,0 +1,33 @@
+package com.oath.oak.common.integer;
+
+import com.oath.oak.OakSerializer;
+
+import java.nio.ByteBuffer;
+
+public class OakIntSerializer implements OakSerializer<Integer> {
```
Eshcar, line 7: Are these 4 classes of  Int/IntBuffer comparator/serializer needed only for tests or also used in the Druid code?

core/src/main/java/com/oath/oak/common/OakCommonFactory.java
Eshcar: From what I understand this is a combination of 2 design patterns Builder pattern and Factory pattern - Liran correct me if I'm wrong:
The Builder pattern already existed before this PR (I believe it was introduced either by EranM or Anastasia); it allows creating an OakMap object and adding granularity to the object by applying chains of method calls to builder. This way the c'tor of the builder only gets mandatory attributes and the rest are set by the chain. Usually the Builder itself is a static class, and its methods return a builder object this allows the chain of method calls. Finally to get the OakMap object you call the build method, for example
`OakMap<Integer, Integer> oak = builder.build();` 
To be able to create different types of OakMap objects (<int, int>, <string,string>, etc.) you need different builder objects.
This is where the Factory pattern comes in (you may  want to change the name to OakMapBuilderFactory if this is indeed the intent). 
The factory is responsible for generating default builders for a predefined set of OakMap types.  For creating these builders the ctor gets the 4 mandatory parameters that are required for building an OakMap object. 
The 4 parameters include comparator and serializer. These are captured in the new classes like OakIntComparator that are also added in this PR.
I assume the main idea is to have it all in one place and not spreader around in the test classes. And perhaps this is also beneficial for the application.

Did I get this all correct?

core/src/main/java/com/oath/oak/common/integer/OakIntSerializer.java
sanastas: My two cents: Those classes are not used in Druid, but can help a new user to start working with Oak. For example, Ohad Serfati needed just <String,String> and erroneously made a buggy serializer/deserializer, until he realized we have those classes in our test package, so he copy-pasted. Ohad opened an issue asking to take such classes out of test package. If I understand it correctly this is done here (in addition to other changes).

core/src/main/java/com/oath/oak/common/OakCommonFactory.java
liran-funaro: I assume Oak library might include non-map data-structures in the future. So maybe: `OakCommonBuildersFactory`?

Anastasia:
1.a. Used by external users.
1.b. No. It gives the user common builders, but not replace the main builder.

Eshcar: You got everything right.

PR #92, opened 2020-03-30
BUGFIX: buffer limit set incorrectly in OakRKeyBuffer.getByteBuffer()

liran-funaro:
* In addition, add teadDown() function to OakNativeMemoryAllocatorTest to revert the BlocksPool size.

### DIFF
--- a/core/src/main/java/com/oath/oak/OakRKeyBuffer.java
+++ b/core/src/main/java/com/oath/oak/OakRKeyBuffer.java
@@ -130,8 +130,8 @@ public class OakRKeyBuffer implements OakRBuffer, OakUnsafeDirectBuffer {
     @Override
     public ByteBuffer getByteBuffer() {
         ByteBuffer buff = getTemporaryPerThreadByteBuffer().asReadOnlyBuffer();
-        buff.position(headerSize + position);
-        buff.limit(headerSize + position + length);
+        buff.position(position + headerSize);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/oath/oak/OakRKeyBuffer.java
```diff
@@ -130,8 +130,8 @@ private ByteBuffer getTemporaryPerThreadByteBuffer() {
     @Override
     public ByteBuffer getByteBuffer() {
         ByteBuffer buff = getTemporaryPerThreadByteBuffer().asReadOnlyBuffer();
-        buff.position(headerSize + position);
-        buff.limit(headerSize + position + length);
+        buff.position(position + headerSize);
+        buff.limit(position + length);
```
sanastas: 1) I wonder why do we need to set the limit here if Block's getBufferForThread() method already does this? To set position not to include the headerSize is enough.
2) asReadOnlyBuffer() already creates a Buffer object as does the slice() in the last line. Can we avoid 2 objects instead of 1?
3) Also please add a note that length includes the header size (this is true at least for value)

liran-funaro: 1. We have to set the position, but the limit is redundant. I replaced it with a comment.
2. Unfortunately, no. ByteBuffer implementation does not support that.
3. Done.

PR #102, opened 2020-05-13
Introduce a new API which does not expose the internal ByteBuffers by default

liran-funaro:

### DIFF
--- a/benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/MyBuffer.java
+++ b/benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/MyBuffer.java
@@ -2,7 +2,10 @@ package com.oath.oak.synchrobench;
 
 
 import com.oath.oak.OakComparator;
+import com.oath.oak.OakReadBuffer;
 import com.oath.oak.OakSerializer;
+import com.oath.oak.OakUnsafeDirectBuffer;
+import com.oath.oak.OakWriteBuffer;
 import com.oath.oak.common.intbuffer.OakIntBufferComparator;
 import com.oath.oak.common.intbuffer.OakIntBufferSerializer;
 
@@ -29,9 +32,8 @@ public class MyBuffer implements Comparable<MyBuffer> {
         return compareBuffers(this, o);
     }
 
-    public static void serialize(MyBuffer inputBuffer, ByteBuffer targetBuffer) {
-        int targetPos = targetBuffer.position();
-        serialize(inputBuffer, targetBuffer, targetPos);
+    public static void serialize(MyBuffer inputBuffer, OakUnsafeDirectBuffer targetBuffer) {
+        serialize(inputBuffer, targetBuffer.getByteBuffer(), targetBuffer.getOffset());
     }
 
     public static void serialize(MyBuffer inputBuffer, ByteBuffer targetBuffer, int targetPos) {
@@ -41,12 +43,11 @@ public class MyBuffer implements Comparable<MyBuffer> {
         targetPos += Integer.BYTES;
 
         OakIntBufferSerializer.copyBuffer(inputBuffer.buffer, DATA_POS, inputBuffer.capacity / Integer.BYTES,
-            targetBuffer, targetPos);
+                targetBuffer, targetPos);
     }
 
-    public static MyBuffer deserialize(ByteBuffer inputBuffer) {
-        int inputPos = inputBuffer.position();
-        return deserialize(inputBuffer, inputPos);
+    public static MyBuffer deserialize(OakUnsafeDirectBuffer inputBuffer) {
+        return deserialize(inputBuffer.getByteBuffer(), inputBuffer.getOffset());
     }
 
     public static MyBuffer deserialize(ByteBuffer inputBuffer, int inputPos) {
@@ -62,34 +63,39 @@ public class MyBuffer implements Comparable<MyBuffer> {
 
     public static int compareBuffers(ByteBuffer buff1, int pos1, int cap1, ByteBuffer buff2, int pos2, int cap2) {
         return OakIntBufferComparator.compare(buff1, pos1, cap1 / Integer.BYTES,
-            buff2, pos2, cap2 / Integer.BYTES);
+                buff2, pos2, cap2 / Integer.BYTES);
     }
 
-    public static int compareBuffers(ByteBuffer buffer1, ByteBuffer buffer2) {
-        int pos1 = buffer1.position();
-        int pos2 = buffer2.position();
+    public static int compareBuffers(OakUnsafeDirectBuffer buffer1, OakUnsafeDirectBuffer buffer2) {
+        ByteBuffer buf1 = buffer1.getByteBuffer();
+        ByteBuffer buf2 = buffer2.getByteBuffer();
+
+        int pos1 = buffer1.getOffset();
+        int pos2 = buffer2.getOffset();
 
         // In the serialized buffer, the first integer signifies the size.
-        int cap1 = buffer1.getInt(pos1);
-        int cap2 = buffer2.getInt(pos2);
+        int cap1 = buf1.getInt(pos1);
+        int cap2 = buf2.getInt(pos2);
 
         // Thus, the data position starts after the first integer.
         pos1 += Integer.BYTES;
         pos2 += Integer.BYTES;
 
-        return compareBuffers(buffer1, pos1, cap1, buffer2, pos2, cap2);
+        return compareBuffers(buf1, pos1, cap1, buf2, pos2, cap2);
     }
 
-    public static int compareBuffers(MyBuffer key1, ByteBuffer buffer2) {
-        int pos2 = buffer2.position();
+    public static int compareBuffers(MyBuffer key1, OakUnsafeDirectBuffer buffer2) {
+        ByteBuffer buf2 = buffer2.getByteBuffer();
+
+        int pos2 = buffer2.getOffset();
 
         // In the serialized buffer, the first integer signifies the size.
-        int cap2 = buffer2.getInt(pos2);
+        int cap2 = buf2.getInt(pos2);
 
         // Thus, the data position starts after the first integer.
         pos2 += Integer.BYTES;
 
-        return compareBuffers(key1.buffer, DATA_POS, key1.capacity, buffer2, pos2, cap2);
+        return compareBuffers(key1.buffer, DATA_POS, key1.capacity, buf2, pos2, cap2);
     }
 
     public static int compareBuffers(MyBuffer key1, MyBuffer key2) {
@@ -99,13 +105,13 @@ public class MyBuffer implements Comparable<MyBuffer> {
     public static OakSerializer<MyBuffer> defaultSerializer = new OakSerializer<MyBuffer>() {
 
         @Override
-        public void serialize(MyBuffer key, ByteBuffer targetBuffer) {
-            MyBuffer.serialize(key, targetBuffer);
+        public void serialize(MyBuffer key, OakWriteBuffer targetBuffer) {
+            MyBuffer.serialize(key, (OakUnsafeDirectBuffer) targetBuffer);
         }
 
         @Override
-        public MyBuffer deserialize(ByteBuffer serializedKey) {
-            return MyBuffer.deserialize(serializedKey);
+        public MyBuffer deserialize(OakReadBuffer serializedKey) {
+            return MyBuffer.deserialize((OakUnsafeDirectBuffer) serializedKey);
         }
 
         @Override
@@ -121,13 +127,13 @@ public class MyBuffer implements Comparable<MyBuffer> {
         }
 
         @Override
-        public int compareSerializedKeys(ByteBuffer serializedKey1, ByteBuffer serializedKey2) {
-            return compareBuffers(serializedKey1, serializedKey2);
+        public int compareSerializedKeys(OakReadBuffer serializedKey1, OakReadBuffer serializedKey2) {
+            return compareBuffers((OakUnsafeDirectBuffer) serializedKey1, (OakUnsafeDirectBuffer) serializedKey2);
         }
 
         @Override
-        public int compareKeyAndSerializedKey(MyBuffer key, ByteBuffer serializedKey) {
-            return compareBuffers(key, serializedKey);
+        public int compareKeyAndSerializedKey(MyBuffer key, OakReadBuffer serializedKey) {
+            return compareBuffers(key, (OakUnsafeDirectBuffer) serializedKey);
         }
     };
 }
--- a/core/src/main/java/com/oath/oak/Block.java
+++ b/core/src/main/java/com/oath/oak/Block.java
@@ -16,28 +16,28 @@ import static com.oath.oak.OakNativeMemoryAllocator.INVALID_BLOCK_ID;
 
 class Block {
 
-    private final ByteBuffer buffer;
+    /*
+    For performance reasons, we keep two ByteBuffer objects that refers to the same buffer.
+    writeBuffer is the original allocated buffer, and readBuffer is a duplicated version of writeBuffer
+    but with read-only access permissions.
+    This allow the user to supply the a read-only ByteBuffer (when needed) without instantiating a new object.
+     */
+    private final ByteBuffer readBuffer;
+    private final ByteBuffer writeBuffer;
+
     private final int capacity;
     private final AtomicInteger allocated = new AtomicInteger(0);
     private int id; // placeholder might need to be set in the future
-
-    // in order to avoid creating a new ByteBuffer per each reading
-    // (for example binary search through the keys)
-    // keep persistent ByteBuffer objects referring to a slice from a Block's big underlying buffer
-    // in order to make it thread-safe keep separate persistent ByteBuffer per thread
-    private ThreadIndexCalculator threadIndexCalculator;
-    private ByteBuffer[] byteBufferPerThread;
 
     Block(long capacity) {
         assert capacity > 0;
         assert capacity <= Integer.MAX_VALUE; // This is exactly 2GB
         this.capacity = (int) capacity;
-        this.threadIndexCalculator = ThreadIndexCalculator.newInstance();
-        this.byteBufferPerThread = new ByteBuffer[ThreadIndexCalculator.MAX_THREADS];
         this.id = INVALID_BLOCK_ID;
         // Pay attention in allocateDirect the data is *zero'd out*
         // which has an overhead in clearing and you end up touching every page
-        this.buffer = ByteBuffer.allocateDirect(this.capacity);
+        this.writeBuffer = ByteBuffer.allocateDirect(this.capacity);
+        this.readBuffer = this.writeBuffer.asReadOnlyBuffer();
     }
 
     void setID(int id) {
@@ -52,21 +52,15 @@ class Block {
             allocated.getAndAdd(-size);
             throw new OakOutOfMemoryException();
         }
-        // The position and limit of this thread's buffer does not change here.
-        // Any slicing/positioning will be made on demand by the Slice instance itself.
-        // This means that a thread can allocate two slices from the same block at the same time without
-        // duplicating one of them.
         s.update(id, now, size);
-        getBufferForThread(s);
+        readByteBuffer(s);
         return true;
     }
 
     // use when this Block is no longer in any use, not thread safe
     // It sets the limit to the capacity and the position to zero, but didn't zeroes the memory
     void reset() {
-        buffer.clear(); // reset the position
         allocated.set(0);
-        this.threadIndexCalculator = ThreadIndexCalculator.newInstance();
     }
 
     // return how many bytes are actually allocated for this block only, thread safe
@@ -78,7 +72,7 @@ class Block {
     void clean() {
         Field cleanerField = null;
         try {
-            cleanerField = buffer.getClass().getDeclaredField("cleaner");
+            cleanerField = writeBuffer.getClass().getDeclaredField("cleaner");
         } catch (NoSuchFieldException e) {
             e.printStackTrace();
         }
@@ -86,7 +80,7 @@ class Block {
         cleanerField.setAccessible(true);
         Cleaner cleaner = null;
         try {
-            cleaner = (Cleaner) cleanerField.get(buffer);
+            cleaner = (Cleaner) cleanerField.get(writeBuffer);
         } catch (IllegalAccessException e) {
             e.printStackTrace();
         }
@@ -94,23 +88,8 @@ class Block {
         cleaner.clean();
     }
 
-    private ByteBuffer getMyBuffer() {
-        int idx = threadIndexCalculator.getIndex();
-        if (byteBufferPerThread[idx] == null) {
-            // the new buffer object is needed for thread safeness, otherwise
-            // (in single threaded environment)
-            // the setting of position and limit could happen on the main buffer itself
-            // but it happens only once per thread id
-            byteBufferPerThread[idx] = buffer.duplicate();
-        }
-
-        return byteBufferPerThread[idx];
-    }
-
-    void getBufferForThread(Slice s) {
-        // Were not slicing the buffer and not changing its position/limit.
-        // Any slicing/positioning will be made on demand by the Slice instance itself.
-        s.setBuffer(getMyBuffer());
+    void readByteBuffer(Slice s) {
+        s.setBuffer(readBuffer, writeBuffer);
     }
 
     // how many bytes a block may include, regardless allocated/free
--- a/core/src/main/java/com/oath/oak/Chunk.java
+++ b/core/src/main/java/com/oath/oak/Chunk.java
@@ -8,7 +8,6 @@ package com.oath.oak;
 
 import sun.misc.Unsafe;
 
-import java.nio.ByteBuffer;
 import java.util.EmptyStackException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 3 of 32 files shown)


core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -42,7 +42,7 @@
     /*-------------- Members --------------*/
 
     private static final Unsafe unsafe = UnsafeUtils.unsafe;
-    ByteBuffer minKey;       // minimal key that can be put in this chunk
+    MinKeyReadBuffer minKey;       // minimal key that can be put in this chunk
```
sanastas: I understand the reason for some special case of buffer here. But the solution looks to me too narrow. At least I would create an "OnHeapOakAttachedReadBuffer" class. However, the best solution looks to me, to create just KeyBuffer or OakAttachedReadBuffer out of min key (copy it off-heap). Later in rebalance just reference to the relevant off-heap buffer for the min key, anyway in OakMap keys are never released (even if deleted) until chunk rebalances...

sanastas:
A general comment regarding performance. I am referring to the following comment by Liran (explaining performance degradation for put-only scenario and the solution to it):

  1. Additional try/catch block
  2. Two additional method calls (passing lambda expression to a method to execute the serialization).
When I removed these, the performance was comparable to the "master".

First, I still see lambda expression in the serialization. Either I do not understand something, or it is old version of the code. What is right?

Second, if removing "passing lambda expression to a method" improves performance why not to do it more? There are lots of lambdas in the code...

core/src/main/java/com/oath/oak/MinKeyReadBuffer.java
```diff
@@ -0,0 +1,126 @@
...
+
+package com.oath.oak;
+
+import sun.nio.ch.DirectBuffer;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.function.Consumer;
+
+class MinKeyReadBuffer implements OakReadBuffer, OakUnsafeDirectBuffer {
```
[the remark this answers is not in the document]
sanastas: Here we allocate the buffer off-heap but not via our memory manager. Meaning we allocate and release on demand via OS and JVM. We know that direct off-heap allocation on demand is slow. So I do not think it is a good idea to do so... I have already explained the idea to use OakAttachedReadBuffer and reference to any such key, in other comment...

core/src/main/java/com/oath/oak/MinKeyWriteBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+ * Copyright 2018 Oath Inc.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.oath.oak;
+
+import java.nio.ByteBuffer;
+
+class MinKeyWriteBuffer extends MinKeyReadBuffer implements OakWriteBuffer, OakUnsafeDirectBuffer {
```
sanastas: Again if we use existing key, min key doesn't need to be written. Two create two additional classes just for that sounds to me over-engineering. Also that it looks to me not effective, if we need to change it from ByteBuffer, let's use existing keys.

core/src/main/java/com/oath/oak/OakAttachedReadBuffer.java
```diff
@@ -37,48 +37,57 @@ public int capacity() {
...
-    public ByteBuffer getByteBuffer() {
-        return getDuplicatedReadByteBuffer();
+    /*-------------- Internal --------------*/
+
+    /**
+     * This method should be used internally by Oak to write to the buffer.
+     * It creates a new temporary writable buffer and disable it once the write is done.
+     * @param writer a method that writes to a writable buffer
+     */
+    void write(Consumer<OakWriteBuffer> writer) {
```
sanastas: Is it that try/catch that you saw worth eliminating? In general write of the read buffer looks very awkward to me...

core/src/main/java/com/oath/oak/MinKeyBuffer.java
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+package com.oath.oak;
+
+import sun.nio.ch.DirectBuffer;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+class MinKeyBuffer implements OakWriteBuffer, OakUnsafeDirectBuffer {
```
sanastas, line 14: The comment on the previous implementation of the min key buffer class is still relevant with the new code. Main problems:
1. The fact that the buffer is written in the time of its construction doesn't make it writable. It is read only buffer. I believe it should be OakAttachedReadBuffer/KeyBuffer.
2. It off-heap allocation on demand and it is costly. Either put on-heap or... Better to use OakAttachedReadBuffer/KeyBuffer and allocate via reused memory

core/src/main/java/com/oath/oak/OakDetachedReadBuffer.java
```diff
@@ -91,7 +91,7 @@ public double getDouble(int index) {
         if (transformer == null) {
             throw new NullPointerException();
         }
-        return safeAccessToAttachedBuffer(buffer -> transformer.apply(buffer.getDuplicatedReadByteBuffer().slice()));
+        return safeAccessToAttachedBuffer(transformer::apply);
```
sanastas, line 94: I wonder if this syntactic sugar also implies some performance degradation...

core/src/main/java/com/oath/oak/Slice.java
```diff
@@ -12,7 +12,7 @@
 
 // Represents a portion of a bigger block which is part of the underlying managed memory.
 // It is allocated via block memory allocator, and can be de-allocated later
-class Slice implements Comparable<Slice> {
+class Slice implements OakUnsafeDirectBuffer, Comparable<Slice> {
```
sanastas, line 15: Wasn't it already? Weird...

core/src/main/java/com/oath/oak/Slice.java
```diff
@@ -139,43 +145,29 @@ int getVersion() {
     }
 
     long getMetadataAddress() {
-        return ((DirectBuffer) buffer).address() + offset;
+        return ((DirectBuffer) readBuffer).address() + offset;
```
sanastas, line 148: Metadata is also accessed for write... but I assume it is not a problem since it is just an address. But for clearness I would use write buffer

core/src/main/java/com/oath/oak/Slice.java
```diff
@@ -29,7 +29,8 @@
     // Allocation time version
     protected int version;
 
-    protected ByteBuffer buffer;
+    protected ByteBuffer readBuffer;
+    protected ByteBuffer writeBuffer;
```
sanastas, line 33: I currently do not see where writeBuffer and not readBuffer is used, and not both... What writeBuffer is used for?

core/src/main/java/com/oath/oak/ValueUtils.java
```diff
@@ -137,15 +110,15 @@
      * Used to try and read a value off-heap
      *
      * @param result      The result object
-     * @param s           the value's off-heap object
+     * @param value       the value's off-heap object
```
sanastas, line 113: off-heap object --> buffer
everywhere

core/src/main/java/com/oath/oak/Block.java
```diff
@@ -16,28 +16,21 @@
 
 class Block {
 
-    private final ByteBuffer buffer;
+    private final ByteBuffer readBuffer;
```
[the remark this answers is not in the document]
liran-funaro: done

benchmarks/synchrobench/src/main/java/com/oath/oak/synchrobench/MyBuffer.java
```diff
@@ -1,8 +1,7 @@
 package com.oath.oak.synchrobench;
 
 
-import com.oath.oak.OakComparator;
-import com.oath.oak.OakSerializer;
+import com.oath.oak.*;
```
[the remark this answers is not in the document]
liran-funaro: done

core/src/main/java/com/oath/oak/MinKeyReadBuffer.java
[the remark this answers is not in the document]
liran-funaro: Modify to allocate min-key using Oak's allocator

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -602,9 +602,7 @@ private void writeKey(ThreadContext ctx, K key) {
         int intIdx = entryIdx2intIdx(ctx.entryIndex);
 
         memoryManager.allocate(ctx.key, keySize, MemoryManager.Allocate.KEY);
-        // byteBuffer.slice() is set so it protects us from the overwrites of the serializer
-        // TODO: better serializer need to be given OakWBuffer and not ByteBuffer
-        keySerializer.serialize(key, ctx.key.getDataByteBuffer().slice());
+        ctx.key.write(writeKey -> keySerializer.serialize(key, writeKey));
```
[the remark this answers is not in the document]
liran-funaro: Outdated. Disable the buffer inline now.

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -645,10 +643,7 @@ void writeValueStart(ThreadContext ctx, V value, boolean writeForMove) {
             valOffHeapOperator.initHeader(ctx.newValue);
         }
 
-        // since this is a private environment, we can only use ByteBuffer::slice, instead of ByteBuffer::duplicate
-        // and then ByteBuffer::slice
-        // To be safe we create a new ByteBuffer object (for the serializer).
-        valueSerializer.serialize(value, valOffHeapOperator.getValueByteBufferNoHeaderPrivate(ctx.newValue));
+        ctx.newValue.write(writeValue -> valueSerializer.serialize(value, writeValue));
```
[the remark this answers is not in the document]
liran-funaro: Outdated. Disable the buffer inline now.

core/src/main/java/com/oath/oak/Chunk.java
liran-funaro: Modify to allocate min-key using Oak's allocator.
Future work: to modify min-key to be inclusive, then we could use an existing object as min-key.

core/src/main/java/com/oath/oak/MinKeyWriteBuffer.java
liran-funaro: Modify to allocate min-key using Oak's allocator

core/src/main/java/com/oath/oak/OakAttachedReadBuffer.java
liran-funaro: Outdated.

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -645,10 +645,9 @@ void writeValueStart(ThreadContext ctx, V value, boolean writeForMove) {
             valOffHeapOperator.initHeader(ctx.newValue);
         }
 
-        // since this is a private environment, we can only use ByteBuffer::slice, instead of ByteBuffer::duplicate
-        // and then ByteBuffer::slice
-        // To be safe we create a new ByteBuffer object (for the serializer).
-        valueSerializer.serialize(value, valOffHeapOperator.getValueByteBufferNoHeaderPrivate(ctx.newValue));
+        OakAttachedWriteBuffer writeBuffer = new OakAttachedWriteBuffer(ctx.newValue);
```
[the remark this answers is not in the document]
liran-funaro: Changed `OakAttachedWriteBuffer` constructor to be private and only instantiate and disable the buffer in two internal static methods: `static <T> void serialize(Slice s, T obj, OakSerializer<T> serializer)` and `static void compute(Slice s, Consumer<OakWriteBuffer> computer)`.
Both are now properly documented.
This removes code duplication without adding overhead due to the fact that these are static method calls (though we should validate it via Oak's benchmarks).

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -602,9 +602,9 @@ private void writeKey(ThreadContext ctx, K key) {
         int intIdx = entryIdx2intIdx(ctx.entryIndex);
 
         memoryManager.allocate(ctx.key, keySize, MemoryManager.Allocate.KEY);
-        // byteBuffer.slice() is set so it protects us from the overwrites of the serializer
-        // TODO: better serializer need to be given OakWBuffer and not ByteBuffer
-        keySerializer.serialize(key, ctx.key.getDataByteBuffer().slice());
+        OakAttachedWriteBuffer writeBuffer = new OakAttachedWriteBuffer(ctx.key);
```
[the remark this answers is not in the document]
liran-funaro: See comment below.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -992,7 +989,7 @@ boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializ
         while (chunkIter.hasNext()) {
             int nextIndex = chunkIter.next(ctx);
             c.readKeyFromEntryIndex(ctx.tempKey, nextIndex);
-            int cmp = comparator.compareKeyAndSerializedKey(key, ctx.tempKey.getDataByteBuffer());
+            int cmp = comparator.compareKeyAndSerializedKey(key, ctx.tempKey);
```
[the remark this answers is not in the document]
liran-funaro: I'll do you one more. I will just use the existing one in `Chunk`.

core/src/main/java/com/oath/oak/MinKeyBuffer.java
```diff
@@ -0,0 +1,159 @@
...
+        buffer = ByteBuffer.allocateDirect(keySerializer.calculateSize(minKey));
+        keySerializer.serialize(minKey, this);
+        disable();
+    }
+
+    MinKeyBuffer(OakReadBuffer src) {
+        final int size = src.capacity();
+        buffer = ByteBuffer.allocateDirect(size);
+        for (int i = 0; i < size; i++) {
+            buffer.put(i, src.get(i));
```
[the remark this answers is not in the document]
liran-funaro: We might not need it in the future if we change Chunk.minKey to be inclusive.
But for now, I changed it to:
```java
UnsafeUtils.unsafe.copyMemory(src.getAddress(), dst.getAddress(), keySize);
```

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1345,7 +1340,7 @@ private void advanceState() {
             // The check of the boundaries under condition is an optimization.
```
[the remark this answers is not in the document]
liran-funaro: Removed.

core/src/main/java/com/oath/oak/MinKeyBuffer.java
liran-funaro: See comment about MinKeyBuffer above.

core/src/main/java/com/oath/oak/OakDetachedReadBuffer.java
liran-funaro: I believe that the cost of using lambda come from the need to instantiate a new object.
The `transformer::apply` "syntactic sugar" is actually a necessary syntax to allow usage of exiting class methods as input to functions that accept a lambda function, without having to instantiate a new object (i.e., the lambda function object).
So we actually improve performance when we use this syntax, which implies it is not just syntactic sugar.

The actual possible performance issue is a few lines above in the `get{Something}()` methods that instantiate a new lambda function for each get operation.
But we currently do not have any benchmark that evaluates the detached buffers.

core/src/main/java/com/oath/oak/Slice.java
liran-funaro: Before the new-api, the implementation of `OakUnsafeDirectBuffer.getByteBuffer()` would have been oakward (pun intended) in `Slice`, so we couldn't have supported the full `OakUnsafeDirectBuffer` in `Slice`.
Now it is straight forward.

core/src/main/java/com/oath/oak/Slice.java
liran-funaro: done

core/src/main/java/com/oath/oak/Slice.java
liran-funaro: `writeBuffer` is used in `OakAttachedWriteBuffer`.
It must be stored in `Slice` because `OakAttachedWriteBuffer` is instantiated from an existing `Slice` object.

core/src/main/java/com/oath/oak/ValueUtilsImpl.java
```diff
@@ -76,8 +76,9 @@ private static boolean CAS(Slice s, int expectedLock, int newLock, int version)
         if (capacity > ctx.value.getLength()) {
             return moveValue(chunk, ctx, memoryManager, internalOakMap, newVal);
         }
-        ByteBuffer bb = getValueByteBufferNoHeader(ctx.value);
-        serializer.serialize(newVal, bb);
+        OakAttachedWriteBuffer writeBuffer = new OakAttachedWriteBuffer(ctx.value);
```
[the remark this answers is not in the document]
liran-funaro: I already implemented a static serialize method that ensures our unique instantiate/write/disable procedure.
See my other comment on that issue.

core/src/main/java/com/oath/oak/ValueUtilsImpl.java
```diff
@@ -76,8 +76,9 @@ private static boolean CAS(Slice s, int expectedLock, int newLock, int version)
...
+        serializer.serialize(newVal, writeBuffer);
```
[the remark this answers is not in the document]
liran-funaro: The header size is written in the `Slice` object. It is initialized when we instantiate the `KeyBuffer` or `ValueBuffer` object.
So `OakAttachedReadBuffer` and `OakAttachedWriteBuffer` take the header size into account.

core/src/main/java/com/oath/oak/ValueUtils.java
liran-funaro: I changed it as per your request. But please note that it technically not a buffer. It is a `Slice`.
If we want to align ourselves with referring the off-heap handlers as buffers, we might want to rename `Slice` to `OffheapBuffer`.

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1145,9 +1141,9 @@ public final boolean hasNext() {
         }
 
         private void initAfterRebalance() {
-            //TODO - refactor to use ByeBuffer without deserializing.
+            //TODO - refactor to use OakReadBuffer without deserializing.
```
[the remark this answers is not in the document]
liran-funaro: I don't think the same. I think this comment is true and should be added to our future-work list.
It might require adding a few internal methods to support comparing two sterilized keys, but it might improve our performance in some workloads.

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -120,7 +121,7 @@
     static final ReferenceCodec VALUE = new ReferenceCodec(32, 23, 9);
 
     private static final Unsafe unsafe = UnsafeUtils.unsafe;
-    private final MemoryManager memoryManager;
+    final MemoryManager memoryManager;
```
sanastas, line 124: Why "private" was removed?

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
```diff
@@ -25,6 +31,37 @@ void disable() {
         enabled = false;
```
sanastas, line 31: Should disable() also be private?

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
```diff
@@ -34,56 +71,57 @@ void validateAccess() {
...
 
     @Override
     public OakWriteBuffer putDouble(int index, double value) {
         validateAccess();
-        getDataByteBuffer().putDouble(getDataOffset(index), value);
+        writeBuffer.putDouble(getDataOffset(index), value);
         return this;
     }
 
     /*-------------- OakUnsafeDirectBuffer --------------*/
```
sanastas, line 120: I am may be confused, where getAddress/getOffset are implemented?

core/src/main/java/com/oath/oak/EntrySet.java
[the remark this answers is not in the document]
sanastas: Good!

core/src/main/java/com/oath/oak/EntrySet.java
liran-funaro: I removed it because I access it from `Chunk`'s constructor.

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
liran-funaro: Yeah, why not.

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
liran-funaro: They are implemented in `Slice`

core/src/main/java/com/oath/oak/InternalOakMap.java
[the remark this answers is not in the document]
sanastas: "sterilized keys" is it something new due to COVID-19? :) :) 

So as a future work, do you suggest to keep hi/lo boundaries in serialized way from the start?

core/src/main/java/com/oath/oak/OakAttachedReadBuffer.java
```diff
@@ -37,48 +37,41 @@ public int capacity() {
...
 
     @Override
     public double getDouble(int index) {
-        return getDataByteBuffer().getDouble(getDataOffset(index));
-    }
-
-    /*-------------- OakUnsafeDirectBuffer --------------*/
-
-    @Override
-    public ByteBuffer getByteBuffer() {
```
sanastas, line 81: Does OakUnsafeDirectBuffer interface come from Slice?

core/src/main/java/com/oath/oak/OakDetachedReadBuffer.java
sanastas: Agree. Just add a TODO comment for
> The actual possible performance issue is a few lines above in the get{Something}() methods that instantiate a new lambda function for each get operation.

core/src/main/java/com/oath/oak/OakNativeMemoryAllocator.java
```diff
@@ -195,9 +195,7 @@ public boolean isClosed() {
     @Override
     public void readByteBuffer(Slice s) {
         Block b = blocksArray[s.getAllocatedBlockID()];
-        // The returned buffer is this thread's block buffer.
-        // Therefore, a thread cannot read two slices from the same block without duplicating one of them.
-        b.getBufferForThread(s);
+        b.getBuffer(s);
```
sanastas, line 198: The opposite order looks more reasonable to me. Like slice.setBuffer(b.getBuffer)... Currently the way it works is obscure and looks like block is getting a buffer from slice, from the first glance...

core/src/main/java/com/oath/oak/Chunk.java
```diff
@@ -42,57 +41,85 @@
...
-     * @param minKey  minimal key to be placed in chunk
-     * @param creator the chunk that is responsible for this chunk creation
+     * This constructor is only used when creating the first chunk (without a creator).
+     * It does not initialized the minKey, so it must follow a call to {@code setMinKey(K)}.
      */
-    Chunk(ByteBuffer minKey, Chunk<K, V> creator, OakComparator<K> comparator, MemoryManager memoryManager,
-          int maxItems, AtomicInteger externalSize, OakSerializer<K> keySerializer, OakSerializer<V> valueSerializer,
-          ValueUtils valueOperator) {
-
+    Chunk(int maxItems, AtomicInteger externalSize, MemoryManager memoryManager, OakComparator<K> comparator,
```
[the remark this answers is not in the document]
liran-funaro: I considered this option. But notice that we could not differentiate between a constructor that should just copy the key reference (`copyFromMinKey`), from the constructor that duplicates the key (`duplicateMinKey`). They would have the same signature.
So I ended up using this design.

Another approach is to make the constructor public and adding two "createChild" methods.
The first-ever chunk will be created using the constructor (scenario 1).
For scenarios 2 and 3, we will add two methods:
* `c.createChild()` - create a child `Chunk` with the same `minKey` without duplication.
* `c.createChild(KeyBuffer minKey)` - create a child `Chunk` with a `minKey` that is duplicated from the input key.

In addition, we could add another constructor that creates a `Chunk` from a `minKey` object of type `T` and make the other constructor private.

core/src/main/java/com/oath/oak/Slice.java
sanastas: Please add a comment about that, so it is clearer...

core/src/main/java/com/oath/oak/ValueUtils.java
sanastas: Actually you are right it needs to be Slice and everywhere...

core/src/main/java/com/oath/oak/InternalOakMap.java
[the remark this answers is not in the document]
liran-funaro: Not necessarily. This should work fine with either serialized or non-serialized keys.
But obviously we should put more thought into this.

core/src/main/java/com/oath/oak/Slice.java
liran-funaro: We already have a comment for that in `Block`.
I'm against adding repeating comments in multiple places. If you feel this is necessary, then I suggest adding a comment in the form of "See Block for more details".

core/src/main/java/com/oath/oak/OakAttachedReadBuffer.java
liran-funaro: Yup.

core/src/main/java/com/oath/oak/OakNativeMemoryAllocator.java
liran-funaro: Note that `Block` has two buffers. So what you are suggesting is actually:
```java
s.setBuffer(b.getReadBuffer(), b.getWriteBuffer())
````
And this line would repeat twice. Once in `OakNativeMemoryAllocator.readByteBuffer(Slice)`, and once in `Block.allocate(Slice, int)`.

Your suggestion also breaks the convention that buffers are never exposed directly, but rather passed using a `Slice` objects.

If it looked obscure it might be because of the method name `Block.getBuffer(Slice)`.
How about using the same name from `OakBlockMemmoryAllocator`: `void readByteBuffer(Slice s)`?

core/src/main/java/com/oath/oak/ValueUtils.java
liran-funaro: Eventually, we will have to find a more suitable name for `Slice`.
I don't really understand what was wrong with the name you suggested in the paper: `Handler`, or `OffheapHandler`.
But for now, I'll change the comment to `the value's off-heap Slice object`.

core/src/main/java/com/oath/oak/InternalOakMap.java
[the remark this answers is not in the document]
sanastas: OK, I do not see what can be done there, but I put it on the future-TODO list. So we will discuss it later.

core/src/main/java/com/oath/oak/ValueUtils.java
sanastas: OK. Slice is used in the Nova paper. I like this name and want to keep it. But this is not related here.

core/src/main/java/com/oath/oak/OakNativeMemoryAllocator.java
sanastas: Yes, at least rename it

core/src/main/java/com/oath/oak/OakDetachedReadBuffer.java
liran-funaro: Found a workaround that avoids instantiating new lambda objects.

core/src/main/java/com/oath/oak/Chunk.java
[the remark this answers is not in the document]
liran-funaro: Made the agreed-upon changes.

core/src/main/java/com/oath/oak/OakNativeMemoryAllocator.java
liran-funaro: done

PR #104, opened 2020-05-31
Add checkstyle plugin to the core module and fix style issues accordingly

liran-funaro:
The full style description is in the `codestyle/checkstyle.xml` file.

Also, normalize comments to a unified style.


core/src/main/java/com/oath/oak/BlocksPool.java
```diff
@@ -15,7 +15,7 @@
  * However it makes creation of the first Oak slower. This initialization is thread safe, thus
  * multiple concurrent Oak creations will result only in the one Pool.
  * */
-class BlocksPool implements BlocksProvider, Closeable {
+final class BlocksPool implements BlocksProvider, Closeable {
```
sanastas, line 18: Is this 'final' and the following 'static' fields addition, part of the code styling?

core/src/main/java/com/oath/oak/NovaManager.java
```diff
@@ -7,10 +7,10 @@
 
 class NovaManager implements MemoryManager {
     static final int RELEASE_LIST_LIMIT = 1024;
-    private ThreadIndexCalculator threadIndexCalculator;
-    private List<List<Slice>> releaseLists;
-    private AtomicInteger globalNovaNumber;
-    private OakBlockMemoryAllocator allocator;
+    private final ThreadIndexCalculator threadIndexCalculator;
```
sanastas, line 10: I wonder why it would be requested to make those fields final...?

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -281,9 +287,8 @@ private void setEntryFieldLong(int item, OFFSET offset, long value) {
      * CAS from 'expectedIntValue' to 'newIntValue' for field at specified offset
      * of given int-field in the entries array
      */
-    private boolean casEntriesArrayInt(
-        int intFieldIdx, OFFSET offset, int expectedIntValue, int newIntValue) {
-        return unsafe.compareAndSwapInt(entries,
+    private boolean casEntriesArrayInt(int intFieldIdx, OFFSET offset, int expectedIntValue, int newIntValue) {
```
sanastas, line 290: Are there no boundaries on long code line?

core/src/main/java/com/oath/oak/EntrySet.java
```diff
@@ -846,37 +854,41 @@ boolean isEntryDeleted(ValueBuffer tempValue, int ei) {
...
      * @param srcEntrySet another EntrySet to copy from
      * @param srcEntryIdx the entry index to copy from {@code srcEntrySet}
-     * @return            false when this EntrySet is full
-     *
-     * NOT THREAD SAFE
-     * */
-    boolean copyEntry(ValueBuffer tempValue, EntrySet<K,V> srcEntrySet, int srcEntryIdx) {
+     * @return false when this EntrySet is full
+     * <p>
+     * Nope: NOT THREAD SAFE
```
sanastas, line 865: What is "Nope"? Note?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -21,9 +21,6 @@
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import static com.oath.oak.Chunk.*;
```
sanastas, line 24: Wasn't/isn't it needed?

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -676,10 +675,12 @@ Result remove(K key, V oldValue, OakTransformer<V> transformer) {
             if (!ctx.isKeyValid()) {
                 // There is no such key. If we did logical deletion and someone else did the physical deletion,
                 // then the old value is saved in v. Otherwise v is (correctly) null
-                return transformer == null ? ctx.result.withFlag(logicallyDeleted ? TRUE : FALSE) : ctx.result.withValue(v);
+                return transformer == null ? ctx.result.withFlag(logicallyDeleted ? ValueUtils.ValueResult.TRUE :
```
sanastas, line 678: Here and below, hard to understand what is compared with what and what is returned. Please put some parenthesis

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
```diff
@@ -13,13 +13,14 @@
  * An instance of this buffer is only used when the write lock of the key/value referenced by it is already acquired.
  * This is the reason no lock is acquired in each access.
  */
-class OakAttachedWriteBuffer extends OakAttachedReadBuffer implements OakWriteBuffer, OakUnsafeDirectBuffer {
+final class OakAttachedWriteBuffer extends OakAttachedReadBuffer implements OakWriteBuffer, OakUnsafeDirectBuffer {
```
sanastas, line 16: Is it that any class that none inherits from it, is requested to be final?

core/src/main/java/com/oath/oak/Operation.java
```diff
@@ -1,18 +0,0 @@
-/*
```
sanastas, line 1: So sad, rebalance is farer and farer away from what it needs to be :(

core/src/main/java/com/oath/oak/EntrySet.java
liran-funaro: Sure there is. See the `codestyle/checkstyle.xml`. I set it to 120. It seems the correct length because most of the code conforms to this length, and it is also Intellij defaults.
This line (and now all the other lines) conforms to this limitation.

core/src/main/java/com/oath/oak/BlocksPool.java
liran-funaro: Yes. The final is because this class has only private constructors, meaning it already orbit inheritance. The checkstyle plugin enforce the final modifier.
Below, I didn't add the static modifier, I only changed the modifiers order to conform with Java's recommended order.
This is also enforced by the plugin.

core/src/main/java/com/oath/oak/NovaManager.java
liran-funaro: Actually, this is not part of the checkstyle, but rather an Intellij suggestion.
It makes sense to make these members final because they never change.
Making this explicit helps code readability.

core/src/main/java/com/oath/oak/EntrySet.java
liran-funaro: Yes. fixed.

core/src/main/java/com/oath/oak/InternalOakMap.java
liran-funaro: The checkstyle forbids static imports. It makes less readable code.
So now each access to Chunk's static member would have to be something like `Chunk.someStaticMember` instead of `someStaticMember`.
I personally agree that this is clearer to the reader.

core/src/main/java/com/oath/oak/OakAttachedWriteBuffer.java
liran-funaro: No. Just classes that have only private constructors. There isn't a way to inherit from a class with a private constructor, so it should be final.

core/src/main/java/com/oath/oak/Operation.java
liran-funaro: I don't understand this comment. How `Opetation.java` relate to rebalancing?

core/src/main/java/com/oath/oak/InternalOakMap.java
liran-funaro: I don't see how parenthesis would help.
The only change is that now TRUE and FALSE are addressed directly instead of via a static import.
It easier to read because you can immediately know where those TRUE/FALSE values come from.

PR #107, opened 2020-06-01
External and internal class renaming

liran-funaro:
- Remove redundant "Oak" prefix to all internal classes
 - Changed attached/detached to scoped/unscoped
 - Add OakBuffer interface as a parent of OakScopedReadBuffer and OakUnscopedBuffer


core/src/main/java/com/oath/oak/BlockMemoryAllocator.java
```diff
@@ -13,7 +13,7 @@
  * any thread.
  * Note that two allocations cannot be merged into a single allocation, and an allocation currently is not split.
  */
-interface OakBlockMemoryAllocator {
+interface BlockMemoryAllocator {
```
sanastas, line 16: This was initially an outside interface... Before the package was removed, so it is a candidate to be exposed outside in the future...

core/src/main/java/com/oath/oak/InternalOakMap.java
```diff
@@ -1415,50 +1415,50 @@ public T next() {
...
 
-    class EntryStreamIterator extends Iter<Map.Entry<OakDetachedBuffer, OakDetachedBuffer>>
-            implements Map.Entry<OakDetachedBuffer, OakDetachedBuffer> {
+    class EntryStreamIterator extends Iter<Map.Entry<OakUnscopedBuffer, OakUnscopedBuffer>>
+            implements Map.Entry<OakUnscopedBuffer, OakUnscopedBuffer> {
 
-        private final OakDetachedReadBuffer<KeyBuffer> key = new OakDetachedReadBuffer<>(new KeyBuffer());
-        private final OakDetachedReadBuffer<ValueBuffer> value =
-                new OakDetachedReadBuffer<>(new ValueBuffer(valueOperator));
+        private final UnscopedBuffer<KeyBuffer> key = new UnscopedBuffer<>(new KeyBuffer());
```
sanastas, line 1427: So there is no unscoped-read/write- buffer anymore?

core/src/main/java/com/oath/oak/OakBuffer.java
```diff
@@ -12,17 +12,8 @@
...
- * The behaviour of this buffer outside their attached scope is undefined.
- * Such callback method might be the application's serializer and comparator, or a lambda function that can
- * read/store/update the data.
- * This access reduces unnecessary copies and deserialization of the underlying data.
- * In their intended context, the user does not need to worry about concurrent accesses and memory management.
- * Using these buffers outside their intended context may yield unpredicted results, e.g., reading non-consistent data
- * and/or irrelevant data.
  */
-public interface OakReadBuffer {
+public interface OakBuffer {
```
sanastas, line 16: Didn't we say it is not going to be public?

liran-funaro: Correct. Will fix.

core/src/main/java/com/oath/oak/InternalOakMap.java
liran-funaro: Unscoped write buffer never existed before.
Before, I added "read" to the unscoped buffer name. But I think it is redundant.
Don't you think?

core/src/main/java/com/oath/oak/OakBuffer.java
liran-funaro: After working on Druid, I realized I need a common interface between all types of buffers.
So I think this should remain public.

liran-funaro: This is a way for the developer to state: "I don't care what kind of buffer I get (scoped/unscoped/whatever)".

core/src/main/java/com/oath/oak/InternalOakMap.java
sanastas: Unscoped buffer indeed exists for read only. As it is internally used name I am OK with leaving it like this. Maximum we will change it later.

core/src/main/java/com/oath/oak/OakBuffer.java
sanastas: Hmm... OK...

sanastas: Then something need to be said about it in README PR#88

PR #108, opened 2020-06-03
Release notes

sanastas:
Adding the release notes to be updated from now on

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -1,8 +1,6 @@
-======================================================
 
+<hr>
 ## OAK 0.2 Release Notes (June 01, 2020)
->Compared to Oak 0.1.5 that was published on May 14, 2019 
-
 
  01. Keys Disaggregation: Make chunks only reference the keys, thus keys are not copied during the rebalance. Improving the rebalance and ingestion performance. 
  02. Stream Scans: Gives a "stream" view on the elements, meaning only one element can be observed at a time. It greatly improves the scan performance due to objects’ re-usage.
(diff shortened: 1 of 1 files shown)


release-notes.md
```diff
@@ -0,0 +1,18 @@
+======================================================
```
liran-funaro, line 1: What is this line?

release-notes.md
```diff
@@ -0,0 +1,18 @@
...
+
+## OAK 0.2 Release Notes (June 01, 2020)
+>Compared to Oak 0.1.5 that was published on May 14, 2019
```
liran-funaro, line 4: Maybe instead just add below a title for the last release with one sentence of what you remember it was about.
And remove this line from here. For example:

## v0.1.5 (May 14, 2019)
 * API changes

release-notes.md
liran-funaro: Maybe we should have a title here instead:
# Oak Release Notes

release-notes.md
```diff
@@ -0,0 +1,18 @@
...
+ 06. Refactoring for future OakHash: EntrySet class is introduced to encapsulate (1) the entries to integers mapping and (2) managing (allocate/read/write/delete/release) individual key-value entries. Helps in future OakHash coding.
+ 07. No internal objects creation: Eliminate ephemeral objects creation in time of lookup. It greatly reduces the work to be done by GC and improves performance.
+ 08. Safe/Unsafe API: Create a possibility to work with OakBuffers in a fully safe/protected mode, so no wrong memory can be accessed/written erroneously. Safe mode requires some performance degradation, thus an alternative unsafe mode is introduced, providing greater performance and relying on the correct user actions for the security.
+ 09. Improving benchmarks code: Reducing the time it takes to run all benchmarks. Improving the results readability. Improving the benchmark scripts user experience.
+ 10. Correctness bug fixes: Various bugs fixes: in non-zero-copy API, internal memory access, rebalance process, etc.
```
liran-funaro, line 16: You repeated "bug fixes" twice. The second "Various bugs fixes:" is redundant.
"etc." is incorrect here.
I suggest "Correctness bug fixes (partial list): ..."
The same below.

release-notes.md
sanastas: This line I thought to use for the future separation between releases. Like the future release notes should come above this line. I wanted to put it there so future updater won't forget to separate between releases. But if it is confusing I can remove it.

release-notes.md
sanastas: OK, I will do it.

release-notes.md
sanastas: OK, fixed

PR #115, opened 2020-07-09
Modify Block and Slice to hold a single buffer instead of a writable and a readable buffer

liran-funaro:

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,11 +392,6 @@ while (iter.hasNext()) {
 }
 ```
 
-Note: in the above example, the following will throw a `ReadOnlyBufferException` because the buffer mode is read-only:
-```java
-bb.putInt(ub.getOffset(), someInteger);
-```
-
 ##### Unsafe direct buffer access (address)
 
 ```java
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -14,14 +14,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 class Block {
 
-    /*
-    For performance reasons, we keep two ByteBuffer objects that refers to the same buffer.
-    writeBuffer is the original allocated buffer, and readBuffer is a duplicated version of writeBuffer
-    but with read-only access permissions.
-    This allow the user to supply the a read-only ByteBuffer (when needed) without instantiating a new object.
-     */
-    private final ByteBuffer readBuffer;
-    private final ByteBuffer writeBuffer;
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -392,11 +392,6 @@ while (iter.hasNext()) {
 }
 ```
 
-Note: in the above example, the following will throw a `ReadOnlyBufferException` because the buffer mode is read-only:
```
sanastas, line 395: I looked on the original README and didn't understand from where ReadOnlyBufferException should come... Where is the write attempt? Can you tell me?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -14,14 +14,7 @@
...
 class Block {
 
-    /*
-    For performance reasons, we keep two ByteBuffer objects that refers to the same buffer.
-    writeBuffer is the original allocated buffer, and readBuffer is a duplicated version of writeBuffer
-    but with read-only access permissions.
-    This allow the user to supply the a read-only ByteBuffer (when needed) without instantiating a new object.
-     */
-    private final ByteBuffer readBuffer;
-    private final ByteBuffer writeBuffer;
```
sanastas, line 24: Just wonder now... Many sketches may reside on the same block, it means they all have the same hash code. How will this sketches mapping work?

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
```diff
@@ -32,9 +32,6 @@
  * }
  * }
  * </pre>
- * <p>
- * Note: in the above example, the following will throw a ReadOnlyBufferException because the buffer mode is read-only:
```
sanastas, line 36: Again do not see where is the write attempt... Just curious! :)

README.md
liran-funaro: Below (line 397), we try to use `putInt()` on a read-only buffer.
This is no longer will issue an exception because the unsafe API will always produce a writable buffer.

core/src/main/java/com/yahoo/oak/Block.java
liran-funaro: Sketches use both the object-ID of the buffer and its position

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
liran-funaro: Again, below (line 37).

PR #116, opened 2020-07-16
Allow the user to choose a preferred block size

liran-funaro:
BlocksPool:
- Changed static configuration to use bytes instead of #blocks
- Changed #blocks limits to be initialized at the constructor according to the preferred block size
- Changed block size to be long to allow >4GB blocks

ReferenceCodec:
- Add a constructor with size limitations instead of explicit bit offsets
- Add a validation when encoding a reference

NativeMemoryAllocator:
- Add a validation when reading the buffer

EntrySet:
- Unify the key/value encoders
- Modify reference encoding to fit the new limitations and initialize it in the constructor so we can use the actual block size.

Other:
- Implement toString() for Slice (for debugging)
- Throw more informative IndexOutOfBoundsException() in ScopedReadBuffer

### DIFF
--- a/core/src/main/java/com/yahoo/oak/BlocksPool.java
+++ b/core/src/main/java/com/yahoo/oak/BlocksPool.java
@@ -22,33 +22,52 @@ final class BlocksPool implements BlocksProvider, Closeable {
 
     // TODO change the following constants to be configurable
 
-    // Size of a single memory block - currently 256MB
-    static final int BLOCK_SIZE = 256 * 1024 * 1024;
+    static final long MB = 1L << 20;
+    static final long GB = 1L << 30;
 
-    // Number of memory blocks to be pre-allocated
-    private static final int PRE_ALLOC_BLOCKS = 0;
+    // The memory size to pre-allocate on initialization.
+    private static final long PRE_ALLOC_SIZE_BYTES = 0;
 
-    // Number of memory blocks to be allocated at once when not enough blocks available
-    private static final int NEW_ALLOC_BLOCKS = 1;
+    // The minimal memory size to be allocated at once when not enough memory is available.
+    // If the used block size is larger than this number, exactly one block will be allocated.
+    // Otherwise, more than one block might be allocated at once.
+    private static final long NEW_ALLOC_MIN_SIZE_BYTES = 8L * MB;
 
-    // Upper/lower thresholds to the number of unused memory blocks to reserve in the pool for future use.
-    // When the number of unused blocks reaches HIGH_RESERVED_BLOCKS, a group of blocks will be freed
-    // such that the number of blocks will be LOW_RESERVED_BLOCKS.
-    private static final int HIGH_RESERVED_BLOCKS = 32;
-    private static final int LOW_RESERVED_BLOCKS = 24;
+    // Upper/lower thresholds to the quantity of the unused memory to reserve in the pool for future use.
+    // When the unused memory quantity reaches HIGH_RESERVED_SIZE_BYTES, some memory is freed
+    // such that the remaining unused memory will be LOW_RESERVED_SIZE_BYTES.
+    private static final long LOW_RESERVED_SIZE_BYTES = 2L * GB;
+    private static final long HIGH_RESERVED_SIZE_BYTES = 4L * GB;
 
-    private final int blockSize;
+    // The default size of a single memory block to be allocated at once.
+    static final int DEFAULT_BLOCK_SIZE_BYTES = 256 * (int) MB;
+
+    /**
+     * The block size in bytes that is used by this pool.
+     * It is limited to an integer duo to similar limitation of {@code ByteBuffer::allocateDirect(int capacity)}.
+     */
+    private final int blockSizeBytes;
+
+    private final int newAllocBlocks;
+    private final int lowReservedBlocks;
+    private final int highReservedBlocks;
 
     // not thread safe, private constructor; should be called only once
     private BlocksPool() {
-        this.blockSize = BLOCK_SIZE;
-        prealloc(PRE_ALLOC_BLOCKS);
+        this(DEFAULT_BLOCK_SIZE_BYTES);
     }
 
-    // used in tests only!!
-    private BlocksPool(int blockSize) {
-        this.blockSize = blockSize;
-        prealloc(PRE_ALLOC_BLOCKS);
+    // Used internally and for tests.
+    private BlocksPool(int blockSizeBytes) {
+        this.blockSizeBytes = blockSizeBytes;
+        this.newAllocBlocks = convertSizeToBlocks(NEW_ALLOC_MIN_SIZE_BYTES, 1);
+        this.lowReservedBlocks = convertSizeToBlocks(LOW_RESERVED_SIZE_BYTES, 0);
+        this.highReservedBlocks = convertSizeToBlocks(HIGH_RESERVED_SIZE_BYTES, this.lowReservedBlocks + 1);
+        alloc(convertSizeToBlocks(PRE_ALLOC_SIZE_BYTES, 0));
+    }
+
+    private int convertSizeToBlocks(long sizeBytes, int minBlocks) {
+        return Math.max(minBlocks, (int) Math.ceil((float) sizeBytes / (float) blockSizeBytes));
     }
 
     /**
@@ -76,9 +95,26 @@ final class BlocksPool implements BlocksProvider, Closeable {
         }
     }
 
+    /**
+     * Sets the preferred block size. This only has an effect if the block pool was never used before.
+     * @param preferredBlockSizeBytes the preferred block size
+     * @return true if the preferred block size matches the current instance
+     */
+    static boolean preferBlockSize(int preferredBlockSizeBytes) {
+        if (instance == null) {
+            synchronized (BlocksPool.class) { // can be easily changed to lock-free
+                if (instance == null) {
+                    instance = new BlocksPool(preferredBlockSizeBytes);
+                }
+            }
+        }
+
+        return instance.blockSizeBytes == preferredBlockSizeBytes;
+    }
+
     @Override
     public int blockSize() {
-        return blockSize;
+        return blockSizeBytes;
     }
 
     /**
@@ -97,7 +133,7 @@ final class BlocksPool implements BlocksProvider, Closeable {
             if (noMoreBlocks || b == null) {
                 synchronized (BlocksPool.class) { // can be easily changed to lock-free
                     if (blocks.isEmpty()) {
-                        prealloc(NEW_ALLOC_BLOCKS);
+                        alloc(newAllocBlocks);
                     }
                 }
             }
@@ -113,9 +149,9 @@ final class BlocksPool implements BlocksProvider, Closeable {
     public void returnBlock(Block b) {
         b.reset();
         blocks.add(b);
-        if (blocks.size() > HIGH_RESERVED_BLOCKS) { // too many unused blocks
+        if (blocks.size() > highReservedBlocks) { // too many unused blocks
             synchronized (BlocksPool.class) { // can be easily changed to lock-free
-                while (blocks.size() > LOW_RESERVED_BLOCKS) {
+                while (blocks.size() > lowReservedBlocks) {
                     this.blocks.poll().clean();
                 }
             }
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -152,6 +130,25 @@
         this.keySerializer = keySerializer;
         this.valueSerializer = valueSerializer;
         this.valOffHeapOperator = valOffHeapOperator;
+
+        /*
+         * Key/value reference encoding (64 bit):
```
sanastas, line 135: The meaning of number of bits in this comment is not very clear. Is it only upper/lower level limits? In general the comment is not clear.

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -152,6 +130,25 @@
...
+         * Key/value upper limits (default):
+         *   -       block size: 256MB (28 bit)
+         *                       This is the default size, but might be larger/smaller.
+         *   -      data length: 256MB (28 bit)
+         *                       The maximal data length matches the size of the block.
+         *   - number of blocks: 256 (8 bit)
+         *                       The remaining number of bits
+         *   => number of allocated 1K items: ~128 million items (2^26)
+         *
+         * TODO: block size should be derived directly from the memory manager.
```
sanastas, line 147: I agree that the reference structure should be derived from memory manager. But is BlockPool is being used or not is up to memory manager. BlockPool can be its integrated part.

core/src/main/java/com/yahoo/oak/OakMapBuilder.java
```diff
@@ -29,6 +29,7 @@
     private int chunkMaxItems;
     private long memoryCapacity;
     private BlockMemoryAllocator memoryAllocator;
+    private Long preferredBlockSize;
```
sanastas, line 32: long->int, specifically when able to be set from outside.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -22,33 +22,46 @@
...
 
-    // Number of memory blocks to be pre-allocated
-    private static final int PRE_ALLOC_BLOCKS = 0;
+    // The memory size to pre-allocate on initialization
+    private static final long PRE_ALLOC_SIZE_BYTES = 0;
 
-    // Number of memory blocks to be allocated at once when not enough blocks available
-    private static final int NEW_ALLOC_BLOCKS = 1;
+    // The memory size to be allocate at once when not enough memory is available
+    private static final long NEW_ALLOC_SIZE_BYTES = 8 * MB;
```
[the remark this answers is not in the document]
liran-funaro: Yes. This should be configurable by the user.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -22,33 +22,46 @@
...
-    private static final int LOW_RESERVED_BLOCKS = 24;
+    // Upper/lower thresholds to the quantity of the unused memory to reserve in the pool for future use.
+    // When the unused memory quantity reaches HIGH_RESERVED_SIZE_BYTES, some memory is freed
+    // such that the remaining unused memory will be LOW_RESERVED_SIZE_BYTES.
+    private static final long LOW_RESERVED_SIZE_BYTES = 2 * GB;
+    private static final long HIGH_RESERVED_SIZE_BYTES = 4 * GB;
 
-    private final int blockSize;
+    // The default size of a single memory block to be allocated at once
+    static final long DEFAULT_BLOCK_SIZE_BYTES = 256 * MB;
```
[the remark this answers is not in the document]
liran-funaro: I see now that `allocateDirect(int capacity)` accepts only ints anyway. I'll change it back to `int`.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -76,9 +89,25 @@ static void setBlockSize(int blockSize) {
         }
     }
 
+    /**
+     * Sets the preferred block size. This only has an effect if the block pool was never used before.
+     * @param preferredBlockSize the preferred block size
+     */
+    static void preferBlockSize(long preferredBlockSize) {
```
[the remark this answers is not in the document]
liran-funaro: Good idea. I'll do that.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -76,9 +89,25 @@ static void setBlockSize(int blockSize) {
...
+            if (instance == null) {
+                instance = new BlocksPool(preferredBlockSize);
+            }
+        }
+    }
+
     @Override
-    public int blockSize() {
-        return blockSize;
+    public long blockSize() {
```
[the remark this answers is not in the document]
liran-funaro: :+1:

core/src/main/java/com/yahoo/oak/EntrySet.java
liran-funaro: We don't have the bigger picture because we haven't discussed yet the details about the memory management package.
In general, I believe that the `MemoryManager` itself should be responsible for encoding the reference.
But for now, I think that `MemoryManager` should at least have an additional API method that returns the maximal block size: `int getMaximalBlockSize()`. This way we can avoid this issue for now.

core/src/main/java/com/yahoo/oak/OakMapBuilder.java
liran-funaro: :+1:

core/src/main/java/com/yahoo/oak/EntrySet.java
liran-funaro: It is the size limits that are derived from the representation size in bits.
> In general the comment is not clear.
Can you elaborate? What sentence does not makes sense to you?

core/src/main/java/com/yahoo/oak/BlocksPool.java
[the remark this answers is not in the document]
liran-funaro: > What do you think about setting PRE_ALLOC_SIZE_BYTES same way?

Currently, once `preferBlockSize()` is used, we instantiate a new `BlocksPool`.
If we want to add additional user preferences, we would have to defer the instantiation of the pool for later.
This will have a completely different design that I prefer not to get into right now.
Especially since we might refactor this anyway as part of the memory-manager package effort.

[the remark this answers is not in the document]
sanastas: OK

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: Bottom line we should deal with it later. But better remove the line speaking about BlockPool

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: It speaks about the limit in the bit size and says it is 28 bit and another 28 bits and then 8 remains. Or something like that. I believe the limit should be between x bits and till y bits depending on abc... Not a specific number. I find the current explanations confusing. I read it and didn't understand what the limits are...

liran-funaro: I re-edited this comment. I hope it is clearer now.

core/src/main/java/com/yahoo/oak/EntrySet.java
liran-funaro: I removed the TODO, but I think it is necessary to address that in the future. So at least note that in your Oak TODO list.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -22,33 +22,50 @@
...
 
-    // Number of memory blocks to be pre-allocated
-    private static final int PRE_ALLOC_BLOCKS = 0;
+    // The memory size to pre-allocate on initialization
+    private static final long PRE_ALLOC_SIZE_BYTES = 0;
 
-    // Number of memory blocks to be allocated at once when not enough blocks available
-    private static final int NEW_ALLOC_BLOCKS = 1;
+    // The memory size to be allocate at once when not enough memory is available
+    private static final long NEW_ALLOC_SIZE_BYTES = 8L * MB;
```
sanastas, line 32: Why 8MB? Isn't it the final version? This change was in order to leave default as it was or 128MB and make the change for Druid...

sanastas: I am missing something here... How blocks of 256MB is going to fit into 8MB? Is it because of the mathematical maximum always only one block is about to be allocated? I don't think it should be done this way...

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -660,7 +656,7 @@ void writeValueStart(ThreadContext ctx, V value, boolean writeForMove) {
         // The allocated slice is actually the thread's ByteBuffer moved to point to the newly
```
sanastas, line 656: It is important to mention that having a small block size (8MB vs 256MB) also limits the ability to write big values. The value size can be maximum the size of block. Currently, trial to allocate and write a value bigger than a block will result in out-of-bound exception. I suggest to add here a more clear exception, printing explanation that values bigger than block can not be allocated/written. After all it was a user decision to define small blocks...

core/src/main/java/com/yahoo/oak/BlocksPool.java
liran-funaro: It is actually the minimal size of a new allocation. It can be larger if the block size is larger.
I can change the name of this constant to `NEW_ALLOC_MIN_SIZE_BYTES` and change the comment as well.

> I don't think it should be done this way...

How do you think it should be?

core/src/main/java/com/yahoo/oak/EntrySet.java
liran-funaro: Please note that this problem does not relate to these code modifications, but we can address it now if you prefer.

**Background:**
Even with larger block sizes, the maximal value length was 8MB and the maximal key length was 64KB due to the reference encoding limitations.

With a small block size, trying to allocate a key/value larger than the block size will actually issue an `OakOutOfMemoryException` (see NativeMemoryAllocator:114-116), which might make sense albeit we can add there an additional message in the exception that explains the error.

On the other hand, trying to allocate a large value when using a larger block size will not be detected at all.
I somewhat addressed that in the `ReferenceCodec` class by adding an assertion there, but it only applied for testing.
Production code will not get ANY error until it will try to access the data and then it will simply get an out-of-bound exception as you mentioned. But that is also true before this PR.

My suggestions:
* Replacing the assertion in `ReferenceCodec` with an actual production-time validation that may raise `IllegalArgumentException` (may have some minor performance implications).
* Replacing the exception in NativeMemoryAllocator:115 also with `IllegalArgumentException`.

Let me know if you prefer to add these modifications now or discuss them further to be included in a different PR.

core/src/main/java/com/yahoo/oak/BlocksPool.java
sanastas: You can proceed with your suggestion:

>  I can change the name of this constant to NEW_ALLOC_MIN_SIZE_BYTES and change the comment as well.

Or to make NEW_ALLOC_SIZE_BYTES be equal to DEFAULT_BLOCK_SIZE_BYTES and change if blockSizeBytes is different.

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: I think here (when we first get the requested value size) is the place to check both for the block size limitation and for length encoding limitation. The information about what the limitations are, can be taken from ReferenceCodec. 

And although the length encoding limitation was already there, the block size limitation is harder now. When the possibility is given to the user to set the size of the block, it can be set to less than 8MB, thus this limit can be the first to fire.

core/src/main/java/com/yahoo/oak/BlocksPool.java
liran-funaro: Your suggestion nullifies this parameter altogether. It means that the number of allocated blocks is always 1.
Actually, they were always meaningless after my previous set of PRs. We left these parameters in the code only to allow us to fiddle with them for testing, but in production, they are actually useless.
So for now I'll go with `NEW_ALLOC_MIN_SIZE_BYTES`, but we can further discuss them as part of your new memory-manager design review.

PR #118, opened 2020-09-19
Changing EntrySet to be based on entries array based on longs instead of ints

sanastas:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/EntrySet.java
+++ b/core/src/main/java/com/yahoo/oak/EntrySet.java
@@ -91,7 +91,7 @@ class EntrySet<K, V> {
     // (not necessarily first in the array!)
     private static final int HEAD_NEXT_INDEX_SIZE = 1;
 
-    private static final int FIELDS = 3;  // # of fields in each item of entries array
+    private static final int FIELDS = 3;  // # of primitive fields in each item of entries array
 
     private static final Unsafe UNSAFE = UnsafeUtils.unsafe;
     final MemoryManager valuesMemoryManager;
@@ -190,7 +190,7 @@ class EntrySet<K, V> {
      * Converts external entry-index to internal array index.
      * <p>
      * We use the following terminology:
-     *  - intIdx for the integer's index inside the entries array (referred as a set of integers)
+     *  - longIdx for the long's index inside the entries array (referred as a set of longs)
      *  - entryIdx for the index of entry array (referred as a set of entries)
      *
      * @param entryIdx external entry-index
@@ -204,33 +204,14 @@ class EntrySet<K, V> {
     }
 
     /**
-     * getEntryArrayFieldInt gets the integer field of entry at specified offset for given
-     * start of the entry index in the entry array.
-     * The field is read atomically as it is a machine word, however the concurrency of the
-     * mostly updated value is not ensured as no memory fence is issued.
-     */
-    private long getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
-        return entries[intFieldIdx + offset.value]; // used for NEXT
-    }
-
-    /**
-     * getEntryArrayFieldLong atomically reads two integers field of the entries array.
-     * Should be used with OFFSET.VALUE_REFERENCE and OFFSET.KEY_REFERENCE
+     * getEntryArrayFieldLong atomically reads long field of the entries array.
+     * Could be used with any OFFSET.value
      */
     private long getEntryArrayFieldLong(int intStartFieldIdx, OFFSET offset) {
         long arrayOffset =
                 Unsafe.ARRAY_LONG_BASE_OFFSET + (intStartFieldIdx + offset.value) * Unsafe.ARRAY_LONG_INDEX_SCALE;
         assert arrayOffset % 8 == 0;
         return UNSAFE.getLong(entries, arrayOffset);
-    }
-
-    /**
-     * setEntryFieldInt sets the integer field of specified offset to 'value'
-     * for given integer index in the entry array
-     */
-    private void setEntryFieldInt(int intFieldIdx, OFFSET offset, int value) {
-        assert intFieldIdx + offset.value >= 0;
-        entries[intFieldIdx + offset.value] = value; // used for NEXT
     }
 
     private void setEntryFieldLong(int item, OFFSET offset, long value) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
             return 0;
         }
         return ((entryIdx - 1) * FIELDS) + HEAD_NEXT_INDEX_SIZE;
     }
 
     /**
+<<<<<<< HEAD
+     * getEntryArrayFieldLong atomically reads long field of the entries array.
+     * Could be used with any OFFSET.value
+=======
```
liran-funaro, line 210: Merge issue

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
      */
-    private int getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
+    private long getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
         return entries[intFieldIdx + offset.value]; // used for NEXT
     }
 
     /**
      * getEntryArrayFieldLong atomically reads two integers field of the entries array.
      * Should be used with OFFSET.VALUE_REFERENCE and OFFSET.KEY_REFERENCE
+>>>>>>> 7ea91e5b1433fe70d89c56c5d0ca6614b323f6ab
```
liran-funaro, line 223: Merge issue

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
      */
     private long getEntryArrayFieldLong(int intStartFieldIdx, OFFSET offset) {
-        assert offset == OFFSET.VALUE_REFERENCE || offset == OFFSET.KEY_REFERENCE;
         long arrayOffset =
-                Unsafe.ARRAY_INT_BASE_OFFSET + (intStartFieldIdx + offset.value) * Unsafe.ARRAY_INT_INDEX_SCALE;
+                Unsafe.ARRAY_LONG_BASE_OFFSET + (intStartFieldIdx + offset.value) * Unsafe.ARRAY_LONG_INDEX_SCALE;
         assert arrayOffset % 8 == 0;
         return UNSAFE.getLong(entries, arrayOffset);
```
liran-funaro, line 229: Why not reading directly from the array? Same comment for `setEntryFieldLong()`.
```java
return entries[intStartFieldIdx + offset.value];
```

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
+    private long getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
         return entries[intFieldIdx + offset.value]; // used for NEXT
     }
 
     /**
      * getEntryArrayFieldLong atomically reads two integers field of the entries array.
      * Should be used with OFFSET.VALUE_REFERENCE and OFFSET.KEY_REFERENCE
+>>>>>>> 7ea91e5b1433fe70d89c56c5d0ca6614b323f6ab
      */
     private long getEntryArrayFieldLong(int intStartFieldIdx, OFFSET offset) {
```
liran-funaro, line 225:
```suggestion
    private long getEntryArrayFieldLong(int longStartFieldIdx, OFFSET offset) {
```

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
-    /**
-     * setEntryFieldInt sets the integer field of specified offset to 'value'
-     * for given integer index in the entry array
-     */
-    private void setEntryFieldInt(int intFieldIdx, OFFSET offset, int value) {
-        assert intFieldIdx + offset.value >= 0;
-        entries[intFieldIdx + offset.value] = value; // used for NEXT
-    }
-
     private void setEntryFieldLong(int item, OFFSET offset, long value) {
```
liran-funaro, line 232:
```suggestion
    private void setEntryFieldLong(int longStartFieldIdx, OFFSET offset, long value) {
```

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -190,78 +190,61 @@ int getLastEntryIndex() {
...
+=======
      * getEntryArrayFieldInt gets the integer field of entry at specified offset for given
      * start of the entry index in the entry array.
      * The field is read atomically as it is a machine word, however the concurrency of the
      * mostly updated value is not ensured as no memory fence is issued.
      */
-    private int getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
+    private long getEntryArrayFieldInt(int intFieldIdx, OFFSET offset) {
         return entries[intFieldIdx + offset.value]; // used for NEXT
     }
```
liran-funaro, line 218: It is currently not in use. I suggest using it when an int field is required.
```suggestion
    private long getEntryArrayFieldInt(int longStartFieldIdx, OFFSET offset) {
        return entries[intFieldIdx + offset.value] & ((1L << 32) - 1); // used for NEXT
    }
```

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -309,7 +292,7 @@ int getNextEntryIndex(int ei) {
         if (ei == INVALID_ENTRY_INDEX) {
             return INVALID_ENTRY_INDEX;
         }
-        return getEntryArrayFieldInt(entryIdx2intIdx(ei), OFFSET.NEXT);
+        return (int) getEntryArrayFieldLong(entryIdx2LongIdx(ei), OFFSET.NEXT);
```
liran-funaro, line 295: I suggest using here (and below) `getEntryArrayFieldInt()`.
At least use a mask before converting to int:
```suggestion
        return (int) getEntryArrayFieldLong(entryIdx2LongIdx(ei), OFFSET.NEXT) & ((1L << 32) - 1);
```

core/src/main/java/com/yahoo/oak/EntrySet.java
```diff
@@ -593,7 +576,7 @@ boolean deleteValueFinish(ThreadContext ctx) {
 
         assert ctx.valueState == ValueState.DELETED_NOT_FINALIZED;
 
-        int indIdx = entryIdx2intIdx(ctx.entryIndex);
+        int indIdx = entryIdx2LongIdx(ctx.entryIndex);
```
liran-funaro, line 579:
```suggestion
        int longIdx = entryIdx2LongIdx(ctx.entryIndex);
```

liran-funaro (approved):
Looks good to me.

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: good point indeed :)

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: I do not think it worth to keep a function that is not in use. getEntryArrayFieldInt removed for now

core/src/main/java/com/yahoo/oak/EntrySet.java
sanastas: Int access/assignment is currently done via long to int casting. As other part of long is not used. This is good enough.

PR #120, opened 2020-10-22
Value utils refactor Part I

sanastas:

Multiple changes toward encapsulating Memory Management: (1) changing Slice hierarchy (2) eliminating header initiation from ValueUtils interface

### DIFF
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -40,7 +40,7 @@ class Block {
         int now = allocated.getAndAdd(size);
         if (now + size > this.capacity) {
             allocated.getAndAdd(-size);
-            throw new OakOutOfMemoryException("Block " + id + " out of memory");
+            throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
         s.associateBlockAllocation(id, now, size, buffer);
         return true;
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -40,10 +40,9 @@ class Block {
         int now = allocated.getAndAdd(size);
         if (now + size > this.capacity) {
             allocated.getAndAdd(-size);
-            throw new OakOutOfMemoryException();
+            throw new OakOutOfMemoryException("Block " + id + " out of memory");
         }
-        s.setBlockidAndOffsetAndLength(id, now, size);
-        s.setBuffer(buffer);
+        s.associateBlockAllocation(id, now, size, buffer);
         return true;
     }
 
--- a/core/src/main/java/com/yahoo/oak/Header.java
+++ b/core/src/main/java/com/yahoo/oak/Header.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak;
+
+import sun.misc.Unsafe;
+
+class Header {
+
+    /*
+    * Long Header: int version + int lock
+    * 0...  ...31 | 32...                  ...61| 62 63
+    *  version    |   lock: current_readers#    | lock state
+    *
+    * lock state: 0x00 - FREE, 0x01 - LOCKED, 0x10 - DELETED, 0x11 - MOVED
+    *
+    * The length (in integer size) of the data is also held just following the header (in long size)
+    * The length is set once upon header allocation and later can only be read. */
+
+    enum LockStates {
+        FREE(0), LOCKED(1), DELETED(2), MOVED(3);
+
+        public final int value;
+
+        LockStates(int value) {
+            this.value = value;
+        }
+    }
+
+    private static final int VERSION_SIZE = 4;
+    private static final int VERSION_OFFSET = 0;
+
+    private static final int LOCK_STATE_MASK = 0x3;
+    private static final int LOCK_STATE_SHIFT = 2;
+    private static final int LOCK_SIZE = 4;
+    private static final int LOCK_OFFSET = VERSION_SIZE;
+
+    private static final int LENGTH_OFFSET = LOCK_OFFSET+LOCK_SIZE;
+
+    private static Unsafe unsafe = UnsafeUtils.unsafe;
+
+
+    private static int getInt(Slice s, int headerOffsetInBytes) {
+        return unsafe.getInt(s.getMetadataAddress() + headerOffsetInBytes);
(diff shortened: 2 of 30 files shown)


core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -40,10 +40,9 @@ boolean allocate(Slice s, int size) {
         int now = allocated.getAndAdd(size);
         if (now + size > this.capacity) {
             allocated.getAndAdd(-size);
-            throw new OakOutOfMemoryException();
+            throw new OakOutOfMemoryException("Block " + id + " out of memory");
```
liran-funaro, line 43: nit
```suggestion
            throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
```

core/src/main/java/com/yahoo/oak/Header.java
```diff
@@ -0,0 +1,94 @@
...
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak;
+
+import sun.misc.Unsafe;
+
+class Header {
```
liran-funaro, line 11: nit: Shouldn't it be `NativeMemoryManagerHeader`? Or something like that?

core/src/main/java/com/yahoo/oak/Header.java
```diff
@@ -0,0 +1,94 @@
...
+        LockStates(int value) {
+            this.value = value;
+        }
+    }
+
+    private static final int VERSION_SIZE = 4;
+    private static final int VERSION_OFFSET = 0;
+
+    private static final int LOCK_STATE_MASK = 0x3;
+    private static final int LOCK_STATE_SHIFT = 2;
```
liran-funaro, line 37: `LOCK_STATE_MASK` and `LOCK_STATE_SHIFT` are unused.

core/src/main/java/com/yahoo/oak/MemoryManager.java
```diff
@@ -23,13 +23,16 @@
     long allocated();
 
     /**
-     * This method allocates memory out of the off-heap, i.e., the ByteBuffer inside of {@code s} is pointing to the
-     * off-heap. The blockID, is an internal reference to which block the ByteBuffer points, allowing the functions
-     * {@code getSliceFromBlockID} and {@code getByteBufferFromBlockID} to reconstruct the same ByteBuffer.
-     *  @param s        - an allocation object to update with the new allocation
+     * This method allocates memory out of the off-heap, i.e. off-heap cut of the given size in bytes.
```
liran-funaro, line 26: "cut" is a new phrase. What does it mean?

core/src/main/java/com/yahoo/oak/NativeMemoryManager.java
```diff
@@ -106,21 +108,52 @@ public boolean isReferenceDeleted(long reference) {
...
     public long allocated() {
         return allocator.allocated();
     }
 
+    // Native memory manager requires metadata header to be placed before the user data written off-heap
+    // therefore the bigger than requested size is allocated
     @Override
-    public void allocate(Slice s, int size) {
-        boolean allocated = allocator.allocate(s, size);
+    public void allocate(Slice s, int size, boolean existing) {
```
liran-funaro, line 139: I have some reservations that the flag that hints about replacing existing value, implicitly sets the lock state.
This should be at least be addressed in the documentation.
If you have in mind a design that avoids that flag it is even better, but I guess this can wait for future PRs.

core/src/main/java/com/yahoo/oak/NoFreeMemoryManager.java
```diff
@@ -70,7 +71,7 @@ public boolean decodeReference(Slice s, long reference) {
             // key's slice is updated upon reference decoding.
             // Slice is not invalidated between next iterator steps and all the rest information
             // in slice remains the same.
-            s.setOffsetAndLength(rcd.getSecond(reference), rcd.getThird(reference));
+            s.setIterationOnSameBlock(rcd.getSecond(reference), rcd.getThird(reference));
```
liran-funaro, line 74: `setIterationOnSameBlock()` is unclear. Maybe `updateOnSameBlock()`?

liran-funaro (approved):
LGTM

core/src/main/java/com/yahoo/oak/Header.java
sanastas: I do not know yet, as I think Header's design is going to be changed a bit in the future. However, I changed the name to the one that you suggested at least for now.

core/src/main/java/com/yahoo/oak/Header.java
sanastas: removed

core/src/main/java/com/yahoo/oak/MemoryManager.java
sanastas: yes, I have the feeling that Slice is an object that doesn't exactly explains the physical (virtual) memory addresses being references. Like, Slice can be duplicated, not associated, associated, deleted as an object, etc. while the memory addresses it is representing are still the same, and do not go through the same transformation.

Anyway, I feel that the memory itself and the Slice object need to be addresses with different words. "Off-heap cut" (and also Slice) are explained at the beginning of MemoryManager interface file.

core/src/main/java/com/yahoo/oak/NativeMemoryManager.java
sanastas: The "existing" boolean flag parameter is explained in interface method in Memory Manager interface file. However, indeed doesn't explain that this will cause the lock to be treated differently. This is because it is internal implementation and after all (after allocation) the lock is released, this way or the other... I didn't plan to implement it differently, but you are welcome to suggest when we will discuss the  Memory Manager interface for putting it public.

sanastas: If you want we can discuss it further

core/src/main/java/com/yahoo/oak/NoFreeMemoryManager.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/NativeMemoryManager.java
liran-funaro: I don't think this requires a discussion. I only made a suggestion to add a comment to this implementation explaining the lock thing.

sanastas: I have added the comment. Thanks

PR #122, opened 2020-12-17
Final variation of less comparissons per stop bound checks

sanastas:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/Chunk.java
+++ b/core/src/main/java/com/yahoo/oak/Chunk.java
@@ -802,7 +802,7 @@ class Chunk<K, V> {
         private final K from;
         private boolean inclusive;
 
-        static final int SKIP_ENTRIES_FOR_BIGGER_STACK = 1; // 1 is the lowest possible value
+        private final int skipEntriesForBiggerStack = (maxItems/10); // 1 is the lowest possible value
 
         DescendingIter(ThreadContext ctx) {
             KeyBuffer tempKeyBuff = ctx.tempKey;
@@ -913,10 +913,10 @@ class Chunk<K, V> {
             } else if (anchor == 1) { // cannot get below the first index
                 anchor = entrySet.getHeadNextIndex();
             } else {
-                if ((anchor - SKIP_ENTRIES_FOR_BIGGER_STACK) > 1) {
+                if ((anchor - skipEntriesForBiggerStack) > 1) {
                     // try to skip more then one backward step at a time
                     // if it shows better performance
-                    anchor -= SKIP_ENTRIES_FOR_BIGGER_STACK;
+                    anchor -= skipEntriesForBiggerStack;
                 } else {
                     anchor -= 1;
                 }
--- a/core/src/main/java/com/yahoo/oak/Chunk.java
+++ b/core/src/main/java/com/yahoo/oak/Chunk.java
@@ -720,32 +720,39 @@ class Chunk<K, V> {
     /********************************************************************************************/
     /*--------------------------------- Iterators Constructors ---------------------------------*/
 
-    // Ascending iterator from the beginning of the chunk.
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -788,25 +860,62 @@ public boolean hasNext() {
...
-            advance();
+            advance(ctx);
             return toReturn;
         }
-    }
 
-    class DescendingIter implements ChunkIter {
+        @Override
+        public boolean isBoundCheckNeeded() {
+            return needBoundCheckDynamic;
```
liran-funaro, line 869: Here it could be:
```suggestion
            return boundCheck != NO_BOUNDRY_CHECK;
```

core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -788,25 +860,62 @@ public boolean hasNext() {
...
+    }
+
+    class DescendingIter extends ChunkIter {
         private int anchor;
         private int prevAnchor;
         private final IntStack stack;
         private final K from;
-        private boolean inclusive;
+        private boolean fromInclusive;
+        private final int skipEntriesForBiggerStack = (maxItems/10); // 1 is the lowest possible value
```
liran-funaro, line 914: Correctness: if `maxItems` is less than 10, `skipEntriesForBiggerStack` will be zero.

```suggestion
        private final int skipEntriesForBiggerStack = Math.max(1, maxItems / 10); // 1 is the lowest possible value
```

core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -913,14 +1030,20 @@ private void findNewAnchor() {
             } else if (anchor == 1) { // cannot get below the first index
                 anchor = entrySet.getHeadNextIndex();
             } else {
-                if ((anchor - SKIP_ENTRIES_FOR_BIGGER_STACK) > 1) {
+                if ((anchor - skipEntriesForBiggerStack) > 1) {
                     // try to skip more then one backward step at a time
                     // if it shows better performance
-                    anchor -= SKIP_ENTRIES_FOR_BIGGER_STACK;
+                    anchor -= skipEntriesForBiggerStack;
```
liran-funaro, line 1036: Performance: note that this might cause us to miss opportunities to avoid boundary checks.

liran-funaro (commented):
LGTM. Most of my comments are about clarity and design.
However, I found two possible correctness issues and one possible performance issue.

core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -717,67 +717,139 @@ void release() {
...
+        abstract int next(ThreadContext ctx);
 
-        private int next;
+        abstract boolean isBoundCheckNeeded();
 
-        AscendingIter() {
+        abstract boolean afterStop(OakScopedReadBuffer boundKey);
+
+        protected void setStopCheckThreshold(
+            ThreadContext ctx, K to, boolean toInclusive, OakScopedReadBuffer boundKey) {
```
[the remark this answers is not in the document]
sanastas: I think that this brings the symmetry between ascending and descending scan. Descending scan has the internal (fast accessed) minKey that the scan can check upon. Otherwise for descending vs ascending scan we need different traversing logic that I would like to avoid. For now I do not apply all the suggestions related to this idea. We can discuss it during the code review meeting.

core/src/main/java/com/yahoo/oak/Chunk.java
sanastas: If I understand it correctly you are worried that we will check also the keys before midIdx's key. The only alternative that I see is to check for midIdx on every next step (as in ascending scan) and this I would like to avoid. We can discuss it during the code review meeting.

liran-funaro (approved):
:+1:

PR #124, opened 2021-01-19
More benchmarks v01

sanastas:

Small fixes:
Sequential put benchmark, on-heap and off-heap memory limits, decreased default block size, same allocator memory limits for all benchmark iterations


liran-funaro (approved):
LGTM

PR #126, opened 2021-01-26
Moving all memory management related out of ValueUtils

sanastas:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/SyncRecycleMMHeader.java
+++ b/core/src/main/java/com/yahoo/oak/SyncRecycleMMHeader.java
@@ -36,57 +36,181 @@ class SyncRecycleMMHeader {
     private static final int LOCK_SIZE = 4;
     private static final int LOCK_OFFSET = VERSION_SIZE;
 
-    private static final int LENGTH_OFFSET = LOCK_OFFSET+LOCK_SIZE;
+    private static final int LOCK_STATE_MASK = 0x3;
+    private static final int LOCK_STATE_SHIFT = 2;
+
+    private static final int LENGTH_OFFSET = VERSION_SIZE+LOCK_SIZE;
 
     private static Unsafe unsafe = UnsafeUtils.unsafe;
 
 
-    private static int getInt(Slice s, int headerOffsetInBytes) {
-        return unsafe.getInt(s.getMetadataAddress() + headerOffsetInBytes);
+    private static int getInt(long headerAddress, int intOffsetInBytes) {
+        return unsafe.getInt(headerAddress + intOffsetInBytes);
     }
 
-    private static void putInt(Slice s, int headerOffsetInBytes, int value) {
-        unsafe.putInt(s.getMetadataAddress() + headerOffsetInBytes, value);
+    private static void putInt(long headerAddress, int intOffsetInBytes, int value) {
+        unsafe.putInt(headerAddress + intOffsetInBytes, value);
     }
 
-    void setVersion(Slice s) {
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/yahoo/oak/SyncRecycleMemoryManager.java
```diff
@@ -153,17 +153,17 @@ public void allocate(Slice s, int size, boolean existing) {
...
-        assert HEADER.getVersion(s) == allocationVersion;
+        assert HEADER.getOffHeapVersion(s.getMetadataAddress()) == allocationVersion;
     }
 
     @Override
     public void release(Slice s) {
         if (s.length == Slice.UNDEFINED_LENGTH_OR_OFFSET) {
-            ValueUtilsImpl.setLengthFromOffHeap(s);
+            s.getLength(); // this will set the length from off-heap header, if needed
         }
```
liran-funaro, line 167: 1. `s.getLength()` already checks `if (s.length == Slice.UNDEFINED_LENGTH_OR_OFFSET)`, so no need to check it here as well.
2. It is weird to find this piece of code here because the reader might not know about `Slice`'s lazy evaluation mechanism. Maybe add a method to `Slice`: `prefetchDataFromHeader()` that will fetch all data that is lazily stored on the header. This will allow you to put more data on the header in the future if needed.
Here you will call `s.prefetchDataFromHeader()` which will explicitly let the reader know of `Slice`'s lazy evaluation mechanism.

core/src/main/java/com/yahoo/oak/ValueUtilsImpl.java
```diff
@@ -60,47 +22,49 @@ public static void setLengthFromOffHeap(Slice s){
...
             return result.withValue(transformation);
         } finally {
-            unlockRead(value.s);
+            value.s.unlockRead();
         }
     }
 
     @Override
     public <V> ValueResult put(Chunk<?, V> chunk, ThreadContext ctx, V newVal, OakSerializer<V> serializer,
                                MemoryManager memoryManager, InternalOakMap internalOakMap) {
```
liran-funaro, line 31: `MemoryManager memoryManager,` can be removed here as well.

liran-funaro (approved):
This PR is a great step toward exporting the off-heap memory-managers as a package.

core/src/main/java/com/yahoo/oak/SyncRecycleMMHeader.java
```diff
@@ -36,57 +36,181 @@
...
-        return getInt(s, LENGTH_OFFSET);
+    private void initHeader(long headerAddress, LockStates state, int dataLength, int version) {
+        setOffHeapVersion(headerAddress, version);
+        setLockState(headerAddress, state);
+        setDataLength(headerAddress, dataLength);
     }
 
-    void setLength(Slice s, int length) {
-        putInt(s, LENGTH_OFFSET, length);
+    void initHeader(long headerAddress, int dataLength, int version) {
```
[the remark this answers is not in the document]
sanastas: renamed

core/src/main/java/com/yahoo/oak/SyncRecycleMemoryManager.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/ValueUtilsImpl.java
sanastas: removed

PR #129, opened 2021-02-10
Adding increasing keys ingestion benchmark

sanastas:

Adding increasing keys ingestion benchmark

### DIFF
--- a/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
+++ b/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
@@ -91,7 +91,7 @@ public class ThreadLoopOak implements Runnable {
 
         boolean change = Parameters.change;
         int scanLength = 10000;
-        int i = 100;
+        int i = 150;
 
         MyBuffer key = new MyBuffer(Parameters.keySize);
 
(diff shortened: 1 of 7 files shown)


benchmarks/synchrobench/src/test/java/com/yahoo/oak/OakMyBufferTest.java
```diff
@@ -48,6 +48,21 @@ public void testPut() {
...
+        for (long i = 1000000; i > 0; ) {
+            v++;
+            MyBuffer key = new MyBuffer(Parameters.keySize);
+            key.buffer.putInt(0, v);
+            MyBuffer val = new MyBuffer(Parameters.valSize);
+            val.buffer.putInt(0, v);
+            oakBench.putOak(key, val);
+            i--;
+        }
+    }
```
liran-funaro, line 64: Why do you iterate backward over `i` when you don't even use it?
It would be simpler to iterate forward over `v`.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
```diff
@@ -95,7 +95,8 @@ public void run() {
 
         MyBuffer key = new MyBuffer(Parameters.keySize);
 
-        Integer newInt = -1;
+        Integer newInt = (Parameters.keyDistribution == Parameters.KeyDist.RANDOM) ? -1 : Parameters.size;
```
liran-funaro, line 98: The conditional assignment is redundant. You can simply assign `Parameters.size` because in the case of random distribution, it will override anyway.
Please add some comment here explaining why we assign `Parameters.size` and not simply zero.

liran-funaro (approved):
LGTM

benchmarks/synchrobench/src/test/java/com/yahoo/oak/OakMyBufferTest.java
sanastas: fixed

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/ThreadLoopOak.java
sanastas: fixed

PR #147, opened 2021-03-16
buffer overlap and fragmentation due to race condition in Block.allocate

thiyanesh:
Fail eagerly on allocated size comparison to prevent overlap and fragmentation due to race condition.

### DIFF
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -22,7 +22,7 @@ class Block {
 
     Block(long capacity) {
         assert capacity > 0;
-        assert capacity <= Integer.MAX_VALUE; // This is exactly 2GB
+        assert capacity <= Integer.MAX_VALUE; // This is exactly 2GiB
         this.capacity = (int) capacity;
         this.id = NativeMemoryAllocator.INVALID_BLOCK_ID;
         // Pay attention in allocateDirect the data is *zero'd out*
@@ -38,21 +38,20 @@ class Block {
     // The returned buffer doesn't have all zero bytes.
     boolean allocate(Slice s, int size) {
         assert size >= 0; // avoid negative size
-        int now = 0;
-        boolean invalid = false;
-        synchronized (allocated) { // this is suboptimal, will optimize
-            // long to manage integer overflow or use AtomicLong for single getAndAdd for valid scenario
-            long next = 0L + allocated.get() + size;
-            if (next > this.capacity) {
-                invalid = true; // use flag instead of exception to reduce time spent inside synchronized block
-            } else {
-                now = allocated.getAndAdd(size);
-            }
+        long next = 0;
+        boolean updated = false;
+
+        // update if capacity can accommodate requested size. use long to avoid integer overflow
+        while (!updated && (next = ((long) allocated.get()) + size) <= this.capacity) {
+            // compareAndSet manages value change as atomic transaction
+            updated = allocated.compareAndSet((int) (next - size), (int) next);
         }
-        if (invalid) {
+
+        if (!updated) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
-        s.associateBlockAllocation(id, now, size, buffer);
+
+        s.associateBlockAllocation(id, (int) (next - size), size, buffer);
         return true;
     }
 
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -37,9 +37,19 @@ class Block {
     // Block manages its linear allocation. Thread safe.
     // The returned buffer doesn't have all zero bytes.
     boolean allocate(Slice s, int size) {
-        int now = allocated.getAndAdd(size);
-        if (now + size > this.capacity) {
-            allocated.getAndAdd(-size);
+        assert size >= 0; // avoid negative size
+        int now = 0;
+        boolean invalid = false;
+        synchronized (allocated) { // this is suboptimal, will optimize
+            // long to manage integer overflow or use AtomicLong for single getAndAdd for valid scenario
+            long next = 0L + allocated.get() + size;
+            if (next > this.capacity) {
+                invalid = true; // use flag instead of exception to reduce time spent inside synchronized block
+            } else {
+                now = allocated.getAndAdd(size);
+            }
+        }
+        if (invalid) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
         s.associateBlockAllocation(id, now, size, buffer);
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -10,14 +10,14 @@ import sun.misc.Cleaner;
 
 import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 class Block {
 
     private final ByteBuffer buffer;
 
     private final int capacity;
-    private final AtomicInteger allocated = new AtomicInteger(0);
+    private final AtomicLong allocated = new AtomicLong(0);
     private int id; // placeholder might need to be set in the future
 
     Block(long capacity) {
@@ -36,21 +36,16 @@ class Block {
 
     // Block manages its linear allocation. Thread safe.
     // The returned buffer doesn't have all zero bytes.
-    boolean allocate(Slice s, int size) {
+    boolean allocate(Slice s, final int size) {
         assert size > 0;
-
-        int current;
-        long next;
-        do {
-            current = allocated.get();
-            next = (long) current + size;  // use long to avoid integer overflow
-        } while (next <= this.capacity && !allocated.compareAndSet(current, (int) next));
-
-        if (next > this.capacity) {
+        long now = allocated.get();
+        if (now + size <= this.capacity) { // check is only an optimization
+            now = allocated.getAndAdd(size);
+        }
+        if (now + size > this.capacity) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
-
-        s.associateBlockAllocation(id, current, size, buffer);
+        s.associateBlockAllocation(id, (int) now, size, buffer);
         return true;
     }
 
@@ -60,8 +55,10 @@ class Block {
         allocated.set(0);
     }
 
-    // return how many bytes are actually allocated for this block only, thread safe
-    long allocated() {
+    // return upperbound of bytes actually allocated for this block only, thread safe
+    // the returned value can be greater than the actual bytes allocated
+    // do not use this for exact comparison or any precise computation
+    long allocatedWithPossibleDelta() {
         return allocated.get();
     }
 
--- a/core/src/main/java/com/yahoo/oak/Block.java
+++ b/core/src/main/java/com/yahoo/oak/Block.java
@@ -37,21 +37,20 @@ class Block {
     // Block manages its linear allocation. Thread safe.
     // The returned buffer doesn't have all zero bytes.
     boolean allocate(Slice s, int size) {
-        assert size >= 0; // avoid negative size
-        long next = 0;
-        boolean updated = false;
+        assert size > 0;
 
-        // update if capacity can accommodate requested size. use long to avoid integer overflow
-        while (!updated && (next = ((long) allocated.get()) + size) <= this.capacity) {
-            // compareAndSet manages value change as atomic transaction
-            updated = allocated.compareAndSet((int) (next - size), (int) next);
-        }
+        int current;
+        long next;
+        do {
+            current = allocated.get();
+            next = (long) current + size;  // use long to avoid integer overflow
+        } while (next <= this.capacity && !allocated.compareAndSet(current, (int) next));
 
-        if (!updated) {
+        if (next > this.capacity) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
 
-        s.associateBlockAllocation(id, (int) (next - size), size, buffer);
+        s.associateBlockAllocation(id, current, size, buffer);
         return true;
     }
 
--- a/core/src/main/java/com/yahoo/oak/BlocksPool.java
+++ b/core/src/main/java/com/yahoo/oak/BlocksPool.java
@@ -7,6 +7,7 @@
 package com.yahoo.oak;
 
 import java.io.Closeable;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 /**
@@ -18,6 +19,9 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 final class BlocksPool implements BlocksProvider, Closeable {
 
     private static BlocksPool instance = null;
+    private static final String SINGLETON_KEY = "SINGLETON";
+    // keep the reference as ConcurrentHashMap to prevent assignment of non-confirming computeIf___ map instances
+    private static final ConcurrentHashMap<String, BlocksPool> SINGLETON_MAP = new ConcurrentHashMap<>(1);
     private final ConcurrentLinkedQueue<Block> blocks = new ConcurrentLinkedQueue<>();
 
     // TODO change the following constants to be configurable
@@ -78,23 +82,23 @@ final class BlocksPool implements BlocksProvider, Closeable {
      */
     static BlocksPool getInstance() {
         if (instance == null) {
-            synchronized (BlocksPool.class) { // can be easily changed to lock-free
-                if (instance == null) {
-                    instance = new BlocksPool();
-                }
-            }
+            instance = SINGLETON_MAP.computeIfAbsent(SINGLETON_KEY, k -> new BlocksPool());
         }
         return instance;
     }
 
     // used only in OakNativeMemoryAllocatorTest.java
-    static void setBlockSize(int blockSize) {
-        synchronized (BlocksPool.class) { // can be easily changed to lock-free
-            if (instance != null) {
-                instance.close();
+    // do not use in actual production code
+    static void setBlockSize(final int blockSize) {
+        instance = SINGLETON_MAP.computeIfAbsent(SINGLETON_KEY, (k) -> new BlocksPool(blockSize));
+        instance = SINGLETON_MAP.computeIfPresent(SINGLETON_KEY, (k, v) -> {
+            if (v.blockSize() != blockSize) {
+                v.close();
+                return new BlocksPool(blockSize);
+            } else {
+                return v;
             }
-            instance = new BlocksPool(blockSize);
-        }
+        });
     }
 
     /**
@@ -102,13 +106,9 @@ final class BlocksPool implements BlocksProvider, Closeable {
      * @param preferredBlockSizeBytes the preferred block size
      * @return true if the preferred block size matches the current instance
      */
-    static boolean preferBlockSize(int preferredBlockSizeBytes) {
+    static boolean preferBlockSize(final int preferredBlockSizeBytes) {
         if (instance == null) {
-            synchronized (BlocksPool.class) { // can be easily changed to lock-free
-                if (instance == null) {
-                    instance = new BlocksPool(preferredBlockSizeBytes);
-                }
-            }
+            instance = SINGLETON_MAP.computeIfAbsent(SINGLETON_KEY, (k) -> new BlocksPool(preferredBlockSizeBytes));
         }
 
         return instance.blockSizeBytes == preferredBlockSizeBytes;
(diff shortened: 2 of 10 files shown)


core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -37,9 +37,19 @@ void setID(int id) {
...
+            if (next > this.capacity) {
+                invalid = true; // use flag instead of exception to reduce time spent inside synchronized block
+            } else {
+                now = allocated.getAndAdd(size);
+            }
+        }
+        if (invalid) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
         s.associateBlockAllocation(id, now, size, buffer);
```
thiyanesh, line 55: lets consider the following scenario:

1. capacity = 10
2. allocated = 0
3. 4 threads (R1, R2, R3, R4)
4. block id = B1
5. Different execution times of 4 threads are tracked as T1, T2, T3, T4
6. Assume there is only one core available for execution

## Time sequenced operations:
### T1.1: R1.allocate(size = 11)
1. now = 0
2. allocated = 11 (0 + 11)
3. if (0 + 11 > 10) - true (invalid)
4. context switch

### T2.1: R2.allocate(size = 1)
1. now = 11
2. allocated = 12 (11 + 1)
3. if (11 + 1 > 10) - true (invalid, though this is an ambiguous state. still its fine to proceed with the analysis)
4. context switch

### T1.2 : R1.allocate(size = 11): resumes from T1.1
1. now = 0
2. allocated = 12
3. allocated.add(-11)
4. allocated = 1
5. exception due to true in if check in T1.1
6. completed

### T3.1 : R3.allocate(size = 2)
1. now = 1
2. allocated = 3 (1 + 2)
3. if (3 > 10) - false (valid)
4. s.associateBlockAllocation(B1, offset = 1 , size = 2 , buffer) - (uses: 1 and 2)
5. completed

### T2.2: R2.allocate(size = 1): resumes from T2.1
1. now = 11
2. allocated = 3
3. allocated.add(-1) (due to T2.1 check of 11 + 1 > 10)
4. allocated = 2
5. exception due to true in if check in T2.1
6. completed
7. now offset = 0 is in an ambiguous state

### T4.1: R4.allocate(size = 1)
1. now = 2 (from allocated)
2. allocated = 3 (2 + 1)
3. if (3 > 10) - false (valid)
4. s.associateBlockAllocation(B1, offset = 2 , size = 1 , buffer) - (uses: 2)
completed
5. now offset = 2 is clearly in violated state with both R3.allocate(size = 2) and R4.allocate(size = 1) sharing/overlapping at offset 2

can you clarify whether this is a valid scenario?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -37,9 +37,19 @@ void setID(int id) {
     // Block manages its linear allocation. Thread safe.
     // The returned buffer doesn't have all zero bytes.
     boolean allocate(Slice s, int size) {
-        int now = allocated.getAndAdd(size);
-        if (now + size > this.capacity) {
-            allocated.getAndAdd(-size);
+        assert size >= 0; // avoid negative size
```
thiyanesh, line 40: Alternatively using a StampedLock to bound the retries:
```java
    boolean allocate(Slice s, int size) {
        assert size >= 0; // avoid negative size
        long next;
        int retries = 0;
        long readStamp = 0;

        while ((next = 0L + allocated.get() + size) <= this.capacity
                && (readStamp = stampedLock.readLock()) != 0
                && !allocated.compareAndSet((int) (next - size), (int) next)) {

            // release read lock and its stamp
            stampedLock.unlockRead(readStamp);
            readStamp = 0;

            // switch to exclusive mode as collaborative retries seems costlier beyond a limit
            if (++retries > SPIN_COUNT_UPPER_LIMIT) {
                long writeStamp = stampedLock.writeLock();
                next = 0L + allocated.get() + size;
                if (next <= this.capacity) {
                    allocated.getAndAdd(size);
                }
                stampedLock.unlockWrite(writeStamp);
                break;
            }
        }

        if (readStamp != 0) {
            stampedLock.unlockRead(readStamp);
        }

        if (next > this.capacity) {
        }

        s.associateBlockAllocation(id, (int) (next - size), size, buffer);
        return true;
    }
```

Or Can just spin
```java
    boolean allocate(Slice s, int size) {
        assert size >= 0; // avoid negative size
        long next;
        int retries = 0;
        while ((next = 0L + allocated.get() + size) <= this.capacity
            && !allocated.compareAndSet((int) (next - size), (int) next)) {
            retries++; // no-op needed to satisfy codestyle
        }
        if (next > this.capacity) {
        }
        s.associateBlockAllocation(id, (int) (next - size), size, buffer);
        return true;
    }
```

Eshcar:
Thanks @thiyanesh for this PR

I recommend breaking it into several PRs
It can be easily be broken into 3 PRs
1. allocate race condition
2. changes to blocks pool - better practice in allocating a singleton
3. adding final to potentially improve performance and avoid changes to parameters

The main issue here is the race condition, that is the result on non-atomic (non-transactional) increment and decrement operations on `allocated ` which captures the next location to allocate in a buffer

If indeed the allocate method is called not in exclusive mode there is the potential for the race you describe
But I am not sure a (costly) synchronize is necessary, or other complicated solutions
The only reason this race is created is due to the non-monotonic behavior of `allocated `.  But it is only decreased after exceeding the threshold.
In a normal workload we can expect this to happen only once when the block is almost full.
So a possible solution would be to have two monotonic counters. The allocate method would be as follows 
- Increase the first counter. 
- If it is below the threshold then increase the second counter and treat it same as `allocated`. It is guaranteed `allocated` does not exceed the threshold. 
- if the first counter exceeds the threshold then throw an exception
no reduction of any counter. no race condition. 
The only caveat is an unused space at the end of a block. but to me it seems reasonable that once an exception is thrown no additional allocation is allowed from the block

core/src/main/java/com/yahoo/oak/common/OakCommonBuildersFactory.java
```diff
@@ -26,6 +26,7 @@
 
     public static final OakComparator<Integer> DEFAULT_INT_COMPARATOR = new OakIntComparator();
     public static final OakSerializer<Integer> DEFAULT_INT_SERIALIZER = new OakIntSerializer();
+    private static final String EMPTY_STRING = "";
```
Eshcar, line 29: separate PR (?) beautify code

core/src/main/java/com/yahoo/oak/common/intbuffer/OakIntBufferComparator.java
```diff
@@ -39,13 +39,15 @@ public int compareKeyAndSerializedKey(ByteBuffer key, OakScopedReadBuffer serial
         return compare(key, 0, size, unsafeKey.getByteBuffer(), unsafeKey.getOffset(), size);
     }
 
-    public static int compare(ByteBuffer buff1, int pos1, int size1, ByteBuffer buff2, int pos2, int size2) {
```
Eshcar, line 42: separate PR (?) - adding final in multiple locations

Eshcar (changes requested):
see my comment below

thiyanesh:
Thank you for the detailed feedback. I will update the PR with `Block.allocate` change alone. also, will try to solve in using only one existing atomic variable.

core/src/main/java/com/yahoo/oak/common/intbuffer/OakIntBufferComparator.java
thiyanesh: ok. will add a separate PR for the formatting/simple changes.

core/src/main/java/com/yahoo/oak/common/OakCommonBuildersFactory.java
thiyanesh: Sure 👍

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -18,6 +19,9 @@
 final class BlocksPool implements BlocksProvider, Closeable {
 
     private static BlocksPool instance = null;
+    private static final String SINGLETON_KEY = "SINGLETON";
+    // keep the reference as ConcurrentHashMap to prevent assignment of non-confirming computeIf___ map instances
+    private static final ConcurrentHashMap<String, BlocksPool> SINGLETON_MAP = new ConcurrentHashMap<>(1);
```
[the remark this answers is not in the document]
thiyanesh: Sure, 👍

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -37,9 +37,19 @@ void setID(int id) {
...
+        int now = 0;
+        boolean invalid = false;
+        synchronized (allocated) { // this is suboptimal, will optimize
```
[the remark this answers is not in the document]
thiyanesh: Sure. removed the synchronized block and used spin. Do share your feedback, if you still prefer 2 counter approach.

thiyanesh:
@Eshcar ,
If this PR is fine, please squash merge to avoid unnecessary commit history with other changes.
Or do you recommend to close this PR and create a new PR with this change?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -37,12 +37,21 @@ void setID(int id) {
...
-            allocated.getAndAdd(-size);
+        assert size >= 0; // avoid negative size
+        long next = 0;
+        boolean updated = false;
+
+        // update if capacity can accommodate requested size. use long to avoid integer overflow
+        while (!updated && (next = ((long) allocated.get()) + size) <= this.capacity) {
+            // compareAndSet manages value change as atomic transaction
+            updated = allocated.compareAndSet((int) (next - size), (int) next);
+        }
```
liran-funaro, line 48: Could you please modify this to use the `do {} while ()` syntax instead as it would be clearer?
Involving an assignment inside a while statement is prone to errors and considered hard to follow.

```java
do {
    cur = allocated.get();
    next = (long) cur + size;
} while (next <= capacity && !allocated.compareAndSet(cur, (int) next));
```

To validate the capacity later, you can simply use `if (next > capacity) throw ...`.

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -37,12 +37,21 @@ void setID(int id) {
     // Block manages its linear allocation. Thread safe.
     // The returned buffer doesn't have all zero bytes.
     boolean allocate(Slice s, int size) {
-        int now = allocated.getAndAdd(size);
-        if (now + size > this.capacity) {
-            allocated.getAndAdd(-size);
+        assert size >= 0; // avoid negative size
```
liran-funaro, line 40: We probably want to avoid `size==0` as well.

liran-funaro (changes requested):
Great catch on this bug!
Please address my comments below.

Note for other reviewers: this change won't have an effect on the performance because this is exactly how `getAndAdd()` is implemented in Java.

sanastas:
Hi @thiyanesh !

I am so happy you opened this PR and found that real problem! Thank you!
I am Anastasia and I am the lead of Oak project, nice to e-meet you!

The scenario you present is real and thus should be definitely fixed. You have a great eye on concurrency problems and it would be pleasure to continue collaborating!

I less like any locks or synchronized being added. Also, I less like the CompareAndSwap loop, this is because Fetch-And-Add [still has better performance](http://ashkrit.blogspot.com/2014/02/atomicinteger-java-7-vs-java-8.html). Previously `AtomicInteger.getAndAdd()` (redirected to `unsafe.getAndAddInt()`) was implemented with Fetch-and-Add and now I indeed see it turns to CompareAndSwap-loop in Java8. However, the implementations can be different and I wouldn't straightforward use `compareAndSet()`.

What I would suggest as a solution is first of all to remove the decreasing of `allocated` once the request is above capacity and to throw the exception. Looking on `Block.allocate()` usage (in NativeMemoryAllocator) we can see that the exception is just caught and if the request is not problematic (i.e. above the block size in general), allocation request is redirected to next block.

In order not to increase `allocated` too much, I would add a check 
`if (allocated.get() + size > this.capacity)`
just before the `allocated.getAndAdd(size)` to decrease the probability of unnecessary increase.

What do you think  @thiyanesh ? I would very much like to hear your thoughts! If you really want to progress with the CompareAndSet loop this is also a possibility. Also, leave this PR for dealing with this problem only. Please go ahead and open other PRs with all other improvements you have suggested before. Looking forward!

Thanks,
Anastasia

core/src/main/java/com/yahoo/oak/Block.java
thiyanesh: @liran-funaro, Sure, your feedback to use `do while` will make the code more readable. Will update to use `do while`. Thank you.

core/src/main/java/com/yahoo/oak/Block.java
thiyanesh: @liran-funaro Sure, will update the check as `assert size > 0`. I initially added `size >= 0`(included 0) to be backward compatible with any existing scenario using `0` size. 
Logically, users of this method should not pass `0` too, so will update check.
👍

thiyanesh:
@sanastas, Thank you for the detailed feedback.

_`What I would suggest as a solution is first of all to remove the decreasing of allocated once the request is above capacity and to throw the exception. Looking on Block.allocate() usage (in NativeMemoryAllocator) we can see that the exception is just caught and if the request is not problematic (i.e. above the block size in general), allocation request is redirected to next block.`_
True, this can be done if the following scenario is acceptable 
1. T1 requests bigger size
2. T2 and T2 requests valid size (combined size will also be valid)
3. Depending on the order of context switches in concurrently executing T1, T2, T3, the complete block offset(corresponding `bytebuffer`) might be left unused unless the whole block is released later.

If there will be more contentions and its ok to tradeoff space to achieve better time, then can use the approach you have suggested. (throw exception on the first overflow using 2 counters to simulate a completion and throw exceptions)
If there will be less such contentions, compareAndSet might be fine as there will be minimal loops.

_`In order not to increase allocated too much, I would add a check
if (allocated.get() + size > this.capacity)
just before the allocated.getAndAdd(size) to decrease the probability of unnecessary increase.`_
As per my understanding, this would not guarantee the operation to be atomic.
Further, since the `allocated` is maintained as `AtomicInteger` and not as `AtomicLong`, there can different side effects of overflow(negative) which can result in inconsistencies in `    long allocated() {
        return allocated.get();
    }
` response.

Please do share your comment about this(or if i misunderstood your comment).

sanastas:
@thiyanesh , good to hear from you again!

> True, this can be done if the following scenario is acceptable

>    T1 requests bigger size
>    T2 and T2 requests valid size (combined size will also be valid)
>    Depending on the order of context switches in concurrently executing T1, T2, T3, the complete block offset (corresponding bytebuffer) might be left unused unless the whole block is released later.

Do you mean T1 requests a size (single slice) bigger than a block? According to what I think, T1 will fail the pre-check (wether T@ and T3 proceed or not) and than fail into exception on the NativeMemoryAllocator level. While, T2 and T3 can continue to allocate from the same block. So no unused block in this scenario.

If you meant that T1's size is bigger when summed with T2 and/or T3 sizes. Than yes, there might be big portion of block unused. I believe this should be a very rare scenario, as our blocks are quite big, and the concurrency should be very specific. However, truly, we planned to create a slice from this unused space and add it to the free list (of NativeMemoryAllocator) so it can be used by others. But I didn't want to request a complicated solution from you, unless you want to write it. The issue of transferring such an unused space to free list can be fixed later in other PR. Or here if you like.

> If there will be more contentions and its ok to tradeoff space to achieve better time, then can use the approach you have suggested. 

I do not understand this comment fully. I think the solution, I explain above, is good both in space and in time. 

> (throw exception on the first overflow using 2 counters to simulate a completion and throw exceptions)

I didn't suggest using two counters, this is a possibility of course. I do not see a big reason to use two Fetch-and-Add (expensive) instructions instead of one. Why do you prefer 2 counters? Probably I am missing something.

> If there will be less such contentions, compareAndSet might be fine as there will be minimal loops.

You are right if there is little to no contention compareAndSet loop is almost the same as Fetch-and-Add. However, Fetch-and-Add is almost the same as compareAndSet while no contention, and with contention Fetch-and-Add is faster. So, doesn't it sound like Fetch-and-Add is preferable?

> As per my understanding, this would not guarantee the operation to be atomic.

You are right. The check I suggested is for the cases with no concurrency. Just to decrease the possibility of unnecessary increase in the `allocated` above the capacity. The check is not atomic.

> Further, since the `allocated` is maintained as `AtomicInteger` and not as `AtomicLong`, there can different side effects of overflow(negative) which can result in inconsistencies in `long allocated() { return allocated.get(); }` response.

Very good point! As for now, block capacity can not be above 2GB (ByteBuffer limitation, inherited from the past). Thus we can hardly get to this problem. But you are right if we allocate 2GB block and then go a little above capacity. If you would like to fix it, go ahead and define `allocated` to be `AtomicLong`.

Looking forward to hear from you! Please let me know what do you think about my comment. Thanks!

thiyanesh:
> Do you mean T1 requests a size (single slice) bigger than a block?

As you have mentioned, this will result in failure of precondition. No. i didn't intent this. 

> If you meant that T1's size is bigger when summed with T2 and/or T3 sizes.

Yes, i meant this. But even ignoring concurrently executing threads, this will be an issue even for a sequential execution. So, as you have mentioned, this has to be handled independently for all scenarios.

> I do not see a big reason to use two Fetch-and-Add (expensive) instructions instead of one. Why do you prefer 2 counters?

The primary reason suggestion of 2 counter fetch-and-add was:
`long allocated() { return allocated.get(); }` - It might not be good to return an invalid value(either negative/large value if overflow or outside the range of integer for 2GiB even if it's long).
I am not sure about being able to do this in single fetch-and-add.
With 2 counters as mentioned by @Eshcar, the flow will be 
1. supporting internal counter - getAndAdd. if this is invalid(negative or above capacity), then throw exception. don't modify the `allocated` counter and it stays in a valid state.
2. 2 counters simulate the possibility of crossing the limit without actually crossing the limit in the exposed state(`allocated`)

> As for now, block capacity can not be above 2GB (ByteBuffer limitation, inherited from the past). Thus we can hardly get to this problem. But you are right if we allocate 2GB block and then go a little above capacity.

Yes, as you have mentioned the following can simulate the behavior:
With a 2GiB (Integer.MAX_VALUE) limit, using a fetch-and-add can result in negative(as integer) or an invalid bigger value(long).
1. allocated = Integer.MAX_VALUE - 3
2. 2 new  requests each with size = 2
3. any check (allocated.get() + size <= capacity) on these independently will succeed 
4. both of then will proceed to update allocated.getAndAdd(2)
5. this will result in an invalid state of `long allocated() { return allocated.get(); }`

I not sure whether the above descriptions convey the intent. The actual intent was 
1. always increment allocated to only valid states within the range of capacity.
2. ensure the `long allocated`, stays inline with a valid state.
3. (in the existing master branch code with subtraction), avoid overlapping buffer state.

Personally, to satisfy all these 3 conditions in a concurrent execution environment, we might need synchronized or CAS loop or 2 counters with fetch-and-add.

Please do share your feedback and decide whether this PR is relevant or can be solved in different approach.

sanastas:
Hi @thiyanesh ,

Thank you for clarifying! I think I understand now your reason to use two counters and your concern about Block's method `long allocated() { return allocated.get(); } `

In what I have previously suggested there is always a possibility of `allocated` counter of Block being above the capacity. So you see it as wrong output, which is very reasonable.

Please pay attention that Block's `allocated()` method is used only once in `NativeMemoryAllocator` class, just to check whether the block is allocated above its capacity. It is not used for true accounting of Oak's allocated memory. Inside, `NativeMemoryAllocator` class we have another `AtomicLong allocated` counter which is indeed used for true accounting of Oak's allocated memory.

Therefore, the fact that Block's `allocated()` method returns a value above capacity, actually makes no harm. However, your concern is absolutely legitimate, as the code looks like not working right. What do you think about just changing the name of the Block's `allocated()` method (something about being upper limit) and/or adding a good comments around it?

> (in the existing master branch code with subtraction), avoid overlapping buffer state.

I am not sure I understand the requirement above.

So still I would prefer no synchronize, no CAS loop and no double F&A, because I do not work the good path to work slower in order to resolve the corner cases that may rarely happen.

In my eyes the following solution is optimal:
1. Rename Block's `allocated()` method and/or comment
2. Make Block's `allocated` counter AtomicLong
3. Remove decrease of the `allocated` counter when going above capacity
4. [Optional] In order not to increase allocated in sequential execution, add a check `if (allocated.get() + size > this.capacity)` although it is not atomic.
5. [Very optional] add the remaining unused part of the block tot he free list of `NativeMemoryAllocator`

Or any other solution that doesn't impact the normal/good path execution that you want to suggest. What do you think?  Does it make sense to you? 

Thank you for taking care of Oak!!!

sanastas:
@thiyanesh , anything new? Would be great to hear from you! 
I wish your ideas will land to Oak main code. Looking forward!

thiyanesh:
@sanastas ,
please find the updated PR as per the discussions.

PR #148, opened 2021-03-20
reused identical index computation in common serializer and comparator

thiyanesh:
reused identical index computation in common serializer and comparator in OakIntBufferComparator, OakIntBufferSerializer, and OakStringComparator.

### DIFF
--- a/core/src/main/java/com/yahoo/oak/common/OakCommonBuildersFactory.java
+++ b/core/src/main/java/com/yahoo/oak/common/OakCommonBuildersFactory.java
@@ -52,8 +52,10 @@ public class OakCommonBuildersFactory {
 
     public static OakMapBuilder<ByteBuffer, ByteBuffer> getDefaultIntBufferBuilder(int keySize, int valueSize) {
         ByteBuffer minKey = ByteBuffer.allocate(keySize * Integer.BYTES);
+        int index = 0;
         for (int i = 0; i < keySize; i++) {
-            minKey.putInt(Integer.BYTES * i, Integer.MIN_VALUE);
+            minKey.putInt(index, Integer.MIN_VALUE);
+            index += Integer.BYTES;
         }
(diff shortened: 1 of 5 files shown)


thiyanesh:
@sanastas,

please find the separate PR for computation reuse. Few methods in comparator & serializer were recomputing the same offset/index values for the buffers. Currently, i have moved this to single variable and contained the scope within the loop.
Do you prefer to 
1. contain scope with multiplication 
2. or elevate scope just outside the loop and perform addition(instead of multiplication)?

sanastas:
Hi @thiyanesh ,

The code looks good to me. I would prefer second: to elevate scope just outside the loop and perform addition(instead of multiplication).

As to ILP I think it depends on compiler and the underlying hardware, so it would be hard to guess which solution is better performance-wise. You can run Oak benchmarks which use OakStringComparator and OakIntBufferSerializer and probably see the difference between two suggestions. Maybe just "4c-get-copy" benchmark.

Do you need help with running Oak's benchmarks? It is `run.sh -b oak -e 4c-get-copy -t "1 4 8 12" &` where -t is the different numbers of threads you want to try.

sanastas:
However, if you prefer not to invest in this PR anymore, I am ready to accept the PR as is.

thiyanesh:
@sanastas, 
please find the updated PR as per the discussion to use repeated addition in loops.

thiyanesh:
@sanastas,
A small request, if you approve this PR, please "Squash and Merge" to combine this PR into a single commit in master branch.(to suppress unnecessary minor commits in this PR branch)
If your teams approach is not to Squash and merge, then i can send another PR with these changes as single commit. 
Please share about your preferred approach.

sanastas (approved):
The changes look good

sanastas:
Squashed and merged

Issue #132, opened 2021-02-28
Elimination of all ByteBuffer Usage (for Java 8)


sanastas:
Base all of the off-heap memory management on unsafe allocation and access, without ByteBuffer intermediate layer (in addition to existing memory management). The reason for the change is the possibility to gain better throughput without spending CPU cycles and memory on ByteBuffer internals.

That means a new memory manager to be based on using addresses instead of ByteBuffers and allocating the address from unsafe:
`Unsafe().allocateMemory(size);`

That requires a significant code writing: new (alternative) Block, BlockPool, MemoryAllocator, MemoryManager etc.

For more explanations feel free to add questions in this issue.
--

dynaxis:
It doesn't seem that the doc comments on the relevant code are changed in sync with the changes. For instance, `OakUnsafeDirectBuffer`'s `getByteBuffer()` returns a new `ByteBuffer` on every invocation of the method. But the doc comment still has state of the returned `ByteBuffer` shouldn't be modified mentioning it might be shared and used elsewhere.

sanastas:
Hi @dynaxis ,

Thanks for your comment. Here is the comment/documentation about `OakUnsafeDirectBuffer`'s `getByteBuffer()` 

```
public interface OakUnsafeDirectBuffer {
    /**
     * Return the underlying memory address wrapped as a ByteBuffer.
     * This buffer might contain data that is unrelated to the context in which this object was introduced.
     * Note 1: depending on the context (casting from OakScopedReadBuffer or OakScopedWriteBuffer), the buffer mode
     *         might be ready only.
     * Note 2: the buffer internal state (e.g., byte order, position, limit and so on) should not be modified as this
     *         object might be shared and used elsewhere.
     *
     * @return the underlying memory address wrapped as a ByteBuffer.
     */
    ByteBuffer getByteBuffer();

```

Indeed, a new `ByteBuffer` object is returned on each invocation. However, the underlying memory referenced from this `ByteBuffer` is always the same (for the same `OakUnsafeDirectBuffer ` object). Therefore, indeed, modifying those two `ByteBuffers` may lead to corrupted results. This is why the interface is named Oak-*Unsafe*-DirectBuffer and it needs to be used with extra care.

How would you suggest to rephrase the documentation so it is more clear?

sanastas:
Hi @dynaxis ,

[...]

PR #151, opened 2021-03-25
Elimination of all ByteBuffer Usage (for Java 8) #132

li0nr:
resolving issue #132 
removing bytebuffer and replacing them with access though Unsafe 

### DIFF
--- a/benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
+++ b/benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
@@ -101,11 +101,9 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
             ScopedReadBuffer keybb = new ScopedReadBuffer(new Slice());
             ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
             allocator.allocate(keybb.getSlice(), key.calculateSerializedSize());
-            keybb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(keybb.getSlice(), key, MyBuffer.DEFAULT_SERIALIZER);
             newCell.key.set(keybb);
             allocator.allocate(valuebb.getSlice(), value.calculateSerializedSize());
-            valuebb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(valuebb.getSlice(), value, MyBuffer.DEFAULT_SERIALIZER);
             if (!newCell.value.compareAndSet(null, valuebb)) {
                 allocator.free(valuebb.getSlice());
@@ -114,7 +112,6 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
             if (prevValue.value.get() == null) {
                 ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
                 allocator.allocate(valuebb.getSlice(), value.calculateSerializedSize());
-                valuebb.getSlice().duplicateBuffer();
                 ScopedWriteBuffer.serialize(valuebb.getSlice(), value, MyBuffer.DEFAULT_SERIALIZER);
                 if (!prevValue.value.compareAndSet(null, valuebb)) {
                     allocator.free(valuebb.getSlice());
@@ -138,11 +135,9 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
             ScopedReadBuffer keybb = new ScopedReadBuffer(new Slice());
             ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
             allocator.allocate(keybb.getSlice(), key.calculateSerializedSize());
-            keybb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(keybb.getSlice(), key, MyBuffer.DEFAULT_SERIALIZER);
             newCell.key.set(keybb);
             allocator.allocate(valuebb.getSlice(), value.calculateSerializedSize());
-            valuebb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(valuebb.getSlice(), value, MyBuffer.DEFAULT_SERIALIZER);
             if (!newCell.value.compareAndSet(null, valuebb)) {
                 allocator.free(valuebb.getSlice());
@@ -241,7 +236,6 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
             if (prevValue.value.get() == null) {
                 ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
                 allocator.allocate(valuebb.getSlice(), value.calculateSerializedSize());
-                valuebb.getSlice().duplicateBuffer();
                 ScopedWriteBuffer.serialize(valuebb.getSlice(), value, MyBuffer.DEFAULT_SERIALIZER);
                 if (!prevValue.value.compareAndSet(null, valuebb)) {
                     allocator.free(valuebb.getSlice());
@@ -270,11 +264,9 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
             ScopedReadBuffer keybb = new ScopedReadBuffer(new Slice());
             ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
             allocator.allocate(keybb.getSlice(), key.calculateSerializedSize());
-            keybb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(keybb.getSlice(), key, MyBuffer.DEFAULT_SERIALIZER);
             retval.key.set(keybb);
             allocator.allocate(valuebb.getSlice(), value.calculateSerializedSize());
-            valuebb.getSlice().duplicateBuffer();
             ScopedWriteBuffer.serialize(valuebb.getSlice(), value, MyBuffer.DEFAULT_SERIALIZER);
             if (!retval.value.compareAndSet(null, valuebb)) {
                 allocator.free(valuebb.getSlice());
--- a/benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
+++ b/benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
@@ -226,8 +226,7 @@ public class OffHeapList<K extends MyBuffer, V extends MyBuffer> implements Comp
         Consumer<OakScopedWriteBuffer> computeFunction = writeBuffer -> {
             OakUnsafeDirectBuffer buffer = (OakUnsafeDirectBuffer) writeBuffer;
             ByteBuffer buf = buffer.getByteBuffer();
-            int pos = buffer.getOffset();
-            buf.putLong(pos + 1, ~buf.getLong(pos + 1));
+            buf.putLong(1, ~buf.getLong(1));
         };
 
         BiFunction<Object, Cell, Cell> fun = (prevValueO, v) -> {
--- a/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
+++ b/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
@@ -38,15 +38,19 @@ public class MyBuffer implements Comparable<MyBuffer> {
 
     public static void serialize(MyBuffer inputBuffer, OakScopedWriteBuffer targetBuffer) {
         // In the serialized buffer, the first integer signifies the size.
-    	targetBuffer.putInt(0,  inputBuffer.capacity);
+    	int targetPos  = 0;
+    	targetBuffer.putInt(targetPos,  inputBuffer.capacity);
+    	targetPos += Integer.BYTES;
         OakIntBufferSerializer.copyBuffer(inputBuffer.buffer, DATA_POS, inputBuffer.capacity / Integer.BYTES,
-        		targetBuffer, Integer.BYTES);
+        		targetBuffer, targetPos);
     }
 
     public static MyBuffer deserialize(OakScopedReadBuffer inputBuffer) {
-        int capacity = inputBuffer.getInt(0);
+    	int inputPos  = 0;
+    	int capacity = inputBuffer.getInt(inputPos);
+    	inputPos += Integer.BYTES;
         MyBuffer ret = new MyBuffer(capacity);
-        OakIntBufferSerializer.copyBuffer(inputBuffer, Integer.BYTES, capacity / Integer.BYTES, ret.buffer, DATA_POS);
+        OakIntBufferSerializer.copyBuffer(inputBuffer, inputPos, capacity / Integer.BYTES, ret.buffer, DATA_POS);
         return ret;
     }
 
--- a/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
+++ b/benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
@@ -9,14 +9,11 @@ package com.yahoo.oak.synchrobench;
 import com.yahoo.oak.OakComparator;
 import com.yahoo.oak.OakScopedReadBuffer;
 import com.yahoo.oak.OakSerializer;
-import com.yahoo.oak.OakUnsafeDirectBuffer;
-import com.yahoo.oak.UnsafeUtils;
 import com.yahoo.oak.OakScopedWriteBuffer;
 import com.yahoo.oak.common.intbuffer.OakIntBufferComparator;
 import com.yahoo.oak.common.intbuffer.OakIntBufferSerializer;
 
 import java.nio.ByteBuffer;
-import java.util.function.IntPredicate;
 
 public class MyBuffer implements Comparable<MyBuffer> {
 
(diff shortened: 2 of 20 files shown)


core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -42,7 +38,7 @@ boolean allocate(Slice s, int size) {
             allocated.getAndAdd(-size);
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
-        s.associateBlockAllocation(id, now, size, buffer);
+        s.associateBlockAllocation(id, now, size, memAddress, capacity);
```
sanastas, line 41: I also wonder, why to pass the block's capacity all over...? All blocks have the same capacity used to `UnsafeUtils.unsafe.allocateMemory(this.capacity)` in every Block creation. Can we have (for example) static  `getCapacity()` interface in Block class or MemoryManager class/interface? And this can be used anywhere when needed...

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
```diff
@@ -193,7 +193,7 @@ public void readByteBuffer(Slice s) {
         assert blockID > NativeMemoryAllocator.INVALID_BLOCK_ID :
                 String.format("Invalid block-id: %s", s);
         Block b = blocksArray[blockID];
-        s.setBuffer(b.getBuffer());
+        s.setBuffer(b.getmemAdress(), b.getCapacity());
```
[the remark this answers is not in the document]
sanastas: I would also search through the Oak code for the world "ByteBuffer" (without matching the letters case) and see if it is used correctly. Because after your change there should be no internal ByteBuffer usage in Oak...

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
```diff
@@ -47,7 +47,9 @@
      *
      * @return the underlying ByteBuffer.
      */
-    ByteBuffer getByteBuffer();
+    default ByteBuffer getByteBuffer() {
```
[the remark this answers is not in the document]
sanastas: Because I don't think it should be here...

core/src/main/java/com/yahoo/oak/Slice.java
```diff
@@ -251,7 +247,7 @@ public int getLength() {
 
     @Override
     public long getAddress() {
-        return ((DirectBuffer) buffer).address() + getOffset();
+        return memAddress + getOffset();
```
sanastas, line 250: So here you return the exact address and not the start of the entire block as written elsewhere in the code. Please check Slice's `getAddress() `usage and update the comments everywhere.

core/src/main/java/com/yahoo/oak/common/intbuffer/OakIntBufferComparator.java
```diff
@@ -29,16 +30,47 @@ public int compareKeys(ByteBuffer buff1, ByteBuffer buff2) {
...
 
     @Override
     public int compareKeyAndSerializedKey(ByteBuffer key, OakScopedReadBuffer serializedKey) {
         OakUnsafeDirectBuffer unsafeKey = (OakUnsafeDirectBuffer) serializedKey;
-        return compare(key, 0, size, unsafeKey.getByteBuffer(), unsafeKey.getOffset(), size);
+        return compare(key, 0, size, unsafeKey.getAddress(), size);
     }
 
+    
+    public static int compare(long buff1, int size1, long buff2, int size2) {
```
sanastas, line 44: Also here, I think, we can avoid using both `OakUnsafeDirectBuffer` and `UnsafeUtils` and stick with given `OakScopedReadBuffer` interface...

sanastas (changes requested):
Hi Ramy,

This is just my really initial review. Your code is substantial and deals with the very important core part of Oak's memory management. Therefore thorough code review is required and will be done. But we are on the good path! :) I will set a meeting with you.

Thanks,
Anastasia

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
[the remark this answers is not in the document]
sanastas: For now just write an explanatory comment

core/src/main/java/com/yahoo/oak/Slice.java
```diff
@@ -69,7 +65,7 @@ Slice getDuplicatedSlice() {
     // but the same off-heap memory
     @VisibleForTesting
     void duplicateBuffer() {
-        buffer = buffer.duplicate();
+        memAddress = memAddress;
```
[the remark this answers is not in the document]
sanastas: Check if `duplicateBuffer() `can be eliminated at all

core/src/main/java/com/yahoo/oak/Block.java
technionRamy: I needed to attach the capacity with slice since, when wrapping the address with bytebuffer, I just have access to slice.
I can make the capacity static in the Native allocator but wont that affect another Oak instance that wants different capacity?

sanastas: We currently have one BlockPool for all Oak instances, meaning only one block capacity supported. It can be differently defined when the first Oak instance (on the same JVM) is built, after that block capacity cannot be changed.

core/src/main/java/com/yahoo/oak/BlockMemoryAllocator.java
```diff
@@ -28,8 +28,8 @@
     // Returns the memory allocation of this OakMap (this Allocator)
     long allocated();
 
-    // Read the buffer from an allocation object that its parameters are set: blockID, offset and length
-    void readByteBuffer(Slice s);
+    // Attaches the slice with its base address
+    void getSlicebaseAddress(Slice s);
```
sanastas, line 32: coding convention requires name to be `getSliceBaseAddress()`

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
```diff
@@ -37,70 +36,46 @@ public int compareTo(MyBuffer o) {
...
-        targetPos += Integer.BYTES;
-
+    	targetBuffer.putInt(0,  inputBuffer.capacity);
         OakIntBufferSerializer.copyBuffer(inputBuffer.buffer, DATA_POS, inputBuffer.capacity / Integer.BYTES,
-                targetBuffer, targetPos);
-    }
-
-    public static MyBuffer deserialize(OakUnsafeDirectBuffer inputBuffer) {
-        return deserialize(inputBuffer.getByteBuffer(), inputBuffer.getOffset());
+        		targetBuffer, Integer.BYTES);
```
sanastas, line 43: I would remain with `targetPos` internal variable, initiating it initially to zero, then adding `Integer.BYTES`

sanastas: Same for the `deserialize()` below

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
```diff
@@ -37,70 +36,46 @@ public int compareTo(MyBuffer o) {
...
                 buff2, pos2, cap2 / Integer.BYTES);
     }
 
-    public static int compareBuffers(OakUnsafeDirectBuffer buffer1, OakUnsafeDirectBuffer buffer2) {
-        ByteBuffer buf1 = buffer1.getByteBuffer();
-        ByteBuffer buf2 = buffer2.getByteBuffer();
-
-        int pos1 = buffer1.getOffset();
-        int pos2 = buffer2.getOffset();
+    public static int compareBuffers(ByteBuffer buff1, int pos1, int cap1, OakScopedReadBuffer buff2, int pos2, int cap2) {
```
sanastas, line 59: I am a bit lost here... Can you please add some comments? Why so many `compareBuffers`? Why are they all public?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -27,7 +22,8 @@
         this.id = NativeMemoryAllocator.INVALID_BLOCK_ID;
         // Pay attention in allocateDirect the data is *zero'd out*
         // which has an overhead in clearing and you end up touching every page
-        this.buffer = ByteBuffer.allocateDirect(this.capacity);
+        this.memAddress = UnsafeUtils.unsafe.allocateMemory(this.capacity);
```
sanastas, line 25: Better to create an `allocateMemory()` interface of `UnsafeUtils` and not to access the `unsafe` field. It should be private.

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -59,26 +55,11 @@ long allocated() {
...
-            e.printStackTrace();
-        }
-        assert cleaner != null;
-        cleaner.clean();
+        UnsafeUtils.unsafe.freeMemory(memAddress);
     }
 
-    ByteBuffer getBuffer() {
-        return buffer;
+    long getMemAddress() {
```
sanastas, line 61: Maybe to name it `getStartMemAddress()` ?

core/src/main/java/com/yahoo/oak/BlockMemoryAllocator.java
sanastas: Isn't it then `setSliceAddress()`? Is it setting the base address? Or the direct, unique address of this Slice?

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
```diff
@@ -47,8 +47,7 @@
      *
      * @return the underlying ByteBuffer.
      */
-    ByteBuffer getByteBuffer();
-
+    ByteBuffer getByteBuffer(); //description change?
```
sanastas, line 50: Using reflection, can we create a smaller ByteBuffer? Not the entire ByteBuffer of the Block? If yes, this should be much preferable...

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
```diff
@@ -104,10 +104,7 @@ public double getDouble(int index) {
      *
      * @return the underlying ByteBuffer.
      */
-    @Override public ByteBuffer getByteBuffer() {
-        return s.getByteBuffer();
-    }
-
+
```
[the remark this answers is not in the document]
sanastas: I don't think this comment should be totally removed. Just modified and put in other place...

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
```diff
@@ -123,14 +109,23 @@ public double getDouble(int index) {
...
      *
-     * @return the exact memory address of the underlying buffer in the position of the data.
+     * @return the exact memory address of the Buffer in the position of the data.
      */
-    @Override public long getAddress() {
+    @Override public long getAddress() { //for both OakUnsafeDirectBuffer -- OakScopedReadBuffer
         return s.getAddress();
     }
+    
+   //must never be called internally kept in here to beautify the interface
```
sanastas, line 122: It should be called internally, but externally it can be called. So it is not just for the beauty of the interface...

sanastas (changes requested):
Some more comments.

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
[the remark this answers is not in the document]
li0nr: getbytebuffer must never be used internally, the method name is still in here but does nothing to comply with interface
thus the comment is not needed in here

core/src/test/java/com/yahoo/oak/ValueUtilsTest.java
```diff
@@ -32,18 +32,21 @@ public void init() {
     }
 
     private void putInt(int index, int value) {
-        s.getSlice().getByteBuffer().putInt(s.getOffset() + index, value);
+        UnsafeUtils.unsafe.putInt(s.getAddress()+index, value);
```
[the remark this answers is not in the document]
li0nr: ScopedReadBuffer doesnt have put methods, right? or did i miss somthing?

core/src/test/java/com/yahoo/oak/ValueUtilsSimpleTest.java
```diff
@@ -24,7 +24,7 @@ public void init() {
             new SyncRecycleMemoryManager(new NativeMemoryAllocator(128));
         s = syncRecycleMemoryManager.getEmptySlice();
         syncRecycleMemoryManager.allocate(s, 16, false);
-        s.getByteBuffer().putInt(s.getOffset(), 1);
+        UnsafeUtils.unsafe.putInt(s.getAddress(), 1);
```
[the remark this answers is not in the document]
li0nr: in this test slice is used for other stuff as well, 
Moreover if i want to change it i need it to be writebuffer, which has a private constructor 
all in all it will be messier

core/src/main/java/com/yahoo/oak/common/intbuffer/OakIntBufferSerializer.java
```diff
@@ -46,10 +42,18 @@ public int getSizeBytes() {
         return size * Integer.BYTES;
     }
 
-    public static void copyBuffer(ByteBuffer src, int srcPos, int srcSize, ByteBuffer dst, int dstPos) {
+
+    public static void copyBuffer(ByteBuffer src, int srcPos, int srcSize, OakScopedWriteBuffer dst, int dstPos) {
```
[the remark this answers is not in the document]
li0nr: 1. yes it is used in Mybuffer synchrobench
2. mmm... I dont think so?

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
```diff
@@ -37,70 +36,46 @@ public int compareTo(MyBuffer o) {
...
-    }
-
-    public static void serialize(MyBuffer inputBuffer, ByteBuffer targetBuffer, int targetPos) {
+    public static void serialize(MyBuffer inputBuffer, OakScopedWriteBuffer targetBuffer) {
         // In the serialized buffer, the first integer signifies the size.
-        targetBuffer.putInt(targetPos, inputBuffer.capacity);
-        // Thus, the data position starts after the first integer.
-        targetPos += Integer.BYTES;
-
+    	targetBuffer.putInt(0,  inputBuffer.capacity);
```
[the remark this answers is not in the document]
li0nr: yes the offset is calculated in side the scopedreadbuffer, the index is still needed

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
li0nr: for the write buffer there is no variable to indicate the current offset

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
li0nr: what we had initally is just comparing bytebuffers, which is still exposed 
I added the option to compare ScopedReadBuffer

core/src/main/java/com/yahoo/oak/BlockMemoryAllocator.java
li0nr: here we are setting the base address

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
li0nr: in this case it is, we throw exception here

core/src/main/java/com/yahoo/oak/Block.java
li0nr: changed to use the slice length in order to wrap the bytebuffer

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
```diff
@@ -101,11 +101,9 @@ public void putOak(K key, V value) {
             ScopedReadBuffer keybb = new ScopedReadBuffer(new Slice());
             ScopedReadBuffer valuebb = new ScopedReadBuffer(new Slice());
             allocator.allocate(keybb.getSlice(), key.calculateSerializedSize());
-            keybb.getSlice().duplicateBuffer();
```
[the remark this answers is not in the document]
li0nr: it is working fine :D

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
li0nr: done!
in this case we dont need to save the capacity of the allocated memory

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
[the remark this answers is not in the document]
sanastas: 👍

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/MyBuffer.java
sanastas: lets discuss

core/src/main/java/com/yahoo/oak/BlockMemoryAllocator.java
sanastas: lets discuss

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
sanastas: 👍

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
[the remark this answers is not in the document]
sanastas: lets discuss

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
sanastas: lets discuss

core/src/test/java/com/yahoo/oak/ValueUtilsTest.java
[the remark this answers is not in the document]
sanastas: you are right

core/src/test/java/com/yahoo/oak/ValueUtilsSimpleTest.java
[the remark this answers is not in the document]
sanastas: OK

benchmarks/synchrobench/src/test/java/com/yahoo/oak/OakMyBufferTest.java
```diff
@@ -52,7 +52,7 @@ public void testPut() {
     @Test
     public void testIncreasePut() {
         Integer v =  0;
-        for (v = 1; v < 1000000; v++) {
+        for (v = 1; v < 10000; v++) {
```
sanastas, line 55: Why to decrease the input? Any good reason?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -6,15 +6,11 @@
...
-import sun.misc.Cleaner;
-
-import java.lang.reflect.Field;
-import java.nio.ByteBuffer;
 import java.util.concurrent.atomic.AtomicLong;
 
 class Block {
 
-    private final ByteBuffer buffer;
+    private final long memAddress;
```
sanastas, line 13: Please add some comment about what this address is, and explain the ability of having a ByteBuffer here.

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -45,7 +42,7 @@ boolean allocate(Slice s, final int size) {
         if (now + size > this.capacity) {
             throw new OakOutOfMemoryException(String.format("Block %d is out of memory", id));
         }
-        s.associateBlockAllocation(id, (int) now, size, buffer);
+        s.associateBlockAllocation(id, (int) now, size, memAddress, capacity);
```
sanastas, line 45: Didn't we say the capacity is no longer needed?

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -64,26 +61,11 @@ long allocatedWithPossibleDelta() {
...
-        cleanerField.setAccessible(true);
-        Cleaner cleaner = null;
-        try {
-            cleaner = (Cleaner) cleanerField.get(buffer);
-        } catch (IllegalAccessException e) {
-            e.printStackTrace();
-        }
-        assert cleaner != null;
-        cleaner.clean();
+        UnsafeUtils.unsafe.freeMemory(memAddress);
```
sanastas, line 64: Liran had a suggestion that this transfer to another way of memory release may have influenced the memory consumption. This is a bit speculation, as all what `freeMemory()` does' is just releasing memory back to OS. But we can check it by commenting this out and looking on the **on-heap** memory consumption.

core/src/main/java/com/yahoo/oak/Block.java
sanastas: I would name it `blockMemAddress` as the start of the big chunk of memory related to the Block. And inside the Slice I would add Slice's offset and keep `sliceMemAddress`.

core/src/main/java/com/yahoo/oak/OakBuffer.java
```diff
@@ -13,8 +13,8 @@
  * It mimic the standard interface of Java's ByteBuffer, for example, int getInt(int index), char getChar(int index),
  * capacity(), etc.
  */
-public interface OakBuffer {
-
+public interface    OakBuffer {
```
sanastas, line 16: Can you please remove the unnecessary empty spaces?

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
sanastas, line 50: Please do the description change and remove the comment in this line.

sanastas (commented):
Please refer to my last comments. Thanks!

core/src/main/java/com/yahoo/oak/Block.java
li0nr: did not work :(

core/src/main/java/com/yahoo/oak/Slice.java
```diff
@@ -251,13 +236,19 @@ public int getLength() {
...
     @Override
     public String toString() {
         return String.format("Slice(blockID=%d, offset=%,d, length=%,d, version=%d)", blockID, offset, length, version);
     }
+    
+    //must never be called internally kept in here to beautify the interface
+    @Override
+    public ByteBuffer getByteBuffer() { 
+        throw new RuntimeException(); 
+    }
```
liran-funaro, line 251: Please remove this method entirely. 
Note that `Slice` should not implement `OakUnsafeDirectBuffer` anymore, so please remove `implements OakUnsafeDirectBuffer` from `Slice` declaration.
Also, please remove `order()` from `OakBuffer` API. We use native order now.

liran-funaro (changes requested):
Looks good, but needs a few modifications. See my comments below.

core/src/main/java/com/yahoo/oak/Slice.java
li0nr: I think that Slice should still implement the methods of OakUnsafeDirectBuffer, we use these methods in ScopedReadBuffer for example.
what can I do is to remove the implements from the OakUnsafeDirectBuffer but still have the method, is that what you have meant?

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
```diff
@@ -25,12 +25,12 @@
         this.s = other.s.getDuplicatedSlice();
     }
 
-    protected int getDataOffset(int index) {
+    protected long getDataOffset(int index) {
```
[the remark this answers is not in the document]
li0nr: this doesnt return the address of the data it really returns the offset, which I think is never used since we can get the exact address of the data using the getAddress() method

core/src/main/java/com/yahoo/oak/Block.java
```diff
@@ -64,26 +61,11 @@ long allocatedWithPossibleDelta() {
...
-        cleanerField.setAccessible(true);
-        Cleaner cleaner = null;
-        try {
-            cleaner = (Cleaner) cleanerField.get(buffer);
-        } catch (IllegalAccessException e) {
-            e.printStackTrace();
-        }
-        assert cleaner != null;
-        cleaner.clean();
+       // UnsafeUtils.unsafe.freeMemory(memAddress);
```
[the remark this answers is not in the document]
li0nr: this was mistake pushing new commit in few minutes

core/src/main/java/com/yahoo/oak/OakUnsafeDirectBuffer.java
```diff
@@ -36,7 +36,7 @@
 public interface OakUnsafeDirectBuffer {
 
     /**
-     * Allows access to the underlying ByteBuffer of Oak.
+     * Wraps the underlying memory address with ByteBuffer. This method allows access to this ByteBuffer.
```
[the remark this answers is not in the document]
li0nr: thanks, is is much more informative

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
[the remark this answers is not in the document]
liran-funaro: It returns the address and it is used. Read the rest of the file :)

core/src/main/java/com/yahoo/oak/Slice.java
liran-funaro: Yes. Leave the methods you need and remove the ones that are redundant (e.g., `getByteBuffer()`).
Anyway, it should not be declared with `implements OakUnsafeDirectBuffer`.

core/src/main/java/com/yahoo/oak/ScopedReadBuffer.java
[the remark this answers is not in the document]
li0nr: yep my mistake was thinking of something else

Issue #153, opened 2021-03-30
Add checkstyle to the tests


liran-funaro:
Currently, Oak code standards are enforced by the maven code-style plugin.
The code-style rules are detailed in `codestyle/checkstyle.xml`.
However, these rules are currently not enforced on the tests and benchmarks.

Please change the `pom.xml` file to enforce code style check on all modules and fix all the existing code-style issues.

PR #154, opened 2021-03-30
Apply checkstyle to all modules

liran-funaro:
Fixes issue #153 

- Apply checkstyle to all modules
- Add "WhitespaceAround" to checkstyle rules


sanastas:
Looks like there were check-style errors :(

liran-funaro:
> Looks like there were check-style errors :(

Yeah. I didn't fix the issues yet. Now there are no errors.

core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -762,7 +759,7 @@ DescendingIter descendingIter(ThreadContext ctx, K from, boolean fromInclusive,
     // specifier whether the end boundary check needs to be performed on the current scan output
     enum IterEndBoundCheck {
         NEVER_END_BOUNDRY_CHECK,
```
sanastas, line 761: Maybe also here BOUNDARY instead of BOUNDRY?

sanastas (approved):
Very good contribution, minor comment below.

core/src/main/java/com/yahoo/oak/Chunk.java
liran-funaro: fixed

PR #159, opened 2021-04-28
Make UnsafeUtils public

liran-funaro:
This PR merges the changes in 0.2.3.1 to master.


core/src/main/java/com/yahoo/oak/DirectUtils.java
```diff
@@ -13,26 +13,32 @@
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
 
-final class UnsafeUtils {
+public final class DirectUtils {
```
sanastas, line 16: As this is public, I suggest to add some comments here about the usage and also about Java version required. Maybe some very short mentioning in the README as well.

sanastas (approved):
Looks good, small comment below.

core/src/main/java/com/yahoo/oak/DirectUtils.java
liran-funaro: To the best of my knowledge, this should work on any version as long as Oak was compiled on Java8.

sanastas: So just add a comment about it, some users may question about the possible usage

liran-funaro: I added documentation

PR #163, opened 2021-06-29
First part of the Entry Hash set

sanastas:


liran-funaro (commented):
This is part 1 of the review.

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -74,10 +74,10 @@
     final OakSerializer<V> valueSerializer;
 
     /**
-     * Create a new EntrySet
+     * Create a new EntryOrderedSet
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -232,6 +245,55 @@ protected void copyEntriesFrom(EntryArray<K, V> other, int srcEntryIdx, int dest
                 entryIdx2LongIdx(destEntryIndex), fieldCount);
     }
 
+    /*
+     * isValueRefValidAndNotDeleted is used only to check whether the value reference, which is part of the
+     * entry on entry index "ei" is valid and not deleted. No off-heap value deletion mark check.
+     * Reference being marked as deleted is checked.
+     *
+     * Pay attention that (given entry's) value may be deleted asynchronously by other thread just
+     * after this check. For the thread safety use a copy of value reference.
```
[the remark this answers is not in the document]
sanastas: This is a confusing comment. The meaning was to emphasize that the check in a specific point of time, doesn't protect from value (and/or its reference) will be deleted just after. The copy (especially of the reference) will not protect from being deleted. I fixed the comment not the code.

core/src/main/java/com/yahoo/oak/ThreadContext.java
```diff
@@ -56,7 +59,7 @@
     final ValueBuffer tempValue;
 
     ThreadContext(MemoryManager kmm, MemoryManager vmm) {
-        entryIndex = EntrySet.INVALID_ENTRY_INDEX;
+        entryIndex = EntryOrderedSet.INVALID_ENTRY_INDEX;
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+
+    /**
+     * getFullHashEntryIndex returns the full hash entry index (the number only!) of the entry
+     * given by entry index "ei". The method serves internal and external EntryHashSet users.
+     */
+    private int getFullHashEntryIndex(int ei) {
+        assert isIndexInBound(ei);
+        // TODO: need to extract the number from the updates counter and to return only the integer
+        return (int) getEntryFieldLong(ei, HASH_FIELD_OFFSET);
+    }
```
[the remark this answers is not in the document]
sanastas: fixed, the code of managing the hash number will be added later

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+    static final int INVALID_FULL_HASH = 0; // because memory is initially zeroed
+    static final int DEFAULT_COLLISION_ESCAPES = 3;
+    // HASH - the full hash index of this entry (one integer, all bits).
+    private static final int HASH_FIELD_OFFSET = 2;
+
+    // One additional to the key and value references field, which keeps the full hash index of the entry
+    private static final int ADDITIONAL_FIELDS = 1;  // # of primitive fields in each item of entries array
+
+    // number of entries candidates to try in case of collision
+    private AtomicInteger collisionEscapes = new AtomicInteger(DEFAULT_COLLISION_ESCAPES);
```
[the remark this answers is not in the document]
sanastas: Good name, changed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+            // fully deleted (EntryState.DELETED) in this case we cannot compare the key (!)
+            // also deletion linearization point is checked during getEntryState()
+            if (ctx.entryState != EntryState.DELETED &&
+                equalKeyAndEntryKey(ctx.key, key, ctx.entryIndex, fullHashIdx)) {
+                // EntryState.VALID --> the key is found
+                // DELETED_NOT_FINALIZED/UNKNOWN --> key doesn't exists
+                //                      and there is no need to continue to check next entries
+                // INSERT_NOT_FINALIZED --> before linearization point, key doesn't exist
+                // when more than unique keys can be concurrently inserted, need to check further!
+                return (ctx.entryState == EntryState.VALID) ? true : false;
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+
+    /**
+     * lookUp checks whether key exists in the given hashIdx or after.
+     * Given initial hash index for the key, it checks entries[hashIdx] first and continues
+     * to the next entries up to 'collisionEscapes', if key wasn't previously found.
+     * If true is returned, ctx.entryIndex keeps the index of the found entry
+     * and ctx.entryState keeps the state.
+     *
+     * IMPORTANT:
+     *  1. Throws IllegalStateException if too much collisions are found
```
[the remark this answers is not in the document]
sanastas: This is wrong comment, removed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+     *         Otherwise (false), key wasn't found, ctx with it's buffers is invalidated
+     */
+    boolean lookUp(ThreadContext ctx, K key, int hashIdx, long fullHashIdx) {
+        // start from given hash index
+        // and check the next `collisionEscapes` indexes if previous index is occupied
+        int currentLocation = hashIdx;
+        int collisionEscapesLocal = collisionEscapes.get();
+
+        // as far as we didn't check more than `collisionEscapes` indexes
+        while (Math.abs(currentLocation - hashIdx) < collisionEscapesLocal) {
```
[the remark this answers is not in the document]
sanastas: Indeed, it was wrong, fixed in both places

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+
+    /**
+     * findSuitableEntryForInsert finds the entry where given hey is going to be inserted.
+     * Given initial hash index for the key, it checks entries[hashIdx] first and continues
+     * to the next entries up to 'collisionEscapes', if the previous entries are all occupied.
+     * If true is returned, ctx.entryIndex keeps the index of the chosen entry
+     * and ctx.entryState keeps the state.
+     *
+     * IMPORTANT:
+     *  1. Throws IllegalStateException if too much collisions are found
```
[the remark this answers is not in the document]
sanastas: `IllegalStateException` is thrown:
1. ctx.entryState has unexpected value. This would be a bug. For example if new state is added and this coded is not fixed appropriately.
2. Collision chain length is found be increased too much. It is better to get noticed about this situation, as this can make the user to consider another hash function. If assert would be used it would be compiled out soon.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+                    // deleteValueFinish() also changes the entry state to DELETED
+                    // deliberate break through
+                case DELETED: // deliberate break through
+                case UNKNOWN:
+                    // deliberate break through
+                case INSERT_NOT_FINALIZED: // continue allocating value without allocating a key
+                    entryFound = true;
+                    break;
+                default:
+                    throw new IllegalStateException("Unexpected Entry State");
```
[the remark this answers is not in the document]
sanastas: This would be a bug. For example if new state is added and this coded is not fixed appropriately.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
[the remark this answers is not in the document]
liran-funaro: If it is for debugging, use `assert`.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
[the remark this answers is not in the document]
liran-funaro: If it is for debugging, use `assert`. Otherwise, it will be a dead code.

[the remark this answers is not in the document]
sanastas: OK, here I can accept it

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+            }
+            if (entryFound) {
+                break;
+            }
```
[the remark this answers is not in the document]
sanastas: Yes, I got it from Intellij now. Fixed.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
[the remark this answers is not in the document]
sanastas: This is a point I would discuss, once we collect all the review comments.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,566 @@
...
+            // EntryState.INSERT_NOT_FINALIZED --> you can compete to associate value with the same key
+            switch (ctx.entryState) {
+                case VALID:
+                    if (equalKeyAndEntryKey(ctx.key, key, hashIdx, fullHashIdx)) {
+                        // found valid entry has our key, the inserted key must be unique
+                        // the entry state will indicate that insert didn't happen
+                        return true;
+                    }
+                    currentLocation = (currentLocation + 1) % entriesCapacity; // cyclic increase
+                    continue;
```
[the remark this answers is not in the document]
sanastas: If I change `continue` to be `break` the for loop will have only one iteration. So this fix wasn't applied.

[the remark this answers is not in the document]
sanastas: Please let me know if you agree

sanastas:
I have pushed the changes for the comments so far. Looking forward for the next review round.

core/src/main/java/com/yahoo/oak/ReferenceCodec.java
```diff
@@ -77,13 +77,47 @@ protected ReferenceCodec(int firstBitSizeLimit, int secondBitSizeLimit, int thir
...
+    /**
+     * Construct the codec setting the number of bits to be used for each parameter.
+     * This instance would combine only 2 integers inside long.
+     * The bit sizes of 1 parameters can be given and the third is always the remaining bits.
+     * @param firstBitSizeLimit an upper limit on the size of the first parameter (exclusive)
+     *                          if invalid calculate according to other two limits
+     * @param secondBitSizeLimit an upper limit on the size of the second parameter (exclusive)
+     *                          if invalid calculate according to other two limits
+     */
+    protected ReferenceCodec(int firstBitSizeLimit, int secondBitSizeLimit) {
```
liran-funaro, line 89: This can be implemented by calling the other constructor as follows:
```
this(firstBitSizeLimit, firstBitSizeLimit, 0);
```

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+ *   |         entry hash index                                                 |
+ * ---------------------------------------------------------------------------------------
+ * 3 | Key Reference          | Deleted reference (still) encoding all info     |
+ *   |                        | needed to access the key off-heap               |
+ * -----------------------------------------------------------------------------| Deleted entry with
+ * 4 | Value Reference        | Deleted reference(still)  encoding all info     | entry hash index
+ *   |                        | needed to access the key off-heap.              | 1
+ * -----------------------------------------------------------------------------|
+ * 5 | Hash  - FULL entry hash index, which might be different (bigger) than    |
+ *   |         entry hash index                                                 |
```
liran-funaro, line 39: I'm confused. Is a full hash simply the key's hash number? Or is it the hash+version? 
What is a "non-full-hash"?

I suggest calling the "hash+version" a "versioned hash".
The "full-hash" a "hash value".
And if there there is a use for a "non-full-hash", then maybe a "partial hash"? "hash subset"?
It depends on the use. I didn't find it yet :)

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+
+    // number of entries candidates to try in case of collision
+    private AtomicInteger collisionChainLength = new AtomicInteger(DEFAULT_COLLISION_CHAIN_LENGTH);
+
+    private final OakComparator<K> comparator;
+
+    // use reference codec to encode Full Hash Number integer (first) with its updates counter (second)
+    private final ReferenceCodec hashCodec = new ReferenceCodec(
+        FULL_HASH_BITS, // bits# to represent full hash number as integer, bits# to represent update
+        ReferenceCodec.INVALID_BIT_SIZE); // counter are calculated upon previous parameters (also int)
```
liran-funaro, line 71: I like that you found re-use of this code for other purposes. 
In this case, it no longer serves as a reference codec.
In actually more like a [union type](https://en.wikipedia.org/wiki/Union_type).
Maybe we should extract the "union" part from `ReferenceCodec` to a separate `Union` class and use it here.
Then `ReferenceCodec` can extend `Union`.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+    /**
+     * Create a new EntryHashSet
+     * @param vMM   for values off-heap allocations and releases
+     * @param kMM off-heap allocations and releases for keys
+     * @param entriesCapacity how many entries should this EntryOrderedSet keep at maximum
+     * @param keySerializer   used to serialize the key when written to off-heap
+     */
+    EntryHashSet(MemoryManager vMM, MemoryManager kMM, int entriesCapacity, OakSerializer<K> keySerializer,
+        OakSerializer<V> valueSerializer, OakComparator<K> comparator) {
+        // We add additional field for the head (dummy) node
```
liran-funaro, line 84: This is no longer true. This comment also appears in `EntryOrderedSet` but it is not true there either.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+        long fullHashNumberField = getEntryFieldLong(ei, HASH_FIELD_OFFSET);
+        // extract the hash number (first) from the updates counter (second)
+        return hashCodec.getFirst(fullHashNumberField);
+    }
+
+    /**
+     * getFullHashEntryField returns the full hash entry index (the entire field including counter!)
+     * of the entry given by entry index "ei". The method serves internal and external EntryHashSet users.
+     */
+    private long getFullHashEntryField(int ei) {
```
liran-funaro, line 109: So is this going to be `geVersionedHash()`?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+    /**
+     * getFullHashNumber returns the full hash entry index (the number only!) of the entry
+     * given by entry index "ei". The method serves internal and external EntryHashSet users.
+     */
+    private int getFullHashNumber(int ei) {
+        assert isIndexInBound(ei);
+        long fullHashNumberField = getEntryFieldLong(ei, HASH_FIELD_OFFSET);
+        // extract the hash number (first) from the updates counter (second)
+        return hashCodec.getFirst(fullHashNumberField);
+    }
```
liran-funaro, line 103:
```suggestion
    private int getFullHashNumber(int ei) {
        return hashCodec.getFirst(geVersionedHash());
    }
```

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+        return (getFullHashNumber(ei) != INVALID_FULL_HASH);
+    }
+
+    /**
+     * setFullHashEntryIndex sets the full hash entry index (of the entry given by entry index "ei")
+     * to be the "fullHash". To be used while rebalance, during normal path only CAS is used.
+     */
+    private void setFullHashEntryIndex(int ei, long fullHash) {
+        setEntryFieldLong(ei, HASH_FIELD_OFFSET, fullHash);
+    }
```
liran-funaro, line 127: I understand this sets the entry's hash, but the name suggests that it sets the entry index of the hash, which doesn't make sense.
Did you mean `setHash(int ei, long hash)`?
Just as we named the other methods: `setKeyReference(int ei, long value)` etc.

Anyway, it is never used.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+
+    /**
+     * casFullHashIndex CAS the full hash entry index (of the entry given by entry index "ei") to be
+     * the `newFullHash` only if it was `oldFullHash`, the update counter incorporated inside
+     * `oldFullHash` must match. Also update counter is increased as part of the action.
+     */
+    private boolean casFullHashEntryIndex(int ei, long oldFullHash, int newFullHash) {
```
liran-funaro, line 134: `casHash()`?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+    static final int INVALID_FULL_HASH = 0; // because memory is initially zeroed
+    static final int DEFAULT_COLLISION_CHAIN_LENGTH = 3;
+    // HASH - the full hash index of this entry (one integer, all bits).
+    private static final int HASH_FIELD_OFFSET = 2;
+    private static final int FULL_HASH_BITS = 32; // Hash needs to be an integer
+    // One additional to the key and value references field, which keeps the full hash index of the entry
+    private static final int ADDITIONAL_FIELDS = 1;  // # of primitive fields in each item of entries array
+
+    // number of entries candidates to try in case of collision
+    private AtomicInteger collisionChainLength = new AtomicInteger(DEFAULT_COLLISION_CHAIN_LENGTH);
```
liran-funaro, line 64: Intellij: Field 'collisionChainLength' may be 'final'

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+     *  1. Assuming the entry's key is already read into the tempKeyBuff
+     *
+     * @param tempKeyBuff a reusable buffer object for internal temporary usage
+     *                    As a side effect, this buffer will contain the compared
+     *                    serialized key.
+     * @param key         the key to compare
+     * @param hi          the entry index to compare with
+     * @return true - the keys are equal, false - otherwise
+     */
+    private boolean equalKeyAndEntryKey(KeyBuffer tempKeyBuff, K key, int hi, long fullKeyHashIdx) {
```
liran-funaro, line 158: `isKeyAndEntryKeyEqual()`?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+        // check the hash value comparision first
```
liran-funaro, line 159:
```suggestion
        // check the hash value comparison first
```

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+                    // deliberate break through
+                case DELETED: // deliberate break through
+                case UNKNOWN:
+                    // deliberate break through
+                case INSERT_NOT_FINALIZED: // continue allocating value without allocating a key
+                    entryFound = true;
+                    break;
+                default:
+                    // For debugging the case new state is added and this code is not updated
+                    assert false;
```
liran-funaro, line 322: This will be a dead code.
To avoid it, you can add `assert ctx.entryState != null` before the `switch` clause and then remove the `default:` case.
This should compile well without errors or warnings.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+     * The ctx.key will be populated to point to the found key
+     *
+     * @param ctx the context that will follow the operation following this key allocation
+     * @param key the key to write
+     * @param hashIdx hashIdx=fullHash||bit_size_mask
+     * @param fullHashIdx fullHashIdx=hashFunction(key)
+     * @return true only if the allocation was successful.
+     *         Otherwise (false), rebalance is required
+     **/
+    boolean allocateKey(ThreadContext ctx, K key, int hashIdx, int fullHashIdx) {
```
liran-funaro, line 387: `allocateEntryWriteKey()`?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -0,0 +1,561 @@
...
+        // Write given key object "key" (to off-heap) as a serialized key, referenced by entry
+        // that was set in this context ({@code ctx}).
+        writeKey(key, ctx.key);
+
+        // Try to assign our key
+        if (casKeyReference(ctx.entryIndex, ctx.tempKey.getSlice().getReference(), /* old reference */
+            ctx.key.getSlice().getReference() /* new reference */ )) {
+
+            // key reference CASed (only one should succeed) write the entry's full hash,
+            // because it is used for keys comparision (invalid full hash is not used for comparison)
```
liran-funaro, line 421:
```suggestion
            // because it is used for keys comparison (invalid full hash is not used for comparison)
```

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -74,10 +74,10 @@
     final OakSerializer<V> valueSerializer;
 
     /**
-     * Create a new EntrySet
+     * Create a new instance
      * @param vMM   for values off-heap allocations and releases
      * @param kMM off-heap allocations and releases for keys
-     * @param entriesCapacity how many entries should this EntrySet keep at maximum
+     * @param entriesCapacity how many entries should this EntryArray instance keep at maximum
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -388,7 +475,7 @@ void releaseKey(ThreadContext ctx) {
      * Releases the newly allocated value of the input context.
      * Currently the method is used only to release an
      * unreachable value reference, the one that was not yet attached to an entry!
-     * The method is part of EntrySet, because it cares also
+     * The method is part of EntryOrderedSet, because it cares also
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/main/java/com/yahoo/oak/Chunk.java
```diff
@@ -136,93 +136,94 @@ private void duplicateKeyBuffer(KeyBuffer src, KeyBuffer dst) {
...
 
     /**
-     * See {@code EntrySet.allocateEntry(ThreadContext)} for more information
+     * Writes the key off-heap and allocates an entry with the reference pointing to the given key
+     * See {@code EntryOrderedSet.allocateKey(ThreadContext)} for more information
      */
-    boolean allocateEntry(ThreadContext ctx, K key) {
-        return entrySet.allocateEntry(ctx, key);
+    boolean allocateKey(ThreadContext ctx, K key) {
+        return entryOrderedSet.allocateKey(ctx, key);
```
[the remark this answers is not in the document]
sanastas: renamed

core/src/main/java/com/yahoo/oak/ThreadContext.java
```diff
@@ -13,7 +13,7 @@
      * Entry Context
      *-----------------------------------------------------------*/
 
-    /* The index of the key's entry in EntrySet */
+    /* The index of the key's entry in EntryOrderedSet */
```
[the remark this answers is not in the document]
sanastas: fixed

core/src/test/java/com/yahoo/oak/EntryHashSetTest.java
```diff
@@ -0,0 +1,285 @@
...
+    private final ValueUtils valueOperator = new ValueUtils();
+
+    private void allocateSimpleKeyValue(
+        ThreadContext ctx, EntryHashSet ehs, SyncRecycleMemoryManager memoryManager) {
+
+        // simple one key insert
+        assert ehs.allocateKey(ctx, new Integer(5), 7 /*000111*/, 39 /*100111*/ );
+        assert ctx.entryIndex == 7 && ctx.entryState == EntryArray.EntryState.UNKNOWN
+            && ctx.key.getSlice().getReference() != memoryManager.getInvalidReference()
+            && ctx.value.getSlice().getReference() == memoryManager.getInvalidReference();
```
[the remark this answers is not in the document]
sanastas: fixed in the entire file

core/src/main/java/com/yahoo/oak/ReferenceCodec.java
```diff
@@ -19,9 +19,9 @@
  * IMPORTANT: The 3 parameters (first, second and third) cannot be negative numbers!
  * (As negatives have prefix of ones.)
  */
-abstract class ReferenceCodec {
+class ReferenceCodec {
```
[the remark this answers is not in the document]
sanastas: :)

core/src/main/java/com/yahoo/oak/ReferenceCodec.java
sanastas: Not straight forward, then the other constructor needs to be rewritten. I think the code is clearer as it is now. Not fixed, please reopen if you find it wrong.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: Let's name the output of the hash function applied on the key a "hash number", which is integer: h(K) = hash number

So we have 3 issues here:
1. We have the lower X bits of the hash number serving as an index in the EntryArray. This is what I wanted to name "hash index".

2. We keep the hash number (all the bits) inside an entry. To separate it from the above I named it "full hash". I will refer to this name meanwhile in explanation of the issue 3 below.

3. The full hash is kept in a long entry field together with the update counter of this field. Sometimes it is required to read/update the entire "field" (full hash + update counter). But sometimes only the full hash needs to be extracted. I do not want to use word "version" here, because it is used for other purposes in Oak context. So I stick to "update counter".

As always the naming issue is the biggest problem in computer science, and the names I started with are not good enough. Let us decide about the names in the meeting. I am not going to fix it in the meanwhile.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: This is related to the naming issue from the above. Skipping.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: I do not get this message. How it can be final if it is updated with `collisionChainLength.incrementAndGet();` ? Not fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: The name will be clarified when we agree about the issues explained above. Skipping.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: Same as above

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: renamed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: I do not see how this change protects us against new entry state being added without updating the switch... Let's discuss

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: done

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: hashNumber = keyHash
hashIndex = index
getKeyHash & getKeyHashAndUpdateCounter

core/src/main/java/com/yahoo/oak/EntryHashSet.java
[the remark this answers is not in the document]
sanastas: Change to print

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: Done

core/src/main/java/com/yahoo/oak/EntryHashSet.java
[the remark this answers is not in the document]
sanastas: done

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: We didn't find how to protect from this feature possible bug. Leaving it as is meanwhile. Can be changed also later.

PR #169, opened 2021-08-22
Add workflow: test coverage compared with last release (non failing)

liran-funaro:

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -9,8 +9,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - name: Fetch repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - run: git fetch origin master
+    - name: Get latest release commit ID
+      id: latest_release
+      run: echo "::set-output name=commit_id::$(git log --pretty=format:"%h%x09%s" -n 1000 | grep "prepare release" | head -1 | cut -f1)"
     - name: Set up JDK 1.8
       uses: actions/setup-java@v1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -9,8 +9,14 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v2
+    - name: Fetch repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - run: git fetch origin master
+    - name: Get latest release commit ID
+      id: latest_release
+      run: echo "::set-output name=commit_id::$(git log --pretty=format:"%h%x09%s" -n 1000 | grep "prepare release" | head -1 | cut -f1)"
```
sanastas, line 19: Does it mean that the comparison is done only relative to the last commit? Does it take into consideration all squashed commits of this PR relative to master?

.github/workflows/maven.yml
```diff
@@ -25,17 +31,25 @@ jobs:
            -Dtest=!com.yahoo.oak.ComputeTest,!com.yahoo.oak.PutIfAbsentTest
     - name: Setup diff-test-coverage
       run: npm install @connectis/diff-test-coverage@1.5.3
-    - name: Validate test-coverage
+    - name: Validate test-coverage compared with master
       run: git diff origin/master HEAD | node_modules/.bin/diff-test-coverage
             --coverage "**/target/site/jacoco/jacoco.xml"
             --type jacoco
             --color
-            --line-coverage 50
```
sanastas, line 33: When lines/branch/function coverage specification is removed are there any default values for that requested?

sanastas (approved):
Just left some questions as a review

.github/workflows/maven.yml
liran-funaro: I think it is 80%, but it doesn't matter since it doesn't fail anything.

.github/workflows/maven.yml
liran-funaro: It will work the same as before.
* Compare the master branch to the last commit of this PR (so it will include all changes of this PR).
* Compare the last released commit to this PR.

PR #170, opened 2021-08-25
minor bug fixes and added benchmarks

li0nr:

### DIFF
--- a/benchmarks/synchrobench/run.sh
+++ b/benchmarks/synchrobench/run.sh
@@ -23,6 +23,11 @@ declare -A scenarios=(
   ["4f-entrySet-descend"]="--buffer -c -a 100"
   ["4f-entryStreamSet-descend"]="--buffer -c -a 100 --stream-iteration"
   ["not-random-put"]="-a 0 -u 100 --inc" #sequential puts, doesn't need to be part of the regression
+  ["put-remove"]="-a 50 -u 100"
+  ["a-put-remove-get"]="-a 25 -u 50"
+  ["b-put-remote-get"]="-a 05 -u 10"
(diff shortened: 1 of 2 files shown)


liran-funaro:
Thank @li0nr.
Can you please provide an explanation of the bugs that are solved by this PR?

li0nr:
Sure,
line 96 is have more precise allocated value, since when reallocating from the free list the original code with missing adding the newly allocated bytes to the variable _**allocated**_ , while it was decreased in the free function.

line 217, is to tackle a highly concurrent allocation consider this scenario a lot of threads want to allocate 
and we are forced to allocate new block since the **_getAndIncrement_** line 216 was before the assignment to the **_blockArray_** 
other threads may allocate from this block but form their perspective the **_blockArray_** still at null, so when those threads 
try to call **_readMemoryAddress_** they will get null.

benchmarks/synchrobench/run.sh
```diff
@@ -23,6 +23,11 @@ declare -A scenarios=(
   ["4f-entrySet-descend"]="--buffer -c -a 100"
   ["4f-entryStreamSet-descend"]="--buffer -c -a 100 --stream-iteration"
   ["not-random-put"]="-a 0 -u 100 --inc" #sequential puts, doesn't need to be part of the regression
+  ["put-remove"]="-a 50 -u 100"
+  ["a-put-remove-get"]="-a 25 -u 50"
+  ["b-put-remote-get"]="-a 05 -u 10"
```
sanastas, line 28: Can we use more meaningful names for benchmarks? 
Like "25Delete50Put25Get", or whatever percentage is used....
Does '-a' means the percentage of deletions?

PR #179, opened 2021-10-26
supporting JDK17

li0nr:

Hey,
I am opening this PR to allow Oak to work with JDK>=8, with minimal effort and without having to change a lot of code.
While still using Unsafe and other sealed packages, though it is not encouraged anymore, and there exists some new alternatives that could replace them.

my changes introduce a new profile to the poms that we in order compile with JDK>= 8 and to run with JDK>=11
two tests fail when ran with JDK17, due to the changes introduced to the JDK past JDK8, which I need your help with (I still think there would be minimal effort to fix them)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,6 +252,14 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+        	<activation>
+			<jdk>[9,)</jdk>
+            </activation>
+            <properties>
+            	<maven.compiler.release>17</maven.compiler.release>
+           	</properties>
+	    </profile>
     </profiles>
 
     <dependencies>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -84,4 +84,29 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <profiles>
+    	<profile>
+    		<activation>
+    			<jdk>[9,)</jdk>
+   			</activation>
+            <build>
+            <plugins>
+            	<plugin>
+            		<artifactId>maven-surefire-plugin</artifactId>
+           			<configuration>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -252,6 +252,14 @@
...
             </build>
         </profile>
+        <profile>
+        	<activation>
+			<jdk>[9,)</jdk>
+            </activation>
+            <properties>
+            	<maven.compiler.release>17</maven.compiler.release>
+           	</properties>
+	    </profile>
```
liran-funaro, line 262: This fails the compilation if you don't have Java 17 installed. Why is that necessary?

technionRamy: it fails if you have something JDK>= 9 and it is not 17.
if you compile it with JDK8 it runs as usual.
my intention in the PR is to make Oak work with JDK 17 and that why I put 17 in line 260.

if you have JDK 11 you may change the line from 17 to 11 and then it will work with JDK 11.
Or if you wanna just run with 11 and compile 8, then we can change the activation value from [9, to [17,
then this effect will take place just when we detect system with JDK>= 17

liran-funaro: So why is it necessary? If you compile on JDK17, then it will compile for Java 17 by default, right?

technionRamy: As I understand it, if this is not present even if we JDK=17 installed 
since the maven source-target is set to 8, it will still compile with 8.
In my changes it will run and compile with 17

liran-funaro: Yes. This is indeed true. I suggest that for now, when compiling for a specific JVM version, simply change JAVA_HOME to the appropriate value.

PR #180, opened 2021-11-02
Generic benchmarks

liran-funaro:
Generic benchmarks and:
- support event-cache key/values in benchmarks
- consume key/values
- quantile latency measurements
- Chronicle and Memcached implementations
- increasing put preserves cardinality
- parallel (faster) initialization
- insert pre-defined scenarios into Parameters.
- nicer parameters usage/print
- removes redundant (old) code
- fix memory leak in the benchmarks


benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/abstractions/BenchOp.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * The operations that 'CompositionalMap' supports
+ */
+public enum BenchOp {
+    PUT,
+    PUT_IF_ABSENT,
+    PUT_IF_ABSENT_COMPUTE_IF_PRESENT,
+    REMOVE,
+    GET,
+    ASCEND,
```
sanastas, line 21: Mention ASCEND and DESCEND is about scan

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/abstractions/BenchOp.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public static final BenchOp[] ADD = asArray(
+        PUT,
+        PUT_IF_ABSENT,
+        PUT_IF_ABSENT_COMPUTE_IF_PRESENT
+    );
+
+    // 'LinkedHashMap' preserves order of insertion.
+    public static final Map<String, BenchOp[]> GROUPS = new LinkedHashMap<>();
```
sanastas, line 40: Where is it used? Please add a comment

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/OpCounter.java
```diff
@@ -0,0 +1,268 @@
+/*
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak.synchrobench.contention.benchmark;
```
sanastas, line 7: This is a general comment:

This is very significant rewrite of the benchmarks structure for Oak project. I find it hard to get into every detail of such a big code update.

1. First concern: For next conference paper, will we be able to say that we evaluate our "something" using a "well known" synchrobench platform?  For example, even Nova for PLDI still runs evaluation using Oak's synchrobench. Nova runs on Ramy's fork so will not be immediately affected by this change. But still what do you think about divergence from synchrobench? Not that it was too similar from the beginning...

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/OpCounter.java
```diff
@@ -0,0 +1,268 @@
...
+
+import com.yahoo.oak.synchrobench.contention.abstractions.BenchOp;
+import org.apache.datasketches.kll.KllFloatsSketch;
```
sanastas, line 10: I continue my general comment:

2. As this is significant write up and code change, that can not be reviewed well enough to be convinced in its correctness, I suggest to make the following correctness checks:

- 2a. Run *all* benchmarks currently existing on master to see they do not fail or give some weird results after the change.

- 2b. For all benchmarks currently existing on master, run them in their previous version (master) and with this code and verify that the results are similar.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/benchmark/OpCounter.java
```diff
@@ -0,0 +1,268 @@
...
+
+import java.util.Map;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
```
sanastas, line 14: Again continue general discussion:

This is again very big change to the benchmarks that also serve as our basic performance regression test. We need to verify that our "old" regression still performs the same with the same environment. Plus, I am up to adding 2-3 new regression benchmarks with the new abilities that gives this significant change.

sanastas (commented):
I didn't finish. Here are my thoughts so far. Will continue tomorrow.

benchmarks/synchrobench/run.sh
```diff
@@ -12,40 +12,46 @@ function ctrl_c() {
...
 
 declare -A benchmarks=(
-  ["skip-list"]="JavaSkipListMap"
-  ["oak"]="OakMyBufferMap"
-  ["offheap-list"]="OffHeapList"
-  ["concurrent-hash-map"]="JavaHashMap"
-  ["oak-hash"]="OakMyBufferHash"
+  ["skip-list"]="com.yahoo.oak.JavaSkipListMap"
+  ["oak"]="com.yahoo.oak.OakBenchMap"
+  ["offheap-list"]="com.yahoo.oak.OffHeapList"
```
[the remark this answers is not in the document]
liran-funaro: See `RunTest`. It runs all the benchmarks on a small scale to verify they work properly. I actually found many issues with the benchmarks that way (issues that were there before).

benchmarks/synchrobench/run.sh
```diff
@@ -120,11 +130,17 @@ warmup="0"
...
 
 # Defines the test runtime in milliseconds.
-duration="30000"
+duration="30_000"
 
 # Defines the sampling range for queries and insertions.
 range_ratio="2"
 
+# Defines the number of threads used for initialization
+fill_threads="24"
```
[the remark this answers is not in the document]
liran-funaro: If hyperthreading wasn't beneficial, it wouldn't have been exited :)
I noticed significant speedup with 24 threads compared with 12.

benchmarks/synchrobench/run.sh
```diff
@@ -133,42 +149,66 @@ range_ratio="2"
...
+  --value ) value_class="$2"; shift 2 ;;
+  --key-size ) key_size="$2"; shift 2 ;;
+  --value-size ) value_size="$2"; shift 2 ;;
+  --consume-keys ) extra_args="$extra_args --consume-keys"; shift ;;
+  --consume-values ) extra_args="$extra_args --consume-values"; shift ;;
+  --fill-threads ) fill_threads="$2"; shift 2 ;;
+  --latency ) extra_args="$extra_args --latency"; shift ;;
+  -v | --verify ) verify_script=1; shift ;;
+  -- ) shift; break ;;
+  * ) break ;;
```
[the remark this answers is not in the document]
liran-funaro: I'll try to create some kind of unit test for that.

benchmarks/synchrobench/run.sh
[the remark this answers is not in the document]
sanastas: Very good!

benchmarks/synchrobench/run.sh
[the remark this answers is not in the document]
sanastas: Then let us change `test_thread="01 04 08 12 16 20 24 28 32"` to `test_thread="01 04 08 12 16 20 24"`
32 threads were used only on Amazon server that we usually do not use

benchmarks/synchrobench/run.sh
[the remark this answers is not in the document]
sanastas: 👍

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/data/eventcache/Key.java
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak.synchrobench.data.eventcache;
+
+import com.yahoo.oak.synchrobench.contention.abstractions.BenchKey;
+
+public class Key implements BenchKey {
```
sanastas, line 11: Why Key has 2 keys inside? In general, I am lost among so many classes... Please add comment about each class what is the purpose of each.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/data/eventcache/ValueGen.java
```diff
@@ -0,0 +1,160 @@
...
+import net.openhft.chronicle.bytes.Bytes;
+import net.spy.memcached.CachedData;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.openjdk.jmh.infra.Blackhole;
+import sun.misc.Unsafe;
+
+import java.util.Random;
+
+public class ValueGen implements ValueGenerator {
```
sanastas, line 24: Why both `ValueGen` and `ValueGenerator`? Why so many value offsets? Seems like comments per classes are very missed

benchmarks/synchrobench/src/test/java/com/yahoo/oak/RunTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+@RunWith(Parameterized.class)
+public class RunTest {
```
sanastas, line 20: Maybe to name it `RunBenchmarksTest`?

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -17,7 +17,7 @@
  */
 final class BlocksPool implements BlocksProvider, Closeable {
 
-    private static BlocksPool instance = null;
+    private static volatile BlocksPool instance = null;
```
sanastas, line 20: What is the purpose of the change? BlocksPool should be written only once and single threaded. After accesses can be from caches...

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
```diff
@@ -162,6 +162,7 @@ public void close() {
         // Release the hold of the block array and return it the provider.
         Block[] b = blocksArray;
         blocksArray = null;
+        freeList.clear();
```
sanastas, line 165: What is the purpose of the change? To release the on-heap structure? Should be cleared when entire allocator is disconnected with the OakMap/Hash

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
```diff
@@ -19,114 +23,124 @@
...
         };
+    }
 
+    @Override
+    public void init() {
         skipListMap = new ConcurrentSkipListMap<>(comparator);
-        allocator = new NativeMemoryAllocator(OAK_MAX_OFF_MEMORY);
-        mm = new SyncRecycleMemoryManager(allocator);
+        allocator = new NativeMemoryAllocator(MAX_OFF_MEMORY);
+        mm = new SeqExpandMemoryManager(allocator);
```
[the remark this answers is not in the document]
liran-funaro: `OffHeapList` didn't really use the MM. It just uses it to generate the slice. I changed it because `OffHeapList` doesn't use the header anyway, and `SeqExpandMemoryManager` doesn't have a header.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/abstractions/BenchOp.java
liran-funaro: :+1:

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/contention/abstractions/BenchOp.java
liran-funaro: :+1:

benchmarks/synchrobench/src/main/java/com/yahoo/oak/Memcached.java
```diff
@@ -0,0 +1,122 @@
...
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Required installation of Memcached on the machine.
+ */
+public class Memcached extends BenchMap {
```
sanastas, line 27: Where exactly this benchmark is used? I mean in `run.h` I do not see it as part of benchmarks, while chronicle is there. Is it for future use? Why not to add a possibility to run.h?

core/src/main/java/com/yahoo/oak/UnsafeUtils.java
```diff
@@ -13,9 +13,9 @@
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
 
-final class UnsafeUtils {
+public final class UnsafeUtils {
```
sanastas, line 16: Is it instead of PR#159? How PR#159 is related?

pom.xml
```diff
@@ -44,47 +43,44 @@
     <developers>
         <developer>
             <name>Anastasia Braginsky</name>
-            <email>sanastas@oath.com</email>
+            <email>sanastas@yahooinc.com</email>
```
sanastas, line 46: May be to remove people who are not in the company?

sanastas:
My biggest concern now is that we need a design document together with this change. If myself or other committers will need to change something in the way benchmarks are done, it would be hard to do without some guidance.

core/src/main/java/com/yahoo/oak/BlocksPool.java
liran-funaro: Actually, it was IntelliJ suggestion that I accepted. But you are right. I will revert this change.

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
liran-funaro: I found this issue when I was trying the source of the memory leak in the benchmarks.
Technically, when the allocator will be collected by the GC, the `freeList` will also be collected and all its entries as well.
But it seems "more correct" to clear the list to "release" all the slices.
I'm not sure if that list is clear before. Where do we call "clear" before we call "close"? These methods are very similar.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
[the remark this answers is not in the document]
sanastas: Then how the compute works? `ScopedWriteBuffer.compute(prevValue.value.get().getSlice(), computeFunction);` Locks of MM are never used?

If so please add a comment about it

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
sanastas: So this is in the hope that the garbage objects will be collected faster?

At least the comment that is couple of lines below, needs to be updated.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/Memcached.java
liran-funaro: I will add it to `run.sh`, but I can't include it in the unit testing because it will require starting a Memcached service.
The reason I didn't add it was because the performance was not even at the same scale as the others.
But it might have been a configuration issue.

core/src/main/java/com/yahoo/oak/UnsafeUtils.java
liran-funaro: It is used for decoding/encoding the data to a byte array, needed by Memcached transcoder.

pom.xml
liran-funaro: I'm not sure what is common in this case. To remove non-active developers, or leave them for the credit.

benchmarks/synchrobench/src/test/java/com/yahoo/oak/RunTest.java
liran-funaro: :+1:

core/src/main/java/com/yahoo/oak/UnsafeUtils.java
liran-funaro: I removed the dependency on UnsafeUtils.

benchmarks/synchrobench/run.sh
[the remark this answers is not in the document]
liran-funaro: Done.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/data/eventcache/Key.java
liran-funaro: These are fields. I changed them to `field1` and `field2`. Also in `Value`.
I'm not sure what is the comment that you are looking for.
These are just arbitrary numbers. They have no meaning apart from resembling a real use case that we know of.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/data/eventcache/ValueGen.java
liran-funaro: `ValueGenerator` is the interface. I can't name the class and interface with the same name.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OffHeapList.java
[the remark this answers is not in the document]
liran-funaro: compute/serialize or any of the writing functions does not use the locks since we only write in a scoped session (i.e., `ScopedWriteBuffer`). The lock is obtained beforehand by the map.
`OffHeapList` uses a synchronized block to avoid data races.
I'll add a comment mentioning why we use `SeqExpandMemoryManager` in this context.

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
liran-funaro: :+1:

liran-funaro:
> My biggest concern now is that we need a design document together with this change. If myself or other committers will need to change something in the way benchmarks are done, it would be hard to do without some guidance.

I can add a readme file to the synchrobench module.
What would you like to be documented?
Note that the benchmark class usage is documented. You just need to execute it with `-h`/`--help` flag.

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
```diff
@@ -52,13 +52,18 @@
...
+        this.capacity = capacity;
+        this.blocksArray = new Block[calcBlockArraySize()];
         // initially allocate one single block from pool
         // this may lazy initialize the pool and take time if this is the first call for the pool
         allocateNewCurrentBlock();
-        this.capacity = capacity;
+
+    }
+
+    private int calcBlockArraySize() {
```
sanastas, line 63: Where is it used?

benchmarks/synchrobench/run.sh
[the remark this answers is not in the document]
liran-funaro: done.

benchmarks/synchrobench/run.sh
```diff
@@ -52,15 +53,19 @@ declare -A heap_limit=(
   ["concurrent-hash-map"]="28g"
   ["oak-hash"]="24g"
   ["chronicle"]="24g"
+  # Memcached doesn't use the heap
```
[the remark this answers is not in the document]
liran-funaro: Changed to 512m. Anyway, it is just the maximum heap size. The will JVM optimize this size.

core/src/main/java/com/yahoo/oak/NativeMemoryAllocator.java
liran-funaro: I noticed a bug: `clear()` does not re-allocate the block array. So I corrected the code. But since this allocation happens in `clear()` and in the constructor, I added this function to avoid code duplication.

PR #181, opened 2021-11-09
Making a special search for gets only

sanastas:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/HashChunk.java
+++ b/core/src/main/java/com/yahoo/oak/HashChunk.java
@@ -182,8 +182,7 @@ class HashChunk<K, V> extends BasicChunk<K, V> {
 
     /**
      * Look up a key in this chunk.
-     *
-     * @param ctx The context that follows the operation following this key look up.
+     *  @param ctx The context that follows the operation following this key look up.
      *            It will describe the state of the entry (key and value) associated with the input {@code key}.
      *            Following are the possible states of the entry:
      *             (1) {@code key} was not found.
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -200,11 +199,16 @@ int compareKeyAndEntryIndex(KeyBuffer tempKeyBuff, K key, int ei) {
...
-    void lookUp(ThreadContext ctx, K key) {
+    void lookUp(ThreadContext ctx, K key, boolean forGetsOnly) {
         int keyHash = calculateKeyHash(key, ctx);
         int idx = calculateEntryIdx(key, keyHash);
-        entryHashSet.lookUp(ctx, key, idx, keyHash);
+        if (forGetsOnly) {
+            entryHashSet.lookUpForGetOnly(ctx, key, idx, keyHash);
+        } else {
+            entryHashSet.lookUp(ctx, key, idx, keyHash);
+        }
```
liran-funaro, line 211: IMO, it is better to add another function: `HashChunk.lookUpForGetOnly()` rather then add a boolean parameter.
This way, you'll only have to change the places that actually use this method, instead of all the calls to `lookup()`.

liran-funaro (approved):
Great. I think you can merge. Or you can do the change I suggested before. Any way is fine by me.

core/src/main/java/com/yahoo/oak/HashChunk.java
sanastas: Good idea, although `HashChunk.lookUpForGetOnly()` and `HashChunk.lookUp()` are going to be very similar... But fixed anyway

PR #182, opened 2021-11-10
Added iterator support to HashChunk

boris-kap:


core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -19,6 +19,7 @@
         RELEASED
     }
 
+    static final int NONE_NEXT = EntryArray.INVALID_ENTRY_INDEX;
```
sanastas, line 22: I see why `NONE_NEXT` was moved here. Because `BasicChunkIter.next(ThreadContext ctx)` should return it when there is no more entries left to return.

However, `NONE_NEXT` is more part of OrderedChunk, because there we have linked list, with "nexts" in each entry. `NONE_NEXT` is not used in HashChunk. Can something else be invented?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -240,13 +240,26 @@ private EntryState getEntryState(ThreadContext ctx, int idx, K key, int keyHash)
...
 
+    /**
+     * Function checks if an entry at given index is valid.
+     *
+     * @param ctx the context used as temporal storage, used for convenience
+     * @param idx index of the entry to check
+     * @return true if the entry at the given index is valid, false otherwise
+     */
+    public boolean isEntryIdxValid(ThreadContext ctx, int idx) {
+        // if
```
sanastas, line 258: Is it a needed comment?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -240,13 +240,26 @@ private EntryState getEntryState(ThreadContext ctx, int idx, K key, int keyHash)
...
-        // value is invalid
+        // value is invalid, in case another thread inserts the same key
         if (isKeyAndEntryKeyEqual(ctx.key, key, idx, keyHash)) {
             return EntryState.INSERT_NOT_FINALIZED;
         }
         return EntryState.VALID;
     }
 
+    /**
+     * Function checks if an entry at given index is valid.
```
sanastas, line 251: Entry at given index is valid "from scan perspective". It is maybe "valid enough" for scan, but not for some other operation... Please think about more precise name and function explanation.

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -274,7 +274,7 @@ boolean finalizeDeletion(ThreadContext ctx) {
 
 
     /********************************************************************************************/
-    /*------------------------- Methods that are used for rebalance  ---------------------------*/
+    /*------------------------- Methods that are used for rebalanced  ---------------------------*/
```
sanastas, line 277: This is indeed "rebalance", a noun that we invented, although not understood by spell checker :)

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -307,6 +307,64 @@ void printSummaryDebug() {
 
     /********************************************************************************************/
     /*--------------------------------- Iterators Constructors ---------------------------------*/
-    // TODO: update hash iterator later
+    /********************************************************************************************/
```
sanastas, line 310: Why the same comment twice?

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -307,6 +307,64 @@ void printSummaryDebug() {
...
+
 
+    /********************************************************************************************/
+
+    class ChunkIter extends BasicChunkIter {
+        protected int next;         // index of the next entry to be returned
+
+        ChunkIter(ThreadContext ctx) {
+            next = 0; // initial index
+
```
sanastas, line 328: please remove spaces

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -307,6 +307,64 @@ void printSummaryDebug() {
...
+
+            next = getFstVldEntryIdx(ctx);
+        }
+        int getFstVldEntryIdx(ThreadContext ctx) {
```
sanastas, line 332: new line

sanastas: I mean a vertical space between methods

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -307,6 +307,64 @@ void printSummaryDebug() {
...
+
+        @Override
+        boolean hasNext() {
+            return next != NONE_NEXT;
+        }
+
+        /** Returns the index of the entry that should be returned next by the iterator.
+         ** NONE_NEXT is returned when iterator came to its end.
+         **/
+        int next(ThreadContext ctx) {
```
sanastas, line 361: Is it also @Override? Didn't you get code style checker errors?

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -634,6 +634,9 @@ boolean refreshValuePosition(ThreadContext ctx) {
         return ctx.isValueValid();
     }
 
+    public MemoryManager getValuesMemoryManager() {
```
sanastas, line 637: Why this change is needed?

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -1166,12 +1169,12 @@ private void advanceState() {
 
     class ValueIterator extends Iter<OakUnscopedBuffer> {
 
-        private final InternalOakMap<K, V> internalOakMap;
+        //private final InternalOakMap<K, V> internalOakMap;
```
sanastas, line 1172: If this code is not needed, please just remove it

sanastas (commented):
Looks good. Please fix the comments below.

core/src/main/java/com/yahoo/oak/BasicChunk.java
boris-kap: I guess I can just use EntryArray.INVALID_ENTRY_INDEX. Seems to fit the purpose

core/src/main/java/com/yahoo/oak/InternalOakMap.java
boris-kap: Yes, since I remove the memory manager reference from the OakMap, for more clarity.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
boris-kap: if the readValue succeeds, the entry is valid from any perspective, isn't it?

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -277,4 +278,14 @@ void decrementAddedCount() {
...
+
+    /********************************************************************************************/
+    abstract class BasicChunkIter {
+        abstract boolean hasNext();
+
+        /** Returns the index of the entry that should be returned next by the iterator.
+         ** NONE_NEXT is returned when iterator came to its end.
+         **/
+        abstract int next(ThreadContext ctx);
+    }
```
liran-funaro, line 290: This should be an interface.

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -319,6 +319,64 @@ void printSummaryDebug() {
...
-    // TODO: update hash iterator later
+    /********************************************************************************************/
+    /*--------------------------------- Iterators Constructors ---------------------------------*/
+
+    /**
+
+     */
+    HashChunk.ChunkIter hashChunkIter(ThreadContext ctx) {
+        return new HashChunk.ChunkIter(ctx);
+    }
```
liran-funaro, line 330: Intellij: Method 'hashChunkIter(com.yahoo.oak.ThreadContext)' is never used 
Intellij: Raw use of parameterized class 'HashChunk' 
Please check Intellij errors/warnings before submitting a PR. If you are not using Intellij, I suggest using it.

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -634,6 +634,9 @@ boolean refreshValuePosition(ThreadContext ctx) {
         return ctx.isValueValid();
     }
 
+    public MemoryManager getValuesMemoryManager() {
+        return this.valuesMemoryManager;
+    }
```
liran-funaro, line 639: Add an extra line between methods.

liran-funaro (changes requested):
Initial review (I'll continue tomorrow)

core/src/main/java/com/yahoo/oak/BasicChunk.java
boris-kap: in this case the methods become public. Is it OK?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: If there is no concurrent deletion -- yes. Otherwise deletion of this entry can already been started.

core/src/main/java/com/yahoo/oak/InternalOakMap.java
sanastas: OK

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -112,7 +112,7 @@ protected void clear() {
     // Entry marked deleted only off-heap is already considered deleted
     enum EntryState {
         /*
-         * The state of the entry is yet to be checked.
+         * The state of the entry is yet to be checked or Entry is empty
```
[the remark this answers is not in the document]
sanastas: I agree. There can be UNKNOWN for "not checked at all", and EMPTY for "checked, but nothing was there"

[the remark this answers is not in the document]
sanastas: Boris, can you please add the state EMPTY, with TODO comment. While the usage of this state will be differed to other PR. This way there are more chances this will not be forgotten.

sanastas (commented):
I am not continuing forward as there is enough problems so far. I will set us a meeting to discuss. But please refer to the comments I already wrote. Thanks!

core/src/main/java/com/yahoo/oak/FirstLevelHashArray.java
```diff
@@ -106,24 +106,43 @@ private int calculateChunkSize(int inputLsbForSecondLevel) {
         return (int) Math.ceil(Math.pow(2, lsbForSecondLevel));
     }
 
-    private int calculateKeyHash(K key, ThreadContext ctx) {
```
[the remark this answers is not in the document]
boris-kap: it was removed because because it was not used, and it seemed to me very misleading to have a function named "calculateKeyHash" that do not calculate key hash

core/src/main/java/com/yahoo/oak/FirstLevelHashArray.java
```diff
@@ -106,24 +106,43 @@ private int calculateChunkSize(int inputLsbForSecondLevel) {
...
-    private int calculateHashArrayIdx(K key, ThreadContext ctx) {
+    private int calculateHashArrayIdx(int keyHash) {
         // second and not first, because these are actually the most significant bits
-        return hashIndexCodec.getSecond(calculateKeyHash(key, ctx));
+        return hashIndexCodec.getSecond(keyHash);
     }
 
-    HashChunk<K, V> findChunk(K key, ThreadContext ctx, int keyHash) {
-        return chunks.get(calculateHashArrayIdx(key, ctx));
+    HashChunk<K, V> findChunk(int keyHash) {
```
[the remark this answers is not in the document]
boris-kap: thread context is not used in this function, chunk is found based on the keyHash, which is given to the function anyway

core/src/main/java/com/yahoo/oak/FirstLevelHashArray.java
```diff
@@ -106,24 +106,43 @@ private int calculateChunkSize(int inputLsbForSecondLevel) {
...
     HashChunk<K, V> getChunk(int index) {
         assert 0 <= index && index < chunks.length();
         return chunks.get(index);
     }
 
+    HashChunk<K, V> getNextChunk(HashChunk<K, V> curChunk) {
+        int idx = 0;
+
+        // find the index of the current chunk
+        while (idx < chunks.length() && curChunk != getChunk(idx)) {
```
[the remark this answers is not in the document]
boris-kap: I've considered it' but it seemed to me a bit unclean to keep the chunk index. What happens if the chunks are rebalanced and we keep the index and use it to access the chunk?
There number of chunks is relatively small, so probably the iteration over all the chunks every time we a looking for the next one is not as bad as it seems.

core/src/main/java/com/yahoo/oak/FirstLevelHashArray.java
```diff
@@ -106,24 +106,43 @@ private int calculateChunkSize(int inputLsbForSecondLevel) {
...
+            idx++;
+        }
+        // should never happen, but just to be on the safe side
+        if (! (idx < chunks.length())) {
+            return null;
```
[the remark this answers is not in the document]
boris-kap: That's an interesting design question. If the chunks are rebalanced and an old chunk is replaced with a new one, do we want the application to failt on assert, or is ti good enough to invalidate the iterator that already running?

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -110,7 +115,7 @@ V put(K key, V value, OakTransformer<V> transformer) {
         for (int i = 0; i < MAX_RETRIES; i++) {
 
             // find chunk matching key, puts this key hash into ctx.operationKeyHash
-            HashChunk<K, V> c = hashArray.findChunk(key, ctx, calculateKeyHash(key, ctx));
+            HashChunk<K, V> c = hashArray.findChunk(calculateKeyHash(key, ctx));
```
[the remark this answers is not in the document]
boris-kap: the key and context are not used to find the chunk index, so I removed them to have clearer code.

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -517,6 +526,421 @@ boolean putIfAbsentComputeIfPresent(K key, V value, Consumer<OakScopedWriteBuffe
...
+        }
+
+        private void initAfterRebalance() {
+            //TODO - refactor to use OakReadBuffer without deserializing.
+            state.getHashChunk().readKeyFromEntryIndex(ctx.tempKey, state.getIndex());
+            K nextKey = keySerializer.deserialize(ctx.tempKey);
+
+
+            // Update the state to point to last returned key.
+            initState();
```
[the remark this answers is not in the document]
boris-kap: Could be a bug, the rebalance is not implemented, so I skipped the parts related to rebalance.

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -517,6 +526,421 @@ boolean putIfAbsentComputeIfPresent(K key, V value, Consumer<OakScopedWriteBuffe
...
+
+                // build the entry context that sets key references and does not check for value validity.
+                ctx.initEntryContext(curIndex);
+
+
+                chunk.readKey(ctx);
+
+                validState = ctx.isKeyValid();
+
+                assert validState;
```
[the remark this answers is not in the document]
boris-kap: in OakMap a key can not be deleted?

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -517,6 +526,421 @@ boolean putIfAbsentComputeIfPresent(K key, V value, Consumer<OakScopedWriteBuffe
     void printSummaryDebug() {
         hashArray.printSummaryDebug();
     }
+
+    //////////////////////////////////////////////////////////////////////////////
+    /*-------------- Iterators --------------*/
+    //////////////////////////////////////////////////////////////////////////////
+
+    private static final class IteratorState<K, V> {
```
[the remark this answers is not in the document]
boris-kap: not exactly the same, newInstance has different signature. There is some similar code, but to make it inherit from the same class we will have to make it parameterised or typecast the return value.

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -307,6 +311,11 @@ OakUnscopedBuffer get(K key) {
         return transformer.apply(ctx.key);
     }
 
+    private <T> T getValueTransformation(OakScopedReadBuffer key, OakTransformer<T> transformer) {
```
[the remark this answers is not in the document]
boris-kap: Actually, I'm not sure why we need another getValueTransformation  function, I've copied the code from the InternalOakMap to have it the same for both of them.

liran-funaro (commented):
Initial comments

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -171,6 +173,139 @@ protected UnscopedValueBufferSynced getValueUnscopedBuffer(ThreadContext ctx) {
...
+        void advance(boolean needsValue) {
+            boolean validState = false;
+
+            while (!validState) {
+                if (state == null) {
+                    throw new NoSuchElementException();
+                }
+
+                final BasicChunk<K, V> chunk = state.getChunk();
+                if (chunk.state() == BasicChunk.State.RELEASED) {
```
sanastas, line 262: Please add a TODO comment not to access the keys on the RELEASED chunk once the key might be released

core/src/main/java/com/yahoo/oak/InternalOakHash.java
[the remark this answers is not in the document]
sanastas: Then it looks like it need to go to InternalOakBasics and only once

core/src/main/java/com/yahoo/oak/OrderedChunk.java
```diff
@@ -672,20 +672,20 @@ DescendingIter descendingIter(ThreadContext ctx, K from, boolean fromInclusive,
...
+    abstract class ChunkIter implements BasicChunkIter {
         protected int next;         // index of the next entry to be returned
         protected K endBound;       // stop bound key, or null if no stop bound
         protected boolean endBoundInclusive;  // inclusion flag for "to"
 
         protected IterEndBoundCheck isEndBoundCheckNeeded = IterEndBoundCheck.NEVER_END_BOUNDARY_CHECK;
         protected int midIdx = sortedCount.get() / 2; // approximately index of the middle key in the chunk
 
-        abstract boolean hasNext();
+        //abstract boolean hasNext();
```
sanastas, line 683: Please remove dead code if unneeded

core/src/test/java/com/yahoo/oak/ParametrizedMapApiTest.java
```diff
@@ -216,29 +216,45 @@ public void keySet() {
...
+
     }
 
     @Test
     public void entrySet() {
-        if (oak instanceof OakHashMap) {
-            // TODO: currently iterators are not supported for Hash, remove this later
-            return;
-        }
+        //if (oak instanceof OakHashMap) {
```
sanastas, line 254: Please remove this dead code

sanastas (approved):

core/pom.xml
```diff
@@ -32,8 +32,8 @@
             <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
-                    <source>${java.version}</source>
-                    <target>${java.version}</target>
+                    <source>8</source>
```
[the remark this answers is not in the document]
boris-kap: No reason, must be some automatic change by intelliJ or some other tool

PR #186, opened 2021-11-28
Off-heap (direct) entry array

liran-funaro:
* Refactor chunks/entry-array: put only the array access logic into `EntryArray` and the rest of the logic into `Chunk`.
* Add support for both `EntryArrayHeap` and `EntryArrayDirect` implementing `EntryArray` interface.
* Add `OakSharedConfig` to pass Oak configuration between instances of `InternalOak` and `Chunks`.
* Fix memory-leak in unit tests


core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+/*
+* EntryArray is array of entries based on primitive array of longs.
```
sanastas, line 13: It is weird that class BasicChunk starts with comment about EntryArray. At least BasicChunk should be explained in few words, saying that BasicChunk has an EntryArray, then about EntryArray.

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
...
-    /*-------------- Members --------------*/
-    // to compare serilized and object keys
-    protected OakComparator<K> comparator;
     // in split/compact process, represents parent of split (can be null!)
     private final AtomicReference<BasicChunk<K, V>> creator;
     // chunk can be in the following states: normal, frozen or infant(has a creator)
     private final AtomicReference<State> state;
     private final AtomicReference<Rebalancer<K, V>> rebalancer;
     private final AtomicInteger pendingOps;
-    private final int maxItems;
```
sanastas, line 31: Why maxItems is removed?

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
...
         this.pendingOps = new AtomicInteger();
         this.rebalancer = new AtomicReference<>(null); // to be updated on rebalance
         this.statistics = new Statistics();
     }
 
+    // EntryState is the state of the entry after inspection of the states of its key and value
+    // For Example:
+    // Entry in which a key is allocated, but without a valid value, is considered not existing
+    // Entry marked deleted only off-heap is already considered deleted
+    enum EntryState {
```
sanastas, line 124: It is really weird to me that EntryState is part of the chunk. If so, I would move it to a separate enum class. And please add EMPTY state, even though for now it is not going to be used.

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
...
+         * @return is the value valid
+         */
+        boolean isValid() {
+            return this.ordinal() >= EntryState.VALID.ordinal();
+        }
+    }
+
+    /**
+     * Returns the number of entries allocated and not deleted for this EntryArray instance.
+     * Although, in case instance is used as an linked list, nextFreeIndex is can be used to calculate
```
sanastas, line 166: `nextFreeIndex` is unrelated (in the comment)

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
...
+    /**
+     * Completes the deletion of a mapping
+     * @param ctx the call context
+     * @return true if the deletion indeed updated the entry to be deleted as a unique operation
+     */
+    abstract boolean deleteValueFinish(ThreadContext ctx);
+
+
+    /* ########################################################################
+       # Former Chunk
```
sanastas, line 480: What does it mean "Former Chunk"?

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -154,61 +602,78 @@ void release() {
...
-    /**
-     * See concrete implementation for more information
-     */
-    abstract void releaseNewValue(ThreadContext ctx);
 
     /**
      * Writes the key off-heap and allocates an entry with the reference pointing to the given key
      * See concrete implementation for more information
      */
     abstract boolean allocateEntryAndWriteKey(ThreadContext ctx, K key);
```
sanastas, line 618: `allocateEntryAndWriteKey` method can move up with all of the rest similar methods...

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -154,61 +602,78 @@ void release() {
...
+    }
 
     /**
-     * When deleting an entry the first step is marking (CAS) the delete bit in the value's off-heap
-     * header, this is the Linearization Point (LP) of deletion.
-     * Later series of steps need to be performed to accomplish the entry deletion.
-     * This method accomplish the rest of the algorithm coming after LP.
-     * Other threads seeing a marked value call this function before they proceed (e.g.,
-     * before performing a successful {@code putIfAbsent()}).
+     * As written in {@link BasicChunk#writeValueCommit(ThreadContext)}, when changing an entry,
```
sanastas, line 645: Is `writeValueCommit()` part of the BasicChunk?

core/src/main/java/com/yahoo/oak/EntryArrayHeap.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public int fieldCount() {
+        return entryCount;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void clear() {
```
sanastas, line 59: Is the clear() usage is exercised as part of the tests?

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -9,55 +9,503 @@
...
+        assert valuesMemoryManager.isReferenceConsistent(ctx.value.getSlice().getReference());
+    }
+
+    /**
+     * Find the state of a the value that is pointed by {@code value}.
+     * Thus, {@code readValue(value, ei)} should be called prior to this method with the same {@code value} instance.
+     *
+     * @param value a buffer object that contains the value buffer
+     */
+    private EntryState getValueState(ValueBuffer value) {
```
sanastas, line 362: It is weird we have both `getEntryState()` and `getValueState()` in one file. They are doing basically the same just one for Hash's algorithm insertion/deletion and the later for Map's algorithm of insertion/deleted... Not experienced coder will get confused...

sanastas (commented):
I didn't finish, I have contradicting feelings about the change of putting all the entry sets logic in BasicChunk. Because we have no good separation between Hash logic and Map logic... But I will continue review.

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -97,6 +97,16 @@ static void setBlockSize(int blockSize) {
         }
     }
 
+    static void clear() {
```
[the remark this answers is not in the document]
liran-funaro: Before, I've noticed that the memory usage during testing is too high. I've fixed some tests to correctly close Oak. But I also wanted to clear all the used blocks to release the memory.
This function is used for testing to release all the blocks but still leave the pool operational.
I think that we should expose this API to the users due to Issue #192 .

core/src/main/java/com/yahoo/oak/BasicChunk.java
liran-funaro: Because it can be derived from the entry array capacity.

core/src/main/java/com/yahoo/oak/BasicChunk.java
liran-funaro: I don't think it is weird. A chunk consists of an entry array and it wants to describe the state of an entry inside the chunk. It is very specific to chunks so it makes sense to be part of this class.

core/src/main/java/com/yahoo/oak/BasicChunk.java
liran-funaro: This is a comment for the reviewer and later will be removed. Before this PR, the methods below it were implemented in Chunk, and the methods above it were implemented in EntryArray.

core/src/main/java/com/yahoo/oak/EntryArrayHeap.java
liran-funaro: I don't remember where it is used. But this method was part of EntryArray before this PR.

core/src/main/java/com/yahoo/oak/BasicChunk.java
liran-funaro: You're right. These should be moved to the specific Chunk implementation.

liran-funaro:
@sanastas 
> we have no good separation between Hash logic and Map logic

I think it is possible to attain this separation with not too much effort. If you find methods that should be moved like the ones you already mentioned, then let me know.
Anyway, I won't start applying the changes you requested and merging with the master until we feel confident that we want to go ahead with this change.

core/src/main/java/com/yahoo/oak/BasicChunk.java
sanastas: Then it is better to keep `getMaxItems()` method and redirect to array capacity

core/src/main/java/com/yahoo/oak/BasicChunk.java
sanastas: I am suggesting to change the comment and to make it clearer...

core/src/main/java/com/yahoo/oak/BasicChunk.java
sanastas: OK

sanastas:
I want to ask @boris-kap to make the review to this PR as well, so we have more then one opinion.

core/src/main/java/com/yahoo/oak/BlocksPool.java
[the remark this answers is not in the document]
sanastas: My understanding of `clear()` (e.g. vs `close()`) is that it makes the data structure "clear", but keeps the structure so it can be "filled" again. Thus the entire memory is not released. Thus issue #192 requires another API to resolve the user request there. Let's discuss.

liran-funaro:
Broke down into number of PRs and rethink chunk/array design

PR #187, opened 2021-11-29
Consist entry

sanastas:
1. Ensuring the get state of the entry for lookUps is consistent
2. Rewriting the process of getting the state of the entry for inserts
3. Benchmarks' README clarification

### DIFF
--- a/core/src/main/java/com/yahoo/oak/EntryHashSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryHashSet.java
@@ -8,7 +8,7 @@ package com.yahoo.oak;
 
 import java.util.concurrent.atomic.AtomicInteger;
 
-/* EntryHashSet keeps a set of entries placed according to the key hash.
+/* EntryHashSet keeps a set of entries placed according to the key hash number modulo capacity.
  * Entry is reference to key and value, both located off-heap.
  * EntryHashSet provides access, updates and manipulation on each entry, provided its entry index.
  *
--- a/core/src/main/java/com/yahoo/oak/EntryHashSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryHashSet.java
@@ -266,6 +266,19 @@ class EntryHashSet<K, V> extends EntryArray<K, V> {
         return readValue(ctx.tempValue, idx);
     }
 
+    /**
+     * Function checks if an entry at given index is valid for the scan purpose
+     *
+     * @param ctx the context used as temporal storage, used for convenience
+     * @param idx index of the entry to check
+     * @return true if the entry at the given index is valid, false otherwise
+     */
+    public boolean isEntryIndexValidForScan(ThreadContext ctx, int idx) {
+        // tempValue us used rather than value, since as a side effect value is updated by the readValue operation
+        // it may replace the value that is already read by the `advance` iterator method
+        return readValue(ctx.tempValue, idx);
+    }
+
     int getCollisionChainLength() {
         return collisionChainLength.get();
     }
--- a/core/src/main/java/com/yahoo/oak/EntryHashSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryHashSet.java
@@ -216,6 +216,7 @@ class EntryHashSet<K, V> extends EntryArray<K, V> {
         if (getKeyReference(idx) == keysMemoryManager.getInvalidReference()) {
             return EntryState.UNKNOWN;
         }
+
         // try to get consistent read of key, value, hash code, its update counter and valid bit
         do {
             ctx.keyHashAndUpdateCnt = getKeyHashAndUpdateCounter(idx);
--- a/core/src/main/java/com/yahoo/oak/EntryHashSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryHashSet.java
@@ -56,7 +56,7 @@ class EntryHashSet<K, V> extends EntryArray<K, V> {
 
     static final int DEFAULT_COLLISION_CHAIN_LENGTH = 4;
     // HASH - the key hash of this entry (long includes the update counter)
-    // hey hash is first, update counter is second, valid bit is third
+    // key hash is first, update counter is second, valid bit is third
     private static final int HASH_FIELD_OFFSET = 2;
     private static final int KEY_HASH_BITS = 32; // Hash needs to be an integer
     private static final int UPDATE_COUNTER_BITS = 31;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -56,6 +56,7 @@
 
     static final int DEFAULT_COLLISION_CHAIN_LENGTH = 4;
     // HASH - the key hash of this entry (long includes the update counter)
+    // hey hash is first, update counter is second, valid bit is third
```
liran-funaro, line 59:
```suggestion
    // key hash is first, update counter is second, valid bit is third
```

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -196,55 +207,48 @@ private boolean isKeyAndEntryKeyEqual(KeyBuffer tempKeyBuff, K key, int idx, int
...
-                    if (isKeyHashValid(idx)) {
-                        return EntryState.DELETED_NOT_FINALIZED;
+        // try to get constant read of key, value, hash code, its update counter and valid bit
+        do {
+            ctx.keyHashAndUpdateCnt = getKeyHashAndUpdateCounter(idx);
+            // entry is occupied (anyhow), read the key, remember the result
+            keyReadResult = readKey(ctx.key, idx);
+            // read value, if value is invalid or deleted reference the read will fail
+            valueReadResult = readValue(ctx.value, idx);
+        } while (HASH_CODEC.getSecond(ctx.keyHashAndUpdateCnt) != getUpdCntFromHash(idx));
```
liran-funaro, line 224: You read here the cached hash instead of the entry's hash value.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -196,55 +207,48 @@ private boolean isKeyAndEntryKeyEqual(KeyBuffer tempKeyBuff, K key, int idx, int
...
+
+        if (valueReadResult) { // was value read successful?
```
liran-funaro, line 226: Do you assume that if the value read was successful, then the key read was also successful?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -196,55 +207,48 @@ private boolean isKeyAndEntryKeyEqual(KeyBuffer tempKeyBuff, K key, int idx, int
...
+            // Entry is Valid, but can be in deletion process
+            // checking the off-heap value header's delete bit
+            return (ctx.value.getSlice().isDeleted() != ValueUtils.ValueResult.FALSE) ?
+                EntryState.DELETED_NOT_FINALIZED : EntryState.VALID;
+        } else {
+            // value invalid or deleted
+            // Entry is either not yet inserted, or in deletion process or already deleted
+            if (keyReadResult) { // was key read successful?
```
liran-funaro, line 234: `else if (keyReadResult) ... `

liran-funaro (changes requested):
First quick pass. I'll do another pass after the first round of corrections.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
liran-funaro: Never mind. I didn't see the right part.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -196,55 +207,48 @@ private boolean isKeyAndEntryKeyEqual(KeyBuffer tempKeyBuff, K key, int idx, int
...
-                // for later progressing with deleted entry read current key slice
-                // (value is read during deleted key check, unless deleted)
-                if (readKey(ctx.key, idx)) {
-                    // key is not deleted: either this deletion is not yet finished,
-                    // or this is a new assignment on top of deleted entry
-                    // Check the key hash:
-                    // if invalid --> this is INSERT_NOT_FINALIZED if valid --> DELETED_NOT_FINALIZED
-                    if (isKeyHashValid(idx)) {
-                        return EntryState.DELETED_NOT_FINALIZED;
+        // try to get constant read of key, value, hash code, its update counter and valid bit
```
liran-funaro, line 217: constant? or did you mean consistent?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -108,6 +109,16 @@ private int getKeyHash(int ei) {
         return HASH_CODEC.getFirst(getKeyHashAndUpdateCounter(ei));
     }
 
+    /**
+     * getUpdCntFromHash returns the key hash's update counter of the entry given by entry index "ei"
+     * (without key hash itself!).
+     */
+    private int getUpdCntFromHash(int ei) {
+        assert isIndexInBound(ei);
+        // extract the hash number (first) from the updates counter (second)
```
liran-funaro, line 118: I don't understand this comment: "extract the hash number (first)".
What is first here?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
liran-funaro, line 224: Why compare just the update counter? You can compare the entire hash?
This would actually do less work since you won't have to extract the counter from the value.

liran-funaro (approved):
Looks good. Just some minor comments.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: fixed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: No. Key can be not valid if the entire entry can be in process of being deleted. This will be detected in the next line.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: changed

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: Indeed some weird comment, probably old one. Rewritten.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: agree, fixed

PR #188, opened 2021-12-18
Nova mem mgr v01

li0nr:

Introducing Nova : a Lock Free Off-Heap Memory Manager.

Nova can be used to replace the SeqExpandMemoryManager or and the SyncRecycleMemoryManager.

*Note: this PR replaces already opened one, adjusted to your latest changes!

### DIFF
--- a/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
+++ b/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak;
+
+public class DeletedMemoryAccessException extends RuntimeException {
+    public DeletedMemoryAccessException() {
+        super();
+    }
+
+    public DeletedMemoryAccessException(String message) {
+        super(message);
+    }
+}
--- a/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
@@ -6,7 +6,10 @@
 
 package com.yahoo.oak;
 
+import com.yahoo.oak.ValueUtils.ValueResult;
+
 import java.util.concurrent.atomic.AtomicInteger;
+
 
 /* EntryOrderedSet keeps a set of entries linked to a link list. Entry is reference to key and value, both located
  * off-heap. EntryOrderedSet provides access, updates and manipulation on each entry, provided its index.
@@ -306,5 +309,21 @@ class EntryOrderedSet<K, V> extends EntryArray<K, V> {
         }
 
         return true;
+    }
+    
+    void releaseAllDeletedKeys() {
+        KeyBuffer key = new KeyBuffer(config.keysMemoryManager.getEmptySlice());
+        for (int i = 0; i < entriesCapacity ; i++) {
+            if (config.keysMemoryManager.isReferenceDeleted(getValueReference(i))) {
+                if (config.keysMemoryManager.isReferenceValidAndNotDeleted(getKeyReference(i))) {
+                    key.s.decodeReference(getKeyReference(i));
+                    if (key.s.isDeleted() != ValueResult.TRUE) {
+                        if (key.s.logicalDelete() == ValueResult.TRUE) {
+                            key.s.release();
+                        }
+                    }
+                }
+            }
+        }
     }
 }
--- a/core/src/main/java/com/yahoo/oak/InternalOakMap.java
+++ b/core/src/main/java/com/yahoo/oak/InternalOakMap.java
@@ -46,14 +46,14 @@ class InternalOakMap<K, V>  extends InternalOakBasics<K, V> {
         Comparator<Object> mixedKeyComparator = (o1, o2) -> {
             if (o1 instanceof OakScopedReadBuffer) {
                 if (o2 instanceof OakScopedReadBuffer) {
-                    return comparator.compareSerializedKeys((OakScopedReadBuffer) o1, (OakScopedReadBuffer) o2);
+                    return KeyUtils.compareSerializedKeys( (KeyBuffer) o1, (KeyBuffer) o2, comparator);
                 } else {
                     // Note the inversion of arguments, hence sign flip
-                    return (-1) * comparator.compareKeyAndSerializedKey((K) o2, (OakScopedReadBuffer) o1);
+                    return (-1) * KeyUtils.compareSerializedKeys( (KeyBuffer) o1, (KeyBuffer) o2, comparator);
                 }
             } else {
                 if (o2 instanceof OakScopedReadBuffer) {
-                    return comparator.compareKeyAndSerializedKey((K) o1, (OakScopedReadBuffer) o2);
+                    return KeyUtils.compareKeyAndSerializedKey((K) o1, (KeyBuffer) o2, comparator);
                 } else {
                     return comparator.compareKeys((K) o1, (K) o2);
                 }
@@ -112,7 +112,7 @@ class InternalOakMap<K, V>  extends InternalOakBasics<K, V> {
 
         // since skiplist isn't updated atomically in split/compaction, our key might belong in the next orderedChunk
         // we need to iterate the chunks until we find the correct one
-        while ((next != null) && (config.comparator.compareKeyAndSerializedKey(key, next.minKey) >= 0)) {
+        while ((next != null) && (KeyUtils.compareKeyAndSerializedKey(key, next.minKey, config.comparator) >= 0)) {
             curr = next;
             next = curr.next.getReference();
         }
@@ -315,51 +315,55 @@ class InternalOakMap<K, V>  extends InternalOakBasics<K, V> {
         ThreadContext ctx = getThreadContext();
 
         for (int i = 0; i < MAX_RETRIES; i++) {
-            OrderedChunk<K, V> c = findChunk(key); // find orderedChunk matching key
-            c.lookUp(ctx, key);
-            // If there is a matching value reference for the given key, and it is not marked as deleted,
-            // then this put changes the slice pointed by this value reference.
-            if (ctx.isValueValid()) {
-                // there is a value and it is not deleted
-                Result res = config.valueOperator.exchange(c, ctx, value, transformer, getValueSerializer());
-                if (res.operationResult == ValueUtils.ValueResult.TRUE) {
-                    return (V) res.value;
+            try {
+                OrderedChunk<K, V> c = findChunk(key); // find orderedChunk matching key
+                c.lookUp(ctx, key);
+                // If there is a matching value reference for the given key, and it is not marked as deleted,
+                // then this put changes the slice pointed by this value reference.
+                if (ctx.isValueValid()) {
+                    // there is a value and it is not deleted
+                    Result res = config.valueOperator.exchange(c, ctx, value, transformer, getValueSerializer());
+                    if (res.operationResult == ValueUtils.ValueResult.TRUE) {
+                        return (V) res.value;
+                    }
+                    // it might be that this chunk is proceeding with rebalance -> help
+                    helpRebalanceIfInProgress(c);
+                    // Exchange failed because the value was deleted/moved between lookup and exchange. Continue with
+                    // insertion.
+                    continue;
                 }
-                // it might be that this chunk is proceeding with rebalance -> help
-                helpRebalanceIfInProgress(c);
-                // Exchange failed because the value was deleted/moved between lookup and exchange. Continue with
-                // insertion.
-                continue;
-            }
-
-            if (isAfterRebalanceOrValueUpdate(c, ctx)) {
-                continue;
-            }
-
-            // AT THIS POINT EITHER (in all cases context is updated):
-            // (1) Key wasn't found (key and value not valid)
-            // (2) Key was found and it's value is deleted/invalid (key valid value invalid)
-            if (!ctx.isKeyValid()) {
-                if (!allocateAndLinkEntry(c, ctx, key, false)) {
-                    continue; // allocation wasn't successfull and resulted in rebalance - retry
+    
+                if (isAfterRebalanceOrValueUpdate(c, ctx)) {
+                    continue;
                 }
-            }
-
-            c.allocateValue(ctx, value, false); // write value in place
-
-            if (!c.publish()) {
-                c.releaseNewValue(ctx);
-                rebalance(c);
+    
+                // AT THIS POINT EITHER (in all cases context is updated):
+                // (1) Key wasn't found (key and value not valid)
+                // (2) Key was found and it's value is deleted/invalid (key valid value invalid)
+                if (!ctx.isKeyValid()) {
+                    if (!allocateAndLinkEntry(c, ctx, key, false)) {
+                        continue; // allocation wasn't successfull and resulted in rebalance - retry
+                    }
+                }
+    
+                c.allocateValue(ctx, value, false); // write value in place
+    
+                if (!c.publish()) {
+                    c.releaseNewValue(ctx);
+                    rebalance(c);
+                    continue;
+                }
+    
+                if (c.linkValue(ctx) != ValueUtils.ValueResult.TRUE) {
+                    c.releaseNewValue(ctx);
+                    c.unpublish();
+                } else {
+                    c.unpublish();
+                    checkRebalance(c);
+                    return null; // null can be returned only in zero-copy case
+                }
+            } catch (DeletedMemoryAccessException e) {
                 continue;
-            }
-
-            if (c.linkValue(ctx) != ValueUtils.ValueResult.TRUE) {
-                c.releaseNewValue(ctx);
-                c.unpublish();
-            } else {
-                c.unpublish();
-                checkRebalance(c);
-                return null; // null can be returned only in zero-copy case
             }
         }
         throw new RuntimeException("put failed: reached retry limit (1024).");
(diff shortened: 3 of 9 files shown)


core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -185,7 +185,7 @@ private boolean isKeyAndEntryKeyEqual(KeyBuffer tempKeyBuff, K key, int idx, int
         }
 
         assert tempKeyBuff.isAssociated();
-        return (0 == comparator.compareKeyAndSerializedKey(key, tempKeyBuff));
+        return (0 == ((KeyMemoryManager ) keysMemoryManager).compareKeyAndSerializedKey(key, tempKeyBuff, comparator));
```
liran-funaro, line 188: I don't see why this is necessary? All MM that needs to lock the keys has the same implementation: lock / compare / unlock.
The only MM that is different is `SeqExpand`. And its lock/unlock methods do nothing anyway.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -0,0 +1,550 @@
...
+        public void markAsDeleted() {
+            assert version != ReferenceCodecSyncRecycle.INVALID_VERSION;
+            HEADER.logicalDelete(version, getMetadataAddress());
+        }
+
+    }
+    @Override
+    public <K> int compareKeyAndSerializedKey(K key, OakScopedReadBuffer serializedKey, OakComparator<K> cmp) {
+        if (((ScopedReadBuffer) serializedKey).s.lockRead() != ValueResult.TRUE) {
+            throw new DeletedMemoryAccessException();
```
liran-funaro, line 531: It might be already locked. Why do you assume it is deleted?
The same issue is below and in other MMs.

core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak;
+
+public class DeletedMemoryAccessException extends RuntimeException {
```
liran-funaro, line 9: Why is it a runtime exception? Just to avoid declaring it?
Please change it to a regular exception.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -0,0 +1,550 @@
...
+        }
+        int res = cmp.compareKeyAndSerializedKey(key, serializedKey);
```
liran-funaro, line 533: This should be wrapped in try/catch to avoid locking the object forever.
The same issue is below and in other MMs.

liran-funaro (changes requested):
Initial review.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
technionRamy: I am not sure what is your question, 
are you asking why I am wrapping the compare with the memory manger lock/ unlock?
I am doing so, that when we have a concurrent delete with another access, we need to be notified about this.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
technionRamy: In Nova MM the "lock" is not indeed a lock, and the only scenario that we can not "lock" is when the object is deleted.
hmmm any other names for the exception are welcomed!

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
technionRamy: Hmm, again Nova is non blocking we don't really use locks,
in other MM however how could we lock object forever?

core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
technionRamy: I think it is cleaner that way, since it is un-checked exception.
Any reason why to not use RuntimeException ?

liran-funaro: Here is a good explanation: https://stackoverflow.com/a/11589380/2570677
Since this error is not the programmer's fault and it is recoverable, then it should be checked.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
liran-funaro: It is a statement, not a quesion.
I don't understand why each memory manager has to implement its own compare keys function.
(a) It is not modular. MM has nothing to do with key's comparison.
(b) It is not necessary. It can be implemented easily outside the MM.
Please modify your code such that these compare functions will be implemented once (not for each MM), and in another class (not inside MM).
I suggest maybe a new class for key management (like the one we have for values).

core/src/main/java/com/yahoo/oak/SyncRecycleMemoryManager.java
```diff
@@ -470,4 +472,27 @@ public void markAsDeleted() {
             HEADER.markAsDeleted(getMetadataAddress());
         }
     }
+    
+    @Override
+    public <K> int compareKeyAndSerializedKey(K key, OakScopedReadBuffer serializedKey, OakComparator<K> cmp) {
+        if (((ScopedReadBuffer) serializedKey).s.lockRead() != ValueResult.TRUE) {
+            throw new DeletedMemoryAccessException();
```
liran-funaro, line 479: Does this MM lock actually lock?
If so, then "deleted" isn't right here.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
liran-funaro: Then why aren't we using `isDeleted()` directly?
Does Nova not support mutual exclusion?

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
liran-funaro: If the object is locked, then the compare function fails with an exception, then we never call unlock.

liran-funaro (changes requested):

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
technionRamy: If we started the comparison, this means that we already successfully locked the key, we can not fail while comparing.
If we failed because we could not lock the key, then we throw the exception.

liran-funaro: `cmp.compareKeyAndSerializedKey(key, serializedKey);` is a user-provided function. You don't know what it does. Specifically, it might throw an unchecked exception. Then, unlock will not be called.

core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
technionRamy: This turned up to be not such good idea, although it is more "right to do".

especially in the iterator, since iterator is not declared to throw stuff, we need to deal with this and to try and catch in the iterator code, which results in throwing a runtime exception at the end.

Moreover this exception is not meant to be for the programmer, it is internal to Oak, and we retry when get such exception.

liran-funaro:
@technionRamy I think you forgot to add KeyUtil.java

core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
liran-funaro: Your statements are contridicting. If you handle the exception internally, then you have to change the iterator code anyway.
This is all the more reason to change this exception to be checked. So we can validate that we never throw this exception to the user.
Right now, with your current implementation, this exception might be aggregated to the user without you noticing it.

core/src/main/java/com/yahoo/oak/KeyUtils.java
```diff
@@ -0,0 +1,49 @@
...
+            throw new DeletedMemoryAccessException();
+        }
+        try {
+            int res = cmp.compareKeyAndSerializedKey(key, serializedKey);
+            ((ScopedReadBuffer) serializedKey).s.unlockRead();
+            return res;
+        } catch (Exception e) {
+            ((ScopedReadBuffer) serializedKey).s.unlockRead();
+            throw e;
+        }
```
liran-funaro, line 29:
```suggestion
        try {
            return cmp.compareKeyAndSerializedKey(key, serializedKey);
        } finally {
            ((ScopedReadBuffer) serializedKey).s.unlockRead();
        }
```

core/src/main/java/com/yahoo/oak/KeyUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        try {
+            int res = cmp.compareSerializedKeys(serializedKey1, serializedKey2);
+            ((ScopedReadBuffer) serializedKey1).s.unlockRead();
+            ((ScopedReadBuffer) serializedKey2).s.unlockRead();
+            return res;
+        } catch (Exception e) {
+            ((ScopedReadBuffer) serializedKey1).s.unlockRead();
+            ((ScopedReadBuffer) serializedKey2).s.unlockRead();
+            throw e;
+        }
```
liran-funaro, line 47:
```suggestion
        try {
            return cmp.compareSerializedKeys(serializedKey1, serializedKey2);
        } finally {
            ((ScopedReadBuffer) serializedKey1).s.unlockRead();
            ((ScopedReadBuffer) serializedKey2).s.unlockRead();
        }
```

core/src/main/java/com/yahoo/oak/KeyUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    
+    
+    public static <K> int compareKeyAndSerializedKey(K key, OakScopedReadBuffer serializedKey, OakComparator<K> cmp) 
+            throws DeletedMemoryAccessException {
+        if (((ScopedReadBuffer) serializedKey).s.lockRead() != ValueResult.TRUE) {
+            throw new DeletedMemoryAccessException();
+        }
+        try {
+            int res = cmp.compareKeyAndSerializedKey(key, serializedKey);
+            return res;
```
liran-funaro, line 24:
```suggestion
            return cmp.compareKeyAndSerializedKey(key, serializedKey);
```

core/src/main/java/com/yahoo/oak/KeyUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    
+    public static <K> int compareSerializedKeys(OakScopedReadBuffer serializedKey1,
+            OakScopedReadBuffer serializedKey2, OakComparator<K> cmp) throws DeletedMemoryAccessException {
+        if (    ((ScopedReadBuffer) serializedKey1).s.lockRead() != ValueResult.TRUE ||
+                ((ScopedReadBuffer) serializedKey2).s.lockRead() != ValueResult.TRUE) {
+            throw new DeletedMemoryAccessException();
+        }
+        try {
+            int res = cmp.compareSerializedKeys(serializedKey1, serializedKey2);
+            return res;
```
liran-funaro, line 38:
```suggestion
            return cmp.compareSerializedKeys(serializedKey1, serializedKey2);
```

liran-funaro (changes requested):

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -510,68 +527,73 @@ Result remove(K key, V oldValue, OakTransformer<V> transformer) {
...
-            return transformer == null ?
-                ctx.result.withFlag(ValueUtils.ValueResult.TRUE) : ctx.result.withValue(v);
         }
 
         throw new RuntimeException("remove failed: reached retry limit (1024).");
     }
 
     // the zero-copy version of get
-    OakUnscopedBuffer get(K key) {
+    OakUnscopedBuffer get(K key) { //TODO what we do here?
```
sanastas, line 596: Here I would say to try/catch and to repeat again, but there is a chance to get into the infinite loop if nothing changes. In OakMap case we can get onto deleted key only if we are looking on the RELEASED chunk, and the released chunk shouldn't be found in skip-list anyway. So to ensure we are not gwtting to the same released chunk again and again we should accomplish the rebalance of this released chunk and to take it out of the skip-list, before next iteration of the loop.

sanastas: I mean we need to ensure method `inTheMiddleOfRebalance(BasicChunk<K, V> c)` is called within each loop iteration. This is also the answer to other `//TODO what we do here?` comments down this file

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -642,25 +668,28 @@ boolean refreshValuePosition(ThreadContext ctx) {
...
 
-            Result res = valueOperator.transform(ctx.result, ctx.value, transformer);
-            if (res.operationResult == ValueUtils.ValueResult.RETRY) {
+                Result res = valueOperator.transform(ctx.result, ctx.value, transformer);
+                if (res.operationResult == ValueUtils.ValueResult.RETRY) {
+                    continue;
+                }
+                return (T) res.value;
+            } catch (DeletedMemoryAccessException e) {
                 continue;
```
sanastas, line 685: Add `inTheMiddleOfRebalance(c)` call before continuing

core/src/main/java/com/yahoo/oak/KeyUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+class KeyUtils {
+
+
+    /* ==================== Methods for operating on existing off-heap keys ==================== */
+    
+    
+    public static <K> int compareKeyAndSerializedKey(K key, OakScopedReadBuffer serializedKey, OakComparator<K> cmp) 
+            throws DeletedMemoryAccessException {
+        if (((ScopedReadBuffer) serializedKey).s.lockRead() != ValueResult.TRUE) {
```
sanastas, line 19: It is a bit disturbing that lock/unlock methods were "overloaded" for Nova start/stop. Can we use some other API name?

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -0,0 +1,152 @@
...
+
+
+class NovaMMHeader {
+
+    /*
+    * Long NovaMMHeader: int version + int lock
+    * 0...  ...21 | 32...                  ...63
+    *  version    |  length
+    *
+    * The length (Mostly in integer size) of the data is also held just following the header (since the block and
```
sanastas, line 21: Unclear... Is the length part of the header or not? Is it that header keep only version and length. Where is the delete bit?

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -0,0 +1,152 @@
...
+    long getOffHeapMetaData(long headerAddress) {
+        return UnsafeUtils.UNSAFE.getLong(headerAddress);
+    }
+
+    private void setDataLength(long headerAddress, int length) {
+        throw new IllegalAccessError();
+    }
+
+    void initHeader(long headerAddress, int dataLength, int version) {
+        long sliceHeader    = (long) (dataLength) << VERSION_SIZE;
```
sanastas, line 47: Please take a look on the classes/files `UnionCodec.java` and `ReferenceCodec.java` they are comfortable and well tested. So you don't need to do all those bit-wise operations yourself.

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -0,0 +1,152 @@
...
+        int  newVer         = version & 0x3FFFFF; //VERSION_SIZE;
+        long header         = sliceHeader | newVer ;
+        UnsafeUtils.UNSAFE.putLong(headerAddress, header);
+    }
+
+    public  void setTap(AtomicLongArray tap, long ref, int idx) {
+        int i = idx % ThreadIndexCalculator.MAX_THREADS;
+        tap.set(NovaMemoryManager.CACHE_PADDING * (i + 1) + NovaMemoryManager.IDENTRY, idx);
```
sanastas, line 55: Why twice?

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -0,0 +1,152 @@
...
+        tap.set(NovaMemoryManager.CACHE_PADDING * (i + 1) + NovaMemoryManager.REFENTRY, ref);
+    }
+        
+    public  void unsetTap(AtomicLongArray tap, int idx) {
+        int i = idx % ThreadIndexCalculator.MAX_THREADS;
+        tap.set(NovaMemoryManager.CACHE_PADDING * (i + 1) + NovaMemoryManager.IDENTRY, -1);
```
sanastas, line 61: Is it atomic operation?

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -0,0 +1,152 @@
...
+
+    
+    private static boolean cas(long headerAddress, long offHeapMeta) {
+        // Since the writing is done directly to the memory, the endianness of the memory is important here.
+        // Therefore, we make sure that the values are read and written correctly.
+        return UnsafeUtils.UNSAFE.compareAndSwapLong(null, headerAddress, offHeapMeta, offHeapMeta | 1);
+    }
+    
+    
+    ValueUtils.ValueResult lockRead(final int onHeapVersion, long headerAddress) {
```
sanastas, line 83: It is not really a lock/unlock so, as I previously mentioned, confusing... Please think about new (additional) API for start/stop in Slide.

Also checking the bit via `%2` is not clear. Either comment very well, ore use `UnionCodec` for three values

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -0,0 +1,526 @@
...
+
+    
+    
+    private final ThreadIndexCalculator threadIndexCalculator;
+    private final List<List<SliceNova>> releaseLists;
+    private final AtomicInteger globalVersionNumber;
+    private final BlockMemoryAllocator allocator;
+
+    
+    private final AtomicLongArray tap;
```
sanastas, line 36: I am not sure what is better `AtomicLongArray` or normal array with UNSAFE.CAS where needed only...

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -0,0 +1,526 @@
...
+package com.yahoo.oak;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLongArray;
+
+
+class NovaMemoryManager implements MemoryManager {
```
sanastas, line 16: Will it be beneficial to inherit from `SyncRecycleMemoryManager` and overwrite what is needed?

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -0,0 +1,526 @@
...
+        if (curVer == ReferenceCodecSyncRecycle.LAST_VALID_VERSION) {
+            globalVersionNumber.compareAndSet(curVer, VERS_INIT_VALUE);
+        } else {
+            globalVersionNumber.compareAndSet(curVer, curVer + 1);
+        }
+        // if CAS fails someone else updated the version, which is good enough
+    }
+
+    /*=====================================================================*/
+    /*           SliceSyncRecycle                 */
```
sanastas, line 180: Please correct the comment

sanastas: And all mentioning of SliceSyncRecycle in the file

sanastas (commented):
I didn't fully finish, but already have many comments. Let's start from them. But very good work, thanks!

core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
```diff
@@ -310,4 +312,21 @@ boolean isEntrySetValidAfterRebalance() {
 
         return true;
     }
+    
+    void releaseAllDeletedKeys() {
+        KeyBuffer key = new KeyBuffer(keysMemoryManager.getEmptySlice());
+        for (int i = 0; i < entriesCapacity ; i++) {
+            if (valuesMemoryManager.isReferenceDeleted(getValueReference(i))) {
+                if (keysMemoryManager.isReferenceValidAndNotDeleted(getKeyReference(i))) {
+                    //TODO may add exception to isDelted and stop deleting if key is already deleted
```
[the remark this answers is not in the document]
li0nr: Here I thought that we when we start deleting keys from a released chunk, and more than one thread start this deletion
then we can just make sure that thread continue the deleting process, and all the others stop checking.

core/src/main/java/com/yahoo/oak/KeyUtils.java
li0nr: If we used something else, then Nova would not be a simple plugin 
we need to replace/duplicate everything that uses lock to the new API

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
li0nr: the first set is for the thread ID, the second for what block-offset is being held now.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
li0nr: the use of atomicLongArrray is more correct in my opinion, 
since it guarantees the volatile guarantees, after entry is set every read after it being set sees this entry updated (with the new value).
Using just the unsafe CAS does not guarantee this,
we dont care about atomicity we care about preserving the order so we dont delete a slice that is being written to.
(If i recall correctly we there is not difference from performance point of view).

li0nr:
not relevant anymore!

PR #193, opened 2021-12-26
Fix memory leak in tests

liran-funaro:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/BlocksPool.java
+++ b/core/src/main/java/com/yahoo/oak/BlocksPool.java
@@ -97,6 +97,18 @@ final class BlocksPool implements BlocksProvider, Closeable {
         }
     }
 
+    static void clear() {
+        synchronized (BlocksPool.class) { // can be easily changed to lock-free
+            if (instance == null) {
+                return;
+            }
+
+            while (instance.blocks.size() > 0) {
+                instance.blocks.poll().clean();
+            }
+        }
+    }
+
     /**
      * Sets the preferred block size. This only has an effect if the block pool was never used before.
      * @param preferredBlockSizeBytes the preferred block size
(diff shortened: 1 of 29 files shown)


core/src/test/java/com/yahoo/oak/SeqExpandMemoryManagerTest.java
```diff
@@ -22,6 +23,12 @@ public void setUp() {
         seqExpandMemoryManager = new SeqExpandMemoryManager(keysMemoryAllocator);
     }
 
+    @After
+    public void tearDown() {
+        seqExpandMemoryManager.close();
```
sanastas, line 28: Why do we need a special `close()` here? And not in the rest of the tests?

core/src/main/java/com/yahoo/oak/BlocksPool.java
```diff
@@ -97,6 +97,18 @@ static void setBlockSize(int blockSize) {
         }
     }
 
+    static void clear() {
```
sanastas, line 100: Is it some new API we expose outside of Oak? If yes, we should update README. If not, then we probably should expose it somehow, because it might be useful for some cases. However, it should be well explained.

sanastas (approved):
Looks good. Please update README.

core/src/test/java/com/yahoo/oak/SeqExpandMemoryManagerTest.java
liran-funaro: In the other tests, we close the Oak instance, which closes the MM. Here we don't have Oak instance.

core/src/main/java/com/yahoo/oak/BlocksPool.java
liran-funaro: Currently, `BlocksPool` is not exposed outside. So this change is not seen by the user.
Exposing `BlocksPool` functionality to the user should be discussed, but not as part of this PR.

PR #196, opened 2021-12-28
Oak shared config

### LINKED ISSUE #194: Can I use one OakMapBuilder to create more than one OakMap?
OakMapBuilder<Long, AckData> oakMapBuilder = new OakMapBuilder<>(new OakLongComparator(),
            new OakLongSerializer(), new OakAckDataSerializer(), Long.MIN_VALUE).setMemoryCapacity(MEBIBYTE);
        for (int i = 0; i < 1000; i ++) {
            OakMap<Long, AckData> oakMap = oakMapBuilder.buildOrderedMap();
        }
image

And If I use one OakMapBuilder to create more than one OakMap in multi-threaded process, NativeMemoryAllocator will be Instantiated multiple times.

### LINKED ISSUE #195: Hi, I find a strange problem about OakMap.remove()
image

I don't quite understand why ctx.key.getSlice() is not released when OakMap.remove() is called, but only ctx.value.getSlice() is released.
It is also because ctx.key.getSlice() is not released, the blocks are created one by one, and finally my process hangs because of outofmemory.

liran-funaro:
fixes #194 

* Oak shared config
* Allow multiple Oak instantiations with one builder
* Fix more memory leaks in tests
* Fix bad assert pattern in tests

### DIFF
--- a/core/src/main/java/com/yahoo/oak/BasicChunk.java
+++ b/core/src/main/java/com/yahoo/oak/BasicChunk.java
@@ -21,15 +21,16 @@ abstract class BasicChunk<K, V> {
 
 
     /*-------------- Members --------------*/
-    // to compare serilized and object keys
+    protected final OakSharedConfig<K, V> config;
     protected OakComparator<K> comparator;
+
     // in split/compact process, represents parent of split (can be null!)
     private final AtomicReference<BasicChunk<K, V>> creator;
     // chunk can be in the following states: normal, frozen or infant(has a creator)
     private final AtomicReference<State> state;
     private final AtomicReference<Rebalancer<K, V>> rebalancer;
     private final AtomicInteger pendingOps;
-    private final int maxItems;
+    protected final int maxItems;
     protected AtomicInteger externalSize; // for updating oak's size (reference to one global per Oak size)
     protected final Statistics statistics;
 
@@ -37,11 +38,15 @@ abstract class BasicChunk<K, V> {
     /**
      * This constructor is only used internally to instantiate a BasicChunk without a creator and a state.
      * The caller should set the creator and state before returning the BasicChunk to the user.
+     *
+     * @param config shared configuration
+     * @param maxItems maximal capacity
      */
-    protected BasicChunk(int maxItems, AtomicInteger externalSize, OakComparator<K> comparator) {
+    protected BasicChunk(OakSharedConfig<K, V> config, int maxItems) {
+        this.config = config;
         this.maxItems = maxItems;
-        this.externalSize = externalSize;
-        this.comparator = comparator;
+        this.externalSize = config.size;
+        this.comparator = config.comparator;
         this.creator = new AtomicReference<>(null);
         this.state = new AtomicReference<>(State.NORMAL);
         this.pendingOps = new AtomicInteger();
@@ -55,7 +60,6 @@ abstract class BasicChunk<K, V> {
     protected void updateBasicChild(BasicChunk<K, V> child) {
         child.creator.set(this);
         child.state.set(State.INFANT);
-        return;
     }
     /*---------------Abstract Read methods -----------------------*/
     abstract void readKey(ThreadContext ctx);
--- a/core/src/main/java/com/yahoo/oak/BasicChunk.java
+++ b/core/src/main/java/com/yahoo/oak/BasicChunk.java
@@ -23,9 +23,6 @@ abstract class BasicChunk<K, V> {
     /*-------------- Members --------------*/
     protected final OakSharedConfig<K, V> config;
 
-    // to compare serialized and object keys
-    protected OakComparator<K> comparator;
-
     // in split/compact process, represents parent of split (can be null!)
     private final AtomicReference<BasicChunk<K, V>> creator;
     // chunk can be in the following states: normal, frozen or infant(has a creator)
@@ -33,7 +30,6 @@ abstract class BasicChunk<K, V> {
     private final AtomicReference<Rebalancer<K, V>> rebalancer;
     private final AtomicInteger pendingOps;
     protected final int maxItems;
-    protected AtomicInteger externalSize; // for updating oak's size (reference to one global per Oak size)
     protected final Statistics statistics;
 
     /*-------------- Constructors and creators --------------*/
@@ -47,8 +43,6 @@ abstract class BasicChunk<K, V> {
     protected BasicChunk(OakSharedConfig<K, V> config, int maxItems) {
         this.config = config;
         this.maxItems = maxItems;
-        this.externalSize = config.size;
-        this.comparator = config.comparator;
         this.creator = new AtomicReference<>(null);
         this.state = new AtomicReference<>(State.NORMAL);
         this.pendingOps = new AtomicInteger();
--- a/core/src/main/java/com/yahoo/oak/BasicChunk.java
+++ b/core/src/main/java/com/yahoo/oak/BasicChunk.java
@@ -22,6 +22,8 @@ abstract class BasicChunk<K, V> {
 
     /*-------------- Members --------------*/
     protected final OakSharedConfig<K, V> config;
+
+    // to compare serialized and object keys
     protected OakComparator<K> comparator;
 
     // in split/compact process, represents parent of split (can be null!)
--- a/core/src/main/java/com/yahoo/oak/EntryArray.java
+++ b/core/src/main/java/com/yahoo/oak/EntryArray.java
@@ -58,40 +58,38 @@ public class EntryArray<K, V> {
     protected static final int VALUE_REF_OFFSET = 1;
     static final int INVALID_ENTRY_INDEX = -1;
 
-    final MemoryManager valuesMemoryManager;
     final MemoryManager keysMemoryManager;
+    final MemoryManager valuesMemoryManager;
+
+    // for writing the keys into the off-heap
+    final OakSerializer<K> keySerializer;
+    final OakSerializer<V> valueSerializer;
+
     private final long[] entries;    // array is initialized to 0 - this is important!
     private final int fields;  // # of primitive fields in each item of entries array
 
     final int entriesCapacity; // number of entries (not longs) to be maximally held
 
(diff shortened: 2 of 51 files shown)


core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -21,27 +21,34 @@
...
     protected OakComparator<K> comparator;
+
     // in split/compact process, represents parent of split (can be null!)
     private final AtomicReference<BasicChunk<K, V>> creator;
     // chunk can be in the following states: normal, frozen or infant(has a creator)
     private final AtomicReference<State> state;
     private final AtomicReference<Rebalancer<K, V>> rebalancer;
     private final AtomicInteger pendingOps;
-    private final int maxItems;
+    protected final int maxItems;
```
sanastas, line 35: Why not to use getMaxItems()?

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -58,40 +58,38 @@
     protected static final int VALUE_REF_OFFSET = 1;
     static final int INVALID_ENTRY_INDEX = -1;
 
-    final MemoryManager valuesMemoryManager;
     final MemoryManager keysMemoryManager;
+    final MemoryManager valuesMemoryManager;
+
+    // for writing the keys into the off-heap
```
sanastas, line 64: ...keys and values...

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -58,40 +58,38 @@
...
+    final OakSerializer<V> valueSerializer;
+
     private final long[] entries;    // array is initialized to 0 - this is important!
     private final int fields;  // # of primitive fields in each item of entries array
 
     final int entriesCapacity; // number of entries (not longs) to be maximally held
 
     // Counts number of entries inserted & not deleted. Pay attention that not all entries (counted
-    // in number of entries) are finally are finally considered existing by the OrderedChunk above
+    // in number of entries) are finally considered existing by the OrderedChunk above
```
sanastas, line 74: OrderedChunk --> Chunk

core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -40,28 +38,20 @@
...
-        assert Math.pow( 2, hashIndexCodec.getFirstBitSize() ) <= maxItems ;
+    HashChunk(OakSharedConfig<K, V> config, int maxItems, UnionCodec hashIndexCodec) {
+        super(config, maxItems);
+        assert Math.pow( 2, hashIndexCodec.getFirstBitSize() ) <= maxItems;
 
         this.hashIndexCodec = hashIndexCodec;
-        this.entryHashSet = // must be called after setSecondLevelBitsThreshold
-            new EntryHashSet<>(vMM, kMM, getMaxItems(), keySerializer, valueSerializer,
-                comparator);
+        // must be called after setSecondLevelBitsThreshold
```
sanastas, line 46: setSecondLevelBitsThreshold --> hashIndexCodec assignment

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -15,22 +15,30 @@
     /*-------------- Members --------------*/
     protected static final int MAX_RETRIES = 1024;
 
-    private final MemoryManager valuesMemoryManager;
-    private final MemoryManager keysMemoryManager;
+    protected final OakSharedConfig<K, V> config;
 
-    private final OakSerializer<K> keySerializer;
-    private final OakSerializer<V> valueSerializer;
+    protected final MemoryManager keysMemoryManager;
```
sanastas, line 20: Why to keep the config and all those fields? In order to avoid de-referencing? I don't believe we will see any performance differences from accessing `config.keysMemoryManager`...  We may try the benchmarks....

core/src/main/java/com/yahoo/oak/OakMap.java
```diff
@@ -28,60 +28,58 @@
...
     private final K toKey;
-    private boolean toInclusive;
+    private final boolean toInclusive;
     private final boolean isDescending;
 
-    // internal constructor, to create OakMap use OakMapBuilder
-    OakMap(K minKey, OakSerializer<K> keySerializer, OakSerializer<V> valueSerializer, OakComparator<K> oakComparator,
-        int chunkMaxItems, MemoryManager vMM, MemoryManager kMM) {
-
+    private OakMap(InternalOakMap<K, V> internalOakMap,
```
sanastas, line 47: Add a comment this is only for sub-maps creation

core/src/main/java/com/yahoo/oak/OakMap.java
```diff
@@ -28,60 +28,58 @@
...
+        this.valueDeserializeTransformer = config.valueSerializer::deserialize;
         this.entryDeserializeTransformer = entry -> new AbstractMap.SimpleEntry<>(
-                keySerializer.deserialize(entry.getKey()),
-                valueSerializer.deserialize(entry.getValue()));
+                keyDeserializeTransformer.apply(entry.getKey()),
+                valueDeserializeTransformer.apply(entry.getValue()));
+    }
+
+    // internal constructor, to create OakMap use OakMapBuilder
+    OakMap(OakSharedConfig<K, V> config, K minKey, int chunkMaxItems) {
```
sanastas, line 69: Where this construct is used? Should it be private?

core/src/main/java/com/yahoo/oak/OakSharedConfig.java
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright 2020, Verizon Media.
+ * Licensed under the terms of the Apache 2.0 license.
+ * Please see LICENSE file in the project root for terms.
+ */
+
+package com.yahoo.oak;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class OakSharedConfig<K, V> {
```
sanastas, line 11: Please add some explanatory comment

sanastas (commented):
The change looks good to me, some questions/comments inside

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -21,27 +21,34 @@
 
 
     /*-------------- Members --------------*/
-    // to compare serilized and object keys
+    protected final OakSharedConfig<K, V> config;
+
+    // to compare serialized and object keys
```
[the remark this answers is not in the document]
liran-funaro: Each instance holds its own references to the "config items" it needs.
I implemented it that way to avoid another pointer dereference each time we access these elements.
In addition, the config object is stored here to be passed to the child Chunk instance.

core/src/main/java/com/yahoo/oak/BasicChunk.java
liran-funaro: It seems more natural to me to access internal *final* fields directly. I don't see why we need a function call here.
For the same reason, we access `externalSize` and `statistics` directly.

core/src/main/java/com/yahoo/oak/EntryArray.java
liran-funaro: fixed

core/src/main/java/com/yahoo/oak/EntryArray.java
liran-funaro: fixed

core/src/main/java/com/yahoo/oak/HashChunk.java
liran-funaro: fixed

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
liran-funaro: Comparators are accessed frequently. Are you sure?
I think we can have a different PR after this one where we test this hypothesis.

core/src/main/java/com/yahoo/oak/OakMap.java
liran-funaro: Used by the builder. Added comment.

core/src/main/java/com/yahoo/oak/OakMap.java
liran-funaro: Used by the other constructors. Added a comment.

core/src/main/java/com/yahoo/oak/OakSharedConfig.java
liran-funaro: added

sanastas (approved):

fujian-zfj:
Excuse me. I have some questions about the oak shared config.
I used  to think that all oakMaps created by one OakMapBuilder could share the same BlockMempryAllocator, but the current design doesn't. I think that current design brings a probelm. When I use the same OakMapBuilder to create thousands of OakMaps in a multi-threaded environment, the BlockMemoryAllocator will be instantiated multiple times. At the same time, the problem is that thousands of 128M Block will be created in multi-threaded environment, and soon the memory is exhausted.
image

fujian-zfj:
Could all oakMap created by one OakMapBuilder share the same BlockMempryAllocator?
Because in real producetion, it is very frequent to create thousands of OakMaps in a multi-threaded environment.

sanastas:
Hi @fujian-zfj 

Thanks for your valuable suggestion. Indeed, we currently have one `NativeMemoryAllocator` per `OakHash` or `OakMap` instance. All memory allocators are using off-heap memory blocks from the same pool. It is possible to make the Oak instances to share same `NativeMemoryAllocator` (which is multi-threaded), but code needs to be upfated during closing, so closing one Oak instance doesn't close allocator for all others. Also too many Oak instances per memory allocator can cause threads contention and decrease the performance.

Anyway, in your case, a possible solution can be to decrease the block size. Use `builder.setPreferredBlockSize()` **before** any Oak instance is initiated.

Alternatively, if you would like to contribute some code to Oak, we can guide you to make the change yourself. You can get direct and clear instructions what needs to be done, code review and etc. This is also a possibility :)

PR #201, opened 2022-02-01
added hash iterators to benchmarks tests

boris-kap:


benchmarks/synchrobench/src/main/java/com/yahoo/oak/OakBenchHash.java
```diff
@@ -62,12 +62,15 @@ public void close() {
...
+        boolean result = createAndScanView(oakHash, length, blackhole);
+
+        return result;
     }
 
     /** {@inheritDoc} **/
     @Override
     public boolean descendOak(BenchKey from, int length, Blackhole blackhole) {
-        throw new UnsupportedOperationException("ALL ITERATORS ARE NOT YET SUPPORTED FOR HASH");
+        throw new UnsupportedOperationException("Descending iterator are not defined for Hash");
```
sanastas, line 74: In `JavaHashMap` the `ascendOak` is invoked when `descendOak` is requested. I think it is reasonable. Why not to implement it the same way here?

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/maps/BenchOakMap.java
```diff
@@ -119,23 +118,23 @@ public float nonHeapAllocatedGB() {
 
     /**
      * Iterate over a map and consume the keys/values.
-     * @param subMap subset of the map
+     * @param map map, or its subset
```
sanastas, line 121: One "map" is a typo?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -693,4 +693,9 @@ boolean isEntrySetValidAfterRebalance() {
 
         return true;
     }
+
+
```
sanastas, line 697: Please some comment about this method

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -406,37 +406,35 @@ void printSummaryDebug() {
 
     private static final class IteratorState<K, V> extends BasicIteratorState<K, V> {
 
-        // lastKeyAccessed is used to find the index of the last chunk accessed
-        // but calculating the key hash and extracting the MSBs.
-        //keyBufferValid is used to check the validity of lastKeyAccessed field
-        private KeyBuffer lastKeyAccessed = null;
+        // lastKeyHash is used to find the index of the last chunk accessed by extracting the MSBs.
+        // keyBufferValid is used to check the validity of lastKeyHash field
+        private int lastKeyHash;
```
sanastas, line 411: You may use `EntryHashSet.INVALID_KEY_HASH` as default. It is `-1`

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -406,37 +406,35 @@ void printSummaryDebug() {
...
         }
 
+        private boolean getKeyValid() {
+            return this.keyBufferValid;
+        }
         @Override
         public void copyState(BasicIteratorState<K, V> other) {
             super.copyState(other);
-            lastKeyAccessed = ((IteratorState<K, V>) other).lastKeyAccessed;
+            lastKeyHash = ((IteratorState<K, V>) other).lastKeyHash;
```
sanastas, line 437: I am adding a comment here, because I cannot add it to the line where I see something strange. Just saw it few lines below, as part of the review:

```
@Override
        protected void initAfterRebalance() {
            //TODO - refactor to use OakReadBuffer without deserializing.
            getState().getChunk().readKeyFromEntryIndex(ctx.tempKey, getState().getIndex());
            // Update the state to point to last returned key.
            initState();
}
```

I wonder what for the following lines are part of the ` initAfterRebalance()` for hash? Is the key used later?
```
//TODO - refactor to use OakReadBuffer without deserializing.
getState().getChunk().readKeyFromEntryIndex(ctx.tempKey, getState().getIndex());
```

sanastas: Also thinking about it. `initAfterRebalance()` is absolutely wrong now. Not that we invested thoughts into this implementation. But now when you are keeping the hash of the last key. `initAfterRebalance()` can be implemented 95% correctly. Similar to `advanceState()` just look for the key with the same hash on the new chunk...

sanastas (approved):
I finished the review. The code looks good to me. Thanks!
Please fix the comments and we will discuss it a bit more.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OakBenchHash.java
boris-kap: I think it's a bit misleading. But for the sake of unification, we can make it the same as in JavaHashMap

benchmarks/synchrobench/src/main/java/com/yahoo/oak/synchrobench/maps/BenchOakMap.java
boris-kap: no, it's the name of the function parameter

core/src/main/java/com/yahoo/oak/InternalOakHash.java
boris-kap: the function is not implemented, so it can't be wrong :). 
The problem is that splitting a chunk spreads the existing entries between the two new chunks, and the order of keys may change. So , if we just go on scanning, we can either skip some entries or return the same entry twice.

benchmarks/synchrobench/src/main/java/com/yahoo/oak/OakBenchHash.java
sanastas: 👍

core/src/main/java/com/yahoo/oak/InternalOakHash.java
sanastas: That is why I said it is 95% correct and not 100%. Anyway we can refer to it once rebalance is in place.

PR #205, opened 2022-03-22
Add hash to benchmarks

boris-kap:


core/src/main/java/com/yahoo/oak/HashChunk.java
```diff
@@ -332,6 +332,11 @@ HashChunkIter chunkIter(ThreadContext ctx) {
         }
 
 
+        /**
+         * find the index of the next occupied entry in the entryies array
```
sanastas, line 336: The explanation should say the "first" index, this is what name of the function suggests...

core/src/main/java/com/yahoo/oak/EntryHashSet.java
```diff
@@ -665,7 +675,14 @@ boolean isEntryDeleted(ValueBuffer tempValue, int ei) {
         return tempValue.getSlice().isDeleted() != ValueUtils.ValueResult.FALSE;
     }
 
-
+    @Override
+    int getNextNonZeroIndex(int currentIndex) {
+        int nxtIdx = mapOfCleanEntries.nextSetBit(currentIndex + 1);
```
sanastas, line 685: Why do you think it is OK not to use synchronization while reading?

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -56,6 +56,7 @@
      */
     protected static final int KEY_REF_OFFSET = 0;
     protected static final int VALUE_REF_OFFSET = 1;
+    protected static final int OPT_FIELD_OFFSET = 2;
```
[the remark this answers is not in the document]
boris-kap: this is the field after the KEY and VALUE fields, so it is always 2, regardless of the additionalFields Count, isn't it?

core/src/main/java/com/yahoo/oak/EntryHashSet.java
boris-kap: if several processes read read the value at the same time, could it be a problem?

core/src/main/java/com/yahoo/oak/EntryArray.java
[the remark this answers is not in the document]
sanastas: If additionalFieldsCount is zero then there is no meaning in OPT_FIELD_OFFSET. Maybe, then OPT_FIELD_OFFSET can be 0 or somehow eliminated...

core/src/main/java/com/yahoo/oak/EntryHashSet.java
sanastas: Generally, shouldn't be, but it depends how `BitSet` is implemented. If the update is more than a machine word, such an unprotected read may read partially updated structure. Do we know at least approximately whether BitSet is array with relevant bits set or may be a linked list with nodes, for example?

core/src/main/java/com/yahoo/oak/EntryArray.java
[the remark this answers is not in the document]
boris-kap: OPT_FIELD_OFFSET is the offset of the optional fields, which is always 2, in our case. Never 0.

core/src/main/java/com/yahoo/oak/EntryHashSet.java
boris-kap: No, we don't know how it is implemented, we don't want to guess either. So, the synchronisation on function level is not enough, we need to add a lock and acquire it before every write/read on the bitset

PR #209, opened 2022-04-08
Improve doc and implementation of OakUnsafeDirectBuffer.getByteBuffer

dynaxis:

---

This PR improves doc comment on `OakUnsafeDirectBuffer.getByteBuffer` to better reflect the current implementation after removing internal uses of `ByteBuffer`s.

Also it reimplements proper protection of read-only Oak buffers by returning read only view of `ByteBuffer` if source buffer implementation is read only. Such a measure was mistakenly removed in the course of the recent elimination of all `ByteBuffer` usage.


sanastas (commented):
The suggested changes looks good. Thank you!

Can you please a test to check that it is impossible to write into the read-only OakUnsafeDirectBuffer? It is relatively easy, for example in `InternalOakMapTest` there is 

```
  private static Integer slowDeserialize(OakScopedReadBuffer bb) {
        try {
            Thread.sleep(LONG_TRANSFORMATION_DELAY);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        return OakCommonBuildersFactory.DEFAULT_INT_SERIALIZER.deserialize(bb);
    }
```

so it is possible to cast `OakScopedReadBuffer bb` and to try to write into it... What do you think?

dynaxis:
Makes sense. I'll update the current PR soon.

dynaxis:
I added tests for all the implementations of `OakUnsafeDirectBuffer` to test writability of the returned `ByteBuffer`. And found that `UnscopeBuffer` and its subclass violate the spec and also fixed it.

In the end, the tests are the most complex part of this PR. :)

core/src/test/java/com/yahoo/oak/KeyBufferTest.java
```diff
@@ -0,0 +1,34 @@
...
+        sliceUtils = new SliceUtils();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        sliceUtils.close();
+    }
+
+    @Test
+    public void byteBufferShouldBeReadOnly() {
```
liran-funaro, line 29: Can you merge all the test classes you added into one test class? They seem to fit the same pattern and they test similar things. It makes sense to group them together.

liran-funaro (changes requested):
The PR looks good. I request only a minor change in the tests (see below).

core/src/test/java/com/yahoo/oak/KeyBufferTest.java
dynaxis: @liran-funaro The reason I separated them that way was that it's a test among many others that may be added in the future.  If you still think it's better to consolidate them all, then I'll add another commit for doing so. Please give your final answer here. Thanks.

liran-funaro: Thanks for the explanation. Since we don't have other tests for these buffers yet, I think it is better to consolidate them into a single class. But I do agree each test should be in its own method.

liran-funaro: Great. Thank you for your PR.

dynaxis: I pushed another commit to fulfill the request from you. Thank you.

liran-funaro (approved):

dynaxis:
I haven't run all of the tests except for those added by me due to some other tests failing in my environment. So do you think the failing test is due to this PR?

sanastas:
The failures looks unrelated to your PR and also to Oak in general, as this is just a way of string creation. More than that, all the tests are running upon each of your commits/pushes and you have "All checks have passed" meaning there were no failures on mvn flow.

Is it the first time you run all the tests? Maybe it is related to your local environment, Java version, etc.

sanastas:
@dynaxis , pleasure to work with you! Your PR looks very good! Thank you! I am merging to master.

dynaxis:
Happy to work with you and to contribute a small bit to the project I rely on. Thank you.

PR #210, opened 2022-04-20
DeletedMemoryAccessException is Exception

li0nr:

### DIFF
--- a/core/src/main/java/com/yahoo/oak/BasicChunk.java
+++ b/core/src/main/java/com/yahoo/oak/BasicChunk.java
@@ -66,6 +66,9 @@ abstract class BasicChunk<K, V> {
 
     abstract boolean readValueFromEntryIndex(ValueBuffer value, int ei);
 
+    /* the lookUp method throws this exception for OrderedChunks only while trying to access a released chunk.
+     * the lookUp method is not allowed to throw the exception in the case of HashChunk.
+     */
     abstract void lookUp(ThreadContext ctx, K key) throws DeletedMemoryAccessException;
 
     /*-------------- Publishing related methods and getters ---------------*/
--- a/core/src/main/java/com/yahoo/oak/BasicChunk.java
+++ b/core/src/main/java/com/yahoo/oak/BasicChunk.java
@@ -66,7 +66,7 @@ abstract class BasicChunk<K, V> {
 
     abstract boolean readValueFromEntryIndex(ValueBuffer value, int ei);
 
-    abstract void lookUp(ThreadContext ctx, K key);
+    abstract void lookUp(ThreadContext ctx, K key) throws DeletedMemoryAccessException;
 
     /*-------------- Publishing related methods and getters ---------------*/
     /**
--- a/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
+++ b/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
@@ -6,7 +6,7 @@
 
 package com.yahoo.oak;
 
-public class DeletedMemoryAccessException extends RuntimeException {
+public class DeletedMemoryAccessException extends Exception {
     public DeletedMemoryAccessException() {
         super();
     }
--- a/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
+++ b/core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
@@ -6,6 +6,10 @@
 
 package com.yahoo.oak;
 
+/**
+ * This Exception is thrown when using the Nova MemoryManager, such exception indicates that we are trying
+ * to access some off heap memory location that was deleted concurrently to the access.
+ */
 public class DeletedMemoryAccessException extends Exception {
     public DeletedMemoryAccessException() {
         super();
--- a/core/src/main/java/com/yahoo/oak/EntryArray.java
+++ b/core/src/main/java/com/yahoo/oak/EntryArray.java
@@ -182,9 +182,9 @@ public class EntryArray<K, V> {
         int nextIndex = currentIndex;
         while (!isNotZero && isIndexInBound(++nextIndex)) {
 
-            isNotZero = getEntryFieldLong(nextIndex, KEY_REF_OFFSET) != 0 ||
-                        getEntryFieldLong(nextIndex, VALUE_REF_OFFSET) != 0 ||
-                        getEntryFieldLong(nextIndex, OPT_FIELD_OFFSET) != 0;
+            isNotZero = array.getEntryFieldLong(nextIndex, KEY_REF_OFFSET) != 0 ||
+                        array.getEntryFieldLong(nextIndex, VALUE_REF_OFFSET) != 0 ||
+                        array.getEntryFieldLong(nextIndex, OPT_FIELD_OFFSET) != 0;
         }
         if (isIndexInBound(nextIndex)) {
             return nextIndex;
--- a/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
@@ -299,7 +299,9 @@ class EntryOrderedSet<K, V> extends EntryArray<K, V> {
 
     void releaseAllDeletedKeys() {
         KeyBuffer key = new KeyBuffer(config.keysMemoryManager.getEmptySlice());
-        for (int i = 0; i < nextFreeIndex.get() - 1 ; i++) {
+        for (int i = 0; i < nextFreeIndex.get() - 1 && i < numOfEntries.get() ; i++) {
+            //the second condiiton is for case where multiple threads increment the nextfreeIndex
+            //which results in being bigger than the array size
             if (!config.valuesMemoryManager.isReferenceDeleted(getValueReference(i))) {
                 continue;
             }
--- a/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
@@ -297,11 +297,13 @@ class EntryOrderedSet<K, V> extends EntryArray<K, V> {
         return true;
     }
 
+    //This method is called by only one thread on a release chunk.
     void releaseAllDeletedKeys() {
         KeyBuffer key = new KeyBuffer(config.keysMemoryManager.getEmptySlice());
         for (int i = 0; i < nextFreeIndex.get() - 1 && i < numOfEntries.get() ; i++) {
-            //the second condiiton is for case where multiple threads increment the nextfreeIndex
-            //which results in being bigger than the array size
+            //the second condition is for case where multiple threads increment the nextfreeIndex
+            //which results in being bigger than the array size, this happens when multiple threads try
+            // to increase the nextFreeIndex
             if (!config.valuesMemoryManager.isReferenceDeleted(getValueReference(i))) {
                 continue;
             }
--- a/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
+++ b/core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
@@ -299,7 +299,7 @@ class EntryOrderedSet<K, V> extends EntryArray<K, V> {
 
     void releaseAllDeletedKeys() {
         KeyBuffer key = new KeyBuffer(config.keysMemoryManager.getEmptySlice());
-        for (int i = 0; i < numOfEntries.get() ; i++) {
+        for (int i = 0; i < nextFreeIndex.get() - 1 ; i++) {
             if (!config.valuesMemoryManager.isReferenceDeleted(getValueReference(i))) {
                 continue;
             }
--- a/core/src/main/java/com/yahoo/oak/InternalOakBasics.java
+++ b/core/src/main/java/com/yahoo/oak/InternalOakBasics.java
@@ -169,6 +169,7 @@ abstract class InternalOakBasics<K, V> {
             try {
                 chunk.lookUp(ctx, key);
             } catch (DeletedMemoryAccessException e) {
+                assert !(chunk instanceof  HashChunk);
                 continue;
             }
             if (!ctx.isValueValid()) {
@@ -188,30 +189,7 @@ abstract class InternalOakBasics<K, V> {
         throw new RuntimeException("replace failed: reached retry limit (1024).");
     }
 
-    boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializeTransformer) 
-            throws DeletedMemoryAccessException {
-        ThreadContext ctx = getThreadContext();
-
-        for (int i = 0; i < MAX_RETRIES; i++) {
-            // find chunk matching key, puts this key hash into ctx.operationKeyHash
-            BasicChunk<K, V> c = findChunk(key, ctx); // find orderedChunk matching key
-            c.lookUp(ctx, key);
-            if (!ctx.isValueValid()) {
-                return false;
-            }
-
-            ValueUtils.ValueResult res = getValueOperator().compareExchange(c, ctx, oldValue, newValue,
-                    valueDeserializeTransformer, getValueSerializer());
-            if (res == ValueUtils.ValueResult.RETRY) {
-                // it might be that this chunk is proceeding with rebalance -> help
-                helpRebalanceIfInProgress(c);
-                continue;
-            }
-            return res == ValueUtils.ValueResult.TRUE;
-        }
-
-        throw new RuntimeException("replace failed: reached retry limit (1024).");
-    }
+    abstract boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializeTransformer);
 
     abstract boolean putIfAbsentComputeIfPresent(K key, V value, Consumer<OakScopedWriteBuffer> computer);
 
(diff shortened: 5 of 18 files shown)


core/src/main/java/com/yahoo/oak/DeletedMemoryAccessException.java
```diff
@@ -6,7 +6,7 @@
 
 package com.yahoo.oak;
 
-public class DeletedMemoryAccessException extends RuntimeException {
+public class DeletedMemoryAccessException extends Exception {
```
sanastas, line 13: Please add some explanatory comment about the exception

core/src/main/java/com/yahoo/oak/EntryArray.java
```diff
@@ -182,9 +182,9 @@ int getNextNonZeroIndex(int currentIndex) {
         int nextIndex = currentIndex;
         while (!isNotZero && isIndexInBound(++nextIndex)) {
 
-            isNotZero = getEntryFieldLong(nextIndex, KEY_REF_OFFSET) != 0 ||
-                        getEntryFieldLong(nextIndex, VALUE_REF_OFFSET) != 0 ||
-                        getEntryFieldLong(nextIndex, OPT_FIELD_OFFSET) != 0;
+            isNotZero = array.getEntryFieldLong(nextIndex, KEY_REF_OFFSET) != 0 ||
```
sanastas, line 185: Why is this change needed?

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -166,7 +166,11 @@ V replace(K key, V value, OakTransformer<V> valueDeserializeTransformer) {
 
         for (int i = 0; i < MAX_RETRIES; i++) {
             BasicChunk<K, V> chunk = findChunk(key, ctx); // find orderedChunk matching key
-            chunk.lookUp(ctx, key);
+            try {
```
sanastas, line 169: The comment above speaks about orderedChunk, do we know it is not going to happen for hash?

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -184,7 +188,8 @@ V replace(K key, V value, OakTransformer<V> valueDeserializeTransformer) {
         throw new RuntimeException("replace failed: reached retry limit (1024).");
     }
 
-    boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializeTransformer) {
+    boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializeTransformer) 
+            throws DeletedMemoryAccessException {
```
sanastas: It is unclear, why `replace()` method above retries upon exception and this one redirects the exception... ?

sanastas (commented):
I will continue, but I think we need a meeting.

core/src/main/java/com/yahoo/oak/BasicChunk.java
```diff
@@ -66,7 +66,7 @@ protected void updateBasicChild(BasicChunk<K, V> child) {
 
     abstract boolean readValueFromEntryIndex(ValueBuffer value, int ei);
 
-    abstract void lookUp(ThreadContext ctx, K key);
+    abstract void lookUp(ThreadContext ctx, K key) throws DeletedMemoryAccessException;
```
sanastas, line 72: Please add a comment explaining when the exception is thrown. It would better be:
1. For OrderedChunks - only if we access a released chunk
2. For HashChunk - should never happen

Am I right?

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -166,7 +166,11 @@ V replace(K key, V value, OakTransformer<V> valueDeserializeTransformer) {
 
         for (int i = 0; i < MAX_RETRIES; i++) {
             BasicChunk<K, V> chunk = findChunk(key, ctx); // find orderedChunk matching key
-            chunk.lookUp(ctx, key);
+            try {
+                chunk.lookUp(ctx, key);
+            } catch (DeletedMemoryAccessException e) {
+                continue;
```
sanastas, line 176: Can we add an assert that chunk is not hashChunk? As a sanity check...

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
sanastas: Why also here not to catch the exception around `lookUp()`?

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -355,7 +360,7 @@ public boolean hasNext() {
             return (state != null);
         }
 
-        protected abstract void initAfterRebalance();
+        protected abstract void initAfterRebalance() throws DeletedMemoryAccessException;
```
sanastas: Please add a comment describing when exception is thrown, detailed case

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -483,21 +489,28 @@ protected boolean advanceState() {
...
+                while (!chunkIter.hasNext()) { // skip empty chunks
+                    chunk = getNextChunk(chunk);
+                    if (chunk == null) {
+                        //End of iteration
+                        setState(null);
+                        return false;
+                    }
+                    try {
+                        chunkIter = getChunkIter(chunk);
+                    } catch (DeletedMemoryAccessException e) {
```
sanastas: Please add a comment when `getChunkIter()` can return `DeletedMemoryAccessException e`

core/src/main/java/com/yahoo/oak/InternalOakHash.java
```diff
@@ -269,7 +269,11 @@ boolean replace(K key, V oldValue, V newValue, OakTransformer<V> valueDeserializ
         for (int i = 0; i < MAX_RETRIES; i++) {
             // find chunk matching key, puts this key hash into ctx.operationKeyHash
             BasicChunk<K, V> c = findChunk(key, ctx);
-            c.lookUp(ctx, key);
+            try {
+                c.lookUp(ctx, key);
+            } catch (DeletedMemoryAccessException e) {
+                continue;
```
sanastas: Sounds to me like an assert should be here... Generally it shouldn't happen and subject for future work. How do you plan to deal with it right now (in the meantime)?

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -975,6 +976,7 @@ protected void initAfterRebalance() {
         /**
          * Advances next to higher entry.
          * Return previous index
+         * @throws DeletedMemoryAccessException
```
sanastas: Why only the comment??

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -1067,43 +1069,51 @@ protected void initState(boolean isDescending, K lowerBound, boolean lowerInclus
...
+                            setState(null);
+                            return;
+                        }
+                    }
+        
+                    //Init state, not valid yet, must move forward
+                    setPrevState(InternalOakMap.IteratorState.newInstance(null, null));
+                    setState(IteratorState.newInstance(nextOrderedChunk, nextChunkIter));
+                    advanceState();
+                } catch (DeletedMemoryAccessException e) {
```
sanastas, line 1142: Unclear change... From where do we get the exception here? From skiplist of chunks comparisons? It seams to me that the keys in the skiplist are objects. And minKeys on the chunks are never released, right? Let's discuss.

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -1121,7 +1131,7 @@ protected BasicChunk<K, V> getNextChunk(BasicChunk<K, V> current) {
             }
         }
 
-        protected BasicChunk.BasicChunkIter getChunkIter(BasicChunk<K, V> current) {
+        protected BasicChunk.BasicChunkIter getChunkIter(BasicChunk<K, V> current) throws DeletedMemoryAccessException {
```
sanastas, line 1170: minKeys are not released, right? From where the exception?

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -50,12 +50,13 @@ private static boolean atomicallySetDeleted(long headerAddress, long offHeapMeta
     ValueUtils.ValueResult preRead(final int onHeapVersion, long headerAddress) {
         long offHeapHeader = getOffHeapHeader(headerAddress);
         if (RC.isReferenceDeleted(offHeapHeader)) {
-            throw new DeletedMemoryAccessException();
+            return ValueUtils.ValueResult.FALSE;
```
sanastas, line 56: What is the contract of the method? Should be clearly defined. Let'd discuss whether to return false or to throw exception.

core/src/main/java/com/yahoo/oak/NovaMMHeader.java
```diff
@@ -50,12 +50,13 @@ private static boolean atomicallySetDeleted(long headerAddress, long offHeapMeta
...
+            //throw new DeletedMemoryAccessException();
         }
         return ValueUtils.ValueResult.TRUE;
     }
 
-    ValueUtils.ValueResult postRead(final int onHeapVersion, long headerAddress) {
+    ValueUtils.ValueResult postRead(final int onHeapVersion, long headerAddress) throws DeletedMemoryAccessException {
```
sanastas, line 66: Need to add good comments here and above, so the contract is clear.

core/src/main/java/com/yahoo/oak/NovaMemoryManager.java
```diff
@@ -191,6 +191,7 @@ public long getAddress() {
          * {@code FALSE} if the header/off-heap-cut is marked as deleted
          * {@code RETRY} if the header/off-heap-cut was moved, or the version of the off-heap header
          * does not match {@code version}.
+         * @throws DeletedMemoryAccessException
```
sanastas: Why only comment??

core/src/main/java/com/yahoo/oak/OakMap.java
```diff
@@ -267,12 +267,20 @@ public Entry<K, V> lowerEntry(K key) {
         if (this.isSubmap()) {
             throw new UnsupportedOperationException();
         }
+        for (int i = 0; i < InternalOakBasics.MAX_RETRIES; i++) {
```
sanastas: I think the reties should be treated on the `InternalOakMap` level

core/src/main/java/com/yahoo/oak/OrderedChunk.java
```diff
@@ -231,8 +231,9 @@ private int getLastSortedEntryIndex(int sortedCount) {
      * @param key         the key to compare
      * @param ei          the entry index to compare with
      * @return the comparison result
+     * @throws DeletedMemoryAccessException 
      */
-    int compareKeyAndEntryIndex(KeyBuffer tempKeyBuff, K key, int ei) {
+    int compareKeyAndEntryIndex(KeyBuffer tempKeyBuff, K key, int ei) throws DeletedMemoryAccessException {
```
sanastas, line 237: Why exception here, upon `entryOrderedSet.readKey()` failure?

core/src/main/java/com/yahoo/oak/InternalOakMap.java
li0nr: lets discuss

core/src/main/java/com/yahoo/oak/EntryOrderedSet.java
```diff
@@ -299,7 +299,9 @@ boolean isEntrySetValidAfterRebalance() {
 
     void releaseAllDeletedKeys() {
         KeyBuffer key = new KeyBuffer(config.keysMemoryManager.getEmptySlice());
-        for (int i = 0; i < numOfEntries.get() ; i++) {
+        for (int i = 0; i < nextFreeIndex.get() - 1 && i < numOfEntries.get() ; i++) {
```
sanastas, line 303: This looks to me very weird condition and comment.
1. `releaseAllDeletedKeys()` should be called only by one thread that succeeds to change the chunk's state from FROZEN to RELEASED (CAS success). I do not see any reason for other threads to invoke `releaseAllDeletedKeys()`. Can you please check that this is how it happens? And definitely add a header comment about `releaseAllDeletedKeys()` usage intention.
2. This release happens as a very last step in the Chunk rebalance. At the beginning of the rebalance step the thread proceeding with the rebalance waits for all updates to finish. This is why we have publish/unpublish staff. So if `nextFreeIndex` is concurrently updated this is a bug... Maybe to add an explanatory comment here and simplify the loop condition. I think checking only `nextFreeIndex` should be enough.

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
sanastas: Can you please add an explanatory comment here?

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -471,7 +456,8 @@ protected void invalidatePrevState() {
 
         protected abstract BasicChunk<K, V> getNextChunk(BasicChunk<K, V> current);
 
-        protected abstract BasicChunk.BasicChunkIter getChunkIter(BasicChunk<K, V> current);
+        protected abstract BasicChunk.BasicChunkIter getChunkIter(BasicChunk<K, V> current)
```
sanastas, line 463: It would be good to explain in which cases the exception is thrown. As I remember we were asking ourselves as well. Please add a comment.

core/src/main/java/com/yahoo/oak/InternalOakHash.java
sanastas: How to you plan to deal with it? Still seems strange... How can we catch an exception on HashChunk? Why to continue?

sanastas: The same for all following exceptions catching in this file.

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
```diff
@@ -355,6 +338,8 @@ public boolean hasNext() {
             return (state != null);
         }
 
+        protected abstract void initStateWithNotDeletedChunk(BasicChunk chunk);
```
sanastas: Please add a comment explaining what is this (new) function, why is it needed, when is it used, what should it generally do?

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -971,12 +981,33 @@ protected void initAfterRebalance() {
             // Update the state to point to last returned key.
             initState(isDescending, lo, loInclusive, hi, hiInclusive);
         }
+        
+        @Override
+        protected void initStateWithNotDeletedChunk(BasicChunk chunk) {
+            OrderedChunk oChunk = (OrderedChunk) chunk;
+            K nextKey = null;
+            try {
+                nextKey = KeyUtils.deSerializedKey(oChunk.minKey, getKeySerializer());
```
sanastas, line 991: explain that minKey is never released

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -971,12 +981,33 @@ protected void initAfterRebalance() {
...
+            } catch (DeletedMemoryAccessException e) {
+                return ; //should not happen
```
sanastas: If it shouldn't happen better assert!!

core/src/main/java/com/yahoo/oak/InternalOakMap.java
sanastas: I still want to discuss it...

core/src/main/java/com/yahoo/oak/InternalOakMap.java
```diff
@@ -1067,52 +1098,60 @@ protected void initState(boolean isDescending, K lowerBound, boolean lowerInclus
...
-            setPrevState(InternalOakMap.IteratorState.newInstance(null, null));
-            setState(IteratorState.newInstance(nextOrderedChunk, nextChunkIter));
-            advanceState();
+            throw new RuntimeException("reached retry limit (1024).");
         }
 
         @Override
         protected BasicChunk<K, V> getNextChunk(BasicChunk<K, V> current) {
-            OrderedChunk<K, V> currentHashChunk = (OrderedChunk<K, V>) current;
+            OrderedChunk<K, V> currenChunk = (OrderedChunk<K, V>) current;
```
sanastas: Why `currenChunk` and not `currentChunk`? :) Do not save one letter. :)

sanastas (commented):
I have looked over majority of the code. I didn't find something drastic :) Very good job! But we still have some things to discuss. Please go over my comments and also the old ones which still remained unresolved.

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
li0nr: this opened to be discussed

core/src/main/java/com/yahoo/oak/InternalOakBasics.java
li0nr: to be discussed

